PR #26, opened 2021-06-06
update chat2 example

richard-ramos:


examples/chat2/main.go
```diff
@@ -22,21 +22,29 @@ import (
 	"github.com/status-im/go-waku/waku/v2/protocol/store"
 )
 
-var DefaultContentTopic string = "dingpu"
+var DefaultContentTopic string = "/waku/2/huilong/proto"
```
D4nte, line 25: The content topic is actually `/toy-chat/2/huilong/proto` now.

Issue #39, opened 2021-07-21
Implement LIBP2P-DNS-DISCOVERY


richard-ramos:
https://rfc.vac.dev/spec/25/

cammellos:
After speaking with @jm-clius , we will implement ENR DNS discovery and map ENR to multiaddr.

Likely we have already the code for ENR discovery as go-waku already relies on go-ethereum, so the only custom code we need to write is some wiring and the mapping.

jm-clius:
Correct. `nim-waku` is just using a nim implementation of [EIP-1459](https://eips.ethereum.org/EIPS/eip-1459)
POC example tree for the `wakuv2.test` fleet being deployed in https://github.com/status-im/infra-nim-waku/issues/23

If in future for whatever reason we want to implement `25/LIBP2P-DNS-DISCOVERY`, required changes should be minimal.

PR #48, opened 2021-09-01
[Fixes: #39] Add DNS discovery

cammellos:
This commit adds a public method that takes a url and returns a list of
multiaddr.

Once we better understand how to integrate it we can make it so that is
passed as a config when initializing waku, this commit only provides the
basic functionality.


siphiuel (approved):

waku/v2/discovery/enr.go
```diff
@@ -0,0 +1,43 @@
...
+	return multiAddrs, nil
+}
+
+func enodeToMultiAddr(node *enode.Node) (string, error) {
+	peerID, err := peer.IDFromPublicKey(&ECDSAPublicKey{node.Pubkey()})
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("/ip4/%s/tcp/%d/p2p/%s", node.IP(), node.TCP(), peerID), nil
```
richard-ramos, line 42: What do you think if instead of returning a string, we return a multiaddress?
```suggestion
	return ma.NewMultiaddr(fmt.Sprintf("/ip4/%s/tcp/%d/p2p/%s", node.IP(), node.TCP(), peerID)), nil
```

and return an array of multiaddresses in `RetrieveNodes` instead of an array of `string`

cammellos: Sure, sounds good, I'll change that

PR #62, opened 2021-10-04
feat: options to generate/specify/load node privatekey

richard-ramos:


waku/node.go
```diff
@@ -245,3 +242,95 @@ func addPeers(wakuNode *node.WakuNode, addresses []string, protocol protocol.ID)
...
+	if err := os.Remove(path); err != nil && !os.IsNotExist(err) {
+		return err
+	}
+
+	key, err := crypto.GenerateKey()
+	if err != nil {
+		return err
+	}
+
+	privKey := libp2pcrypto.PrivKey((*libp2pcrypto.Secp256k1PrivateKey)(key))
```
PascalPrecht, line 284: I think libp2p crypto supports other private key schemes as well. Shall we rather have an additional option that specifies the PK's scheme and have the client use that instead of only supporting one?

Alternatively, we could bruteforce the creation of `PrivateKey`. When errors on decode we know the scheme must be wrong, then we can try the next.

PascalPrecht: After talking with @jm-clius about this, secp256k1 is considered expected, so no need to check for other PK schemes.

PR #118, opened 2021-10-31
refactor: store and lightpush

richard-ramos:
To reduce the size of wakunode2.go and stop it from becoming a god object. The proper way to consume those protocols functionality will be like this:
```go
node, _ = wakunode.New()

// Lightpush
node.LightPush().Publish(...)

// Store
node.Store().Query(...)
node.Store().Resume(...)
```
In a separate PR I'll do similar changes to Relay and Filter protocols

### DIFF
--- a/examples/chat2/main.go
+++ b/examples/chat2/main.go
@@ -205,7 +205,7 @@ func main() {
 		q := store.Query{
 			ContentTopics: []string{*contentTopicFlag},
 		}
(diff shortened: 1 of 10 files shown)


alaibe (approved):
Yeah it makes much more sense. Do you think it would be valuable to have an interface showing what should be the API for a sub node? e.g being able to a minimum to have Start/Stop

richard-ramos:
Good idea!

PR #119, opened 2021-10-31
refactor: relay

richard-ramos:
```go
node, _ = wakunode.New()

// Relay
node.Relay().Publish(...)
node.Relay().Subscribe(...)
```


alaibe (approved):
:+1: the broadcaster being injected will be very helpful when sending a message, that will avoid to depends on a Node

PR #129, opened 2021-11-05
feat: message retention policy for wakustore

### LINKED ISSUE #69: Production readiness for store protocol
- Days a message can persist
- Memory usage

richard-ramos:
Fixes #69

### DIFF
--- a/waku/node.go
+++ b/waku/node.go
@@ -145,7 +145,7 @@ func Execute(options Options) {
 
 	if options.Store.Enable {
 		maxDays := time.Hour * 24 * time.Duration(options.Store.RetentionMaxDays)
-		nodeOpts = append(nodeOpts, node.WithWakuStoreAndLimits(true, options.Store.ShouldResume, maxDays, options.Store.RetentionMaxMessages))
+		nodeOpts = append(nodeOpts, node.WithWakuStoreAndLimits(options.Store.ShouldResume, maxDays, options.Store.RetentionMaxMessages))
 		if options.UseDB {
 			dbStore, err := persistence.NewDBStore(persistence.WithDB(db), persistence.WithRetentionPolicy(options.Store.RetentionMaxMessages, maxDays))
 			failOnErr(err, "DBStore")
--- a/waku/node.go
+++ b/waku/node.go
@@ -144,7 +144,7 @@ func Execute(options Options) {
 	}
 
 	if options.Store.Enable {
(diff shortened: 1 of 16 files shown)


waku/node.go
```diff
@@ -144,9 +144,10 @@ func Execute(options Options) {
 	}
 
 	if options.Store.Enable {
-		nodeOpts = append(nodeOpts, node.WithWakuStore(true, options.Store.ShouldResume))
+		maxDays := time.Hour * 24 * time.Duration(options.Store.RetentionMaxDays)
```
alaibe, line 147: nit: could be a function of StoreOptionsL
```
func (s *StoreOptions) RetentionMaxDaysDuration() time.Duration
```

waku/persistence/store.go
```diff
@@ -49,17 +54,32 @@ func WithDriver(driverName string, datasourceName string) DBOption {
...
+func WithRetentionPolicy(maxMessages int, maxDays time.Duration) DBOption {
+	return func(d *DBStore) error {
+		d.maxDays = maxDays
+		d.maxMessages = maxMessages
+		return nil
+	}
+}
+
 // Creates a new DB store using the db specified via options.
 // It will create a messages table if it does not exist
```
alaibe, line 66: nit: and clean older records

waku/persistence/store.go
```diff
@@ -84,6 +104,28 @@ func (d *DBStore) createTable() error {
 	return nil
 }
 
+func (d *DBStore) cleanOlderRecords() error {
```
alaibe, line 107: shall we have a goroutine cleaning the record periodically?

waku/v2/protocol/store/message_queue.go
```diff
@@ -0,0 +1,25 @@
+package store
+
+type MessageQueue struct {
```
alaibe, line 3: the message queue doesn t really deal with retention time? only the number of message we store.As far as I understand, the queue is here as a cache layer?

alaibe (approved):
Nice work!

waku/persistence/store.go
richard-ramos: Not sure, currently go-waku only fetches the records from the DB during login, and this function would have been executed before so the records returned would be valid according to the retention policy

waku/v2/protocol/store/message_queue.go
richard-ramos: Yeah, currently it only deals with number of messages. Perhaps the goroutine for cleaning older records can be added here?

PR #132, opened 2021-11-06
refactor: filter

richard-ramos:
Moves the filter logic from wakunode2 to waku_filter


alaibe (approved):
I had such a hard time understanding the filter, this is very good that it is now encapsulated that way :heart:

PR #134, opened 2021-11-06
feat: clean up older records in message queue

### LINKED ISSUE #133: Add goroutine to periodically clear old messages from message queue

richard-ramos:
Fixes #133

It also moves the logic for verification of duplicated messages to the message queue

### DIFF
--- a/waku/v2/protocol/store/message_queue.go
+++ b/waku/v2/protocol/store/message_queue.go
@@ -1,11 +1,39 @@
 package store
 
+import (
+	"sync"
+	"time"
+
+	"github.com/status-im/go-waku/waku/v2/utils"
+)
+
 type MessageQueue struct {
+	sync.RWMutex
+
+	seen        map[[32]byte]struct{}
 	messages    []IndexedWakuMessage
 	maxMessages int
+	maxDuration time.Duration
+
+	quit chan struct{}
+}
+
+type MessageQueueItem struct {
+	Index int
+	Value IndexedWakuMessage
 }
 
 func (self *MessageQueue) Push(msg IndexedWakuMessage) {
+	self.Lock()
+	defer self.Unlock()
+
+	var k [32]byte
+	copy(k[:], msg.index.Digest)
+	if _, ok := self.seen[k]; ok {
+		return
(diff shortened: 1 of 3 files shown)


waku/v2/protocol/store/message_queue.go
```diff
@@ -1,11 +1,39 @@
...
+	"sync"
+	"time"
+
+	"github.com/status-im/go-waku/waku/v2/utils"
+)
+
 type MessageQueue struct {
+	sync.RWMutex
+
+	seen        map[[32]byte]struct{}
```
alaibe, line 13: could messages be a map directly? we wouldn't have to maintain 2 entry?

waku/v2/protocol/store/message_queue.go
```diff
@@ -14,12 +42,73 @@ func (self *MessageQueue) Push(msg IndexedWakuMessage) {
...
+		return
+	case <-ticker.C:
+		self.cleanOlderRecords()
+	}
 }
 
-func NewMessageQueue(maxMessages int) *MessageQueue {
-	return &MessageQueue{
+func (self *MessageQueue) Length() int {
+	self.RLock()
```
alaibe, line 92: do we need to lock here?

alaibe (approved):
This PR is https://media4.giphy.com/media/3ohzdIuqJoo8QdKlnW/giphy.gif?cid=ecf05e478hqan5qh81siiomo575wbns3mcyllvsmg9cr5en3&rid=giphy.gif&ct=g

waku/v2/protocol/store/message_queue.go
richard-ramos: The somewhat difficult part of using a single map instead of a map+slice would be to efficiently iterate the messages for cleaning up older records, as well as keeping the message queue limited to N records, because maps do not preserve order.

alaibe: I missed that part, good call. Indeed the need is an ordered Hash map, in order to build an ordered set (set by default doesn't preserve order). it seems overkill suddenly!

PR #136, opened 2021-11-07
feat: Add relay rpc methods

alaibe:
Let me know what you think of the unsubscribe and recover part in waku relay

### DIFF
--- a/waku/v2/protocol/relay/waku_relay.go
+++ b/waku/v2/protocol/relay/waku_relay.go
@@ -227,9 +227,38 @@ func (w *WakuRelay) Subscribe(ctx context.Context, topic *Topic) (*Subscription,
 	return subscription, nil
 }
 
+func (w *WakuRelay) Unsubscribe(ctx context.Context, topic Topic) error {
+	if _, ok := w.topics[topic]; !ok {
+		return fmt.Errorf("topics %s is not subscribed", (string)(topic))
+	}
+	log.Info("Unsubscribing from topic ", topic)
+	delete(w.topics, topic)
+
+	for _, sub := range w.subscriptions[topic] {
+		sub.Unsubscribe()
+	}
+
+	w.relaySubs[topic].Cancel()
+	delete(w.relaySubs, topic)
+
+	err := w.wakuRelayTopics[topic].Close()
+	if err != nil {
+		return err
+	}
+	delete(w.wakuRelayTopics, topic)
+
+	return nil
+}
+
 func (w *WakuRelay) nextMessage(ctx context.Context, sub *pubsub.Subscription) <-chan *pubsub.Message {
 	msgChannel := make(chan *pubsub.Message, 1024)
 	go func(msgChannel chan *pubsub.Message) {
+		defer func() {
+			if r := recover(); r != nil {
+				log.Debug("recovered msgChannel")
+			}
+		}()
+
 		for {
 			msg, err := sub.Next(ctx)
 			if err != nil {
@@ -264,6 +293,9 @@ func (w *WakuRelay) subscribeToTopic(t Topic, subscription *Subscription, sub *p
 			}
 			// TODO: if there are no more relay subscriptions, close the pubsub subscription
 		case msg := <-subChannel:
+			if msg == nil {
+				return
+			}
 			stats.Record(ctx, metrics.Messages.M(1))
 			wakuMessage := &pb.WakuMessage{}
 			if err := proto.Unmarshal(msg.Data, wakuMessage); err != nil {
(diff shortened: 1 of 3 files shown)


waku/v2/rpc/relay.go
```diff
@@ -0,0 +1,69 @@
...
+	reply.Success = true
+	return nil
+}
+
+func (r *RelayService) DeleteV1Subscription(req *http.Request, args *TopicsArgs, reply *SuccessReply) error {
+	ctx := req.Context()
+	for _, topic := range args.Topics {
+		err := r.node.Relay().Unsubscribe(ctx, (*relay.Topic)(&topic))
+		if err != nil {
+			log.Info("Error unsubscribing from topic:", topic, "err:", err)
```
richard-ramos, line 61:
```suggestion
			log.Error("Error unsubscribing from topic:", topic, "err:", err)
```

waku/v2/protocol/relay/waku_relay.go
```diff
@@ -227,9 +227,38 @@ func (w *WakuRelay) Subscribe(ctx context.Context, topic *Topic) (*Subscription,
...
+	delete(w.relaySubs, *topic)
+
+	err := w.wakuRelayTopics[*topic].Close()
+	if err != nil {
+		return err
+	}
+	delete(w.wakuRelayTopics, *topic)
+
+	return nil
+}
```
richard-ramos, line 251: What do you think if instead of receiving a pointer to a `Topic`,  we receive the `Topic` directly?
I also added the topics and subscription mutex:

```suggestion
func (w *WakuRelay) Unsubscribe(ctx context.Context, topic Topic) error {
	w.topicsMutex.Lock()
	defer w.topicsMutex.Unlock()

	if _, ok := w.topics[topic]; !ok {
		return fmt.Errorf("topics %s is not subscribed", topic)
	}
	log.Info("Unsubscribing from topic ", topic)
	delete(w.topics, topic)

	w.subscriptionsMutex.Lock()
	for _, sub := range w.subscriptions[topic] {
		sub.Unsubscribe()
	}
	subscriptionsMutex.Unlock()

	w.relaySubs[topic].Cancel()
	delete(w.relaySubs, topic)

	err := w.wakuRelayTopics[topic].Close()
	if err != nil {
		return err
	}
	delete(w.wakuRelayTopics, topic)

	return nil
}
```

waku/v2/protocol/relay/waku_relay.go
```diff
@@ -227,9 +227,38 @@ func (w *WakuRelay) Subscribe(ctx context.Context, topic *Topic) (*Subscription,
...
 }
 
+func (w *WakuRelay) Unsubscribe(ctx context.Context, topic *Topic) error {
+	if _, ok := w.topics[*topic]; !ok {
+		return fmt.Errorf("topics %s is not subscribed", (string)(*topic))
+	}
+	log.Info("Unsubscribing from topic ", *topic)
+	delete(w.topics, *topic)
+
+	for _, sub := range w.subscriptions[*topic] {
```
richard-ramos, line 237: Should we also delete the subscriptions? (delete(w.subscriptions, topic)
I'm not sure what would happen if we do that and there are active subscriptions being used

waku/v2/protocol/relay/waku_relay.go
```diff
@@ -227,9 +227,38 @@ func (w *WakuRelay) Subscribe(ctx context.Context, topic *Topic) (*Subscription,
...
+	delete(w.wakuRelayTopics, *topic)
+
+	return nil
+}
+
 func (w *WakuRelay) nextMessage(ctx context.Context, sub *pubsub.Subscription) <-chan *pubsub.Message {
 	msgChannel := make(chan *pubsub.Message, 1024)
 	go func(msgChannel chan *pubsub.Message) {
+		defer func() {
+			if r := recover(); r != nil {
```
richard-ramos, line 257: TIL about `recover()`!

richard-ramos (approved):
:1st_place_medal:

waku/v2/protocol/relay/waku_relay.go
alaibe: I didn't delete them because of this:
```
case <-subscription.quit:
			if w.bcaster != nil {
				w.bcaster.Unregister(subscription.C) // Remove from broadcast list
			}
			// TODO: if there are no more relay subscriptions, close the pubsub subscription
```
Shall we delete them there?

waku/v2/protocol/relay/waku_relay.go
alaibe: Yeah I am not sure if this is the best way but `msgChannel <- msg` cause a panic, I think to improve it we could intercept the close/quit and handle that properly

PR #141, opened 2021-11-09
feat: Add rpc store

alaibe:
Add test for rpc services

### DIFF
--- a/waku/v2/rpc/store.go
+++ b/waku/v2/rpc/store.go
@@ -1,0 +1,68 @@
+package rpc
+
+import (
+	"net/http"
+
+	"github.com/status-im/go-waku/waku/v2/node"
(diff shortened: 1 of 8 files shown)


waku/v2/rpc/store.go
```diff
@@ -0,0 +1,62 @@
...
+			EndTime:       args.EndTime,
+		},
+		options...,
+	)
+	if err != nil {
+		log.Error("Error querying messages:", err)
+		reply.Error = err.Error()
+		return nil
+	}
+	reply.Messages = res.Messages
```
richard-ramos, line 60: reply.Me
```suggestion
	reply.Messages = res.Messages
	reply.pagingInfo = etc etc etc
```

alaibe: oops!

richard-ramos (approved):

PR #143, opened 2021-11-09
feat: select peer with lowest ping time and test peer functions

richard-ramos:


waku/v2/utils/peer.go
```diff
@@ -37,5 +44,75 @@ func SelectPeer(host host.Host, protocolId string) (*peer.ID, error) {
...
+	pingCh := make(chan pingResult, 1000)
+
+	go func() {
+		for _, p := range peers {
+			wg.Add(1)
+			go func(p peer.ID) {
+				defer wg.Done()
+				ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+				defer cancel()
+				result := <-pingService.Ping(ctx, p)
```
alaibe, line 84: Coud we not build the ping service before the go func (line 79) that would avoid to pass it down, I also imagine that this is quite simple lib? Maybe you want to offer to choose which ping service to use?

alaibe (approved):

waku/v2/utils/peer.go
richard-ramos: it seems that i can use Ping function directly without having to instantiate the pingService.  https://github.com/libp2p/go-libp2p/blob/master/p2p/protocol/ping/ping.go#L84-L130  
I'll update the code to avoid passing a ping service!

PR #145, opened 2021-11-10
feat: add admin rpc methods

alaibe:


waku/v2/node/wakunode2.go
```diff
@@ -395,6 +402,35 @@ func (w *WakuNode) Peers() PeerStats {
...
+		peers = append(peers, &Peer{
+			ID:        peerId,
+			Protocols: protocols,
+			Connected: connected,
+			Addrs:     addrs,
+		})
+	}
+
+	return peers, nil
+}
```
richard-ramos, line 432:
```suggestion
func (w *WakuNode) Peers() ([]*Peer, error) {
	var peers []*Peer
	for _, peerId := range w.host.Peerstore().Peers() {
		connected := w.host.Network().Connectedness(peerId) == network.Connected 

		protocols, err := w.host.Peerstore().GetProtocols(peerId)
		if err != nil {
			return nil, err
		}
		addrs := w.host.Peerstore().Addrs(peerId)

		peers = append(peers, &Peer{
			ID:        peerId,
			Protocols: protocols,
			Connected: connected,
			Addrs:     addrs,
		})
	}

	return peers, nil
}
```

waku/v2/rpc/admin.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+
+		// TODO: understand how to use the protocol
+		_, err = a.node.AddPeer(addr, relay.WakuRelayID_v200)
+		if err != nil {
+			log.Error("Error adding peers", err)
+			reply.Success = false
+			reply.Error = err.Error()
+			return nil
+		}
```
richard-ramos, line 50: I think in this case you need to use `a.node.DialPeerWithMultiAddress` or `a.node.DialPeer` (the former requires a multiaddress while the later does the string to multiaddress conversion).

`AddPeer` is used to populate the peerstore, but does not connect to the peer

waku/v2/node/wakunode2.go
```diff
@@ -395,6 +402,35 @@ func (w *WakuNode) Peers() PeerStats {
...
+				connected = true
+				break
+			}
+		}
+
+		protocols, err := w.host.Peerstore().GetProtocols(peerId)
+		if err != nil {
+			return nil, err
+		}
+		addrs := w.host.Peerstore().Addrs(peerId)
```
richard-ramos: Do you know if nim-waku returns the multiaddress ending in `p2p/the-host-id`?

richard-ramos (approved):

richard-ramos (approved):

PR #149, opened 2021-11-15
feat: discoveryV5 - part1

richard-ramos:
- [X] Implement discovery V5 POC
- [X] ENR filtering on waku field
- [X] Test
- [ ] Autoupdate ENR
- [ ] Add discoveryV5 options to `wakunode2.go`
- [ ] Add flags to node.go to start and configure go-waku node
  - discv5-discovery: Enable discovering nodes via Node Discovery v5
  - discv5-udp-port: Listening UDP port for Node Discovery v5."
  - discv5-bootstrap-node: Text-encoded ENR for bootstrap node. Used when connecting to the network. 
  - discv5-enr-auto-update: "Discovery can automatically update its ENR with the IP address and UDP port as seen by other nodes it communicates with. "

### DIFF
--- a/waku/v2/discv5/discover.go
+++ b/waku/v2/discv5/discover.go
@@ -1,0 +1,358 @@
+package discv5
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"math/rand"
+	"net"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum/p2p/discover"
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/ethereum/go-ethereum/p2p/enr"
+	logging "github.com/ipfs/go-log"
+	"github.com/libp2p/go-libp2p-core/discovery"
+	"github.com/libp2p/go-libp2p-core/host"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/status-im/go-waku/waku/v2/utils"
+)
+
+var log = logging.Logger("waku_discv5")
+
+type DiscoveryV5 struct {
+	discovery.Discovery
+
+	params    *discV5Parameters
+	host      host.Host
+	config    discover.Config
+	udpAddr   *net.UDPAddr
+	listener  *discover.UDPv5
+	localnode *enode.LocalNode
+
+	peerCache peerCache
+}
+
+type peerCache struct {
+	sync.RWMutex
+	recs map[peer.ID]peerRecord
+	rng  *rand.Rand
+}
+
+type peerRecord struct {
+	expire int64
+	peer   peer.AddrInfo
+}
+
+type discV5Parameters struct {
+	host             host.Host
+	bootnodes        []*enode.Node
+	advertiseAddress *net.IP
+	udpPort          int
+}
+
+const WakuENRField = "waku2"
+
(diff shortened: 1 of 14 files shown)


waku/v2/discv5/discover.go
```diff
@@ -0,0 +1,357 @@
...
+	discV5.udpAddr = &net.UDPAddr{
+		IP:   ipAddr,
+		Port: params.udpPort,
+	}
+
+	return discV5, nil
+}
+
+func newLocalnode(priv *ecdsa.PrivateKey, ipAddr net.IP, udpPort int, tcpPort int, wakuFlags WakuEnrBitfield, advertiseAddr *net.IP) (*enode.LocalNode, error) {
+	db, err := enode.OpenDB("")
```
alaibe, line 139: why the empty string?

waku/v2/discv5/discover.go
```diff
@@ -0,0 +1,357 @@
...
+	discV5.peerCache.recs = make(map[peer.ID]peerRecord)
+	discV5.localnode = localnode
+	discV5.config = discover.Config{
+		PrivateKey: priv,
+		Bootnodes:  params.bootnodes,
+	}
+	discV5.udpAddr = &net.UDPAddr{
+		IP:   ipAddr,
+		Port: params.udpPort,
+	}
```
alaibe, line 133: any reason of not doing:
```
return &DiscoveryV5{
  host: host,
  params: params,
  .....
}, nil
```

waku/v2/discv5/discover.go
```diff
@@ -0,0 +1,357 @@
...
+
+func (c *DiscoveryV5) Advertise(ctx context.Context, ns string, opts ...discovery.Option) (time.Duration, error) {
+	// Get options
+	var options discovery.Options
+	err := options.Apply(opts...)
+	if err != nil {
+		return 0, err
+	}
+
+	// TODO: once discv5 spec introduces capability and topic discovery, implement this function
```
alaibe, line 243: this looks like it does nothing yet right?

waku/v2/discv5/discover_test.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/status-im/go-waku/tests"
+	"github.com/stretchr/testify/require"
+
+	"github.com/libp2p/go-libp2p"
+	libp2pcrypto "github.com/libp2p/go-libp2p-core/crypto"
+	"github.com/libp2p/go-libp2p-core/discovery"
+	"github.com/libp2p/go-libp2p-core/host"
+)
+
+func createHost(t *testing.T) (host.Host, int, *ecdsa.PrivateKey) {
```
alaibe, line 23: can you not use the MakeHost func from tests package?

alaibe (approved):

alaibe:
Ohh and now that we care about code climate, there is 4 issues to fix ;)

waku/v2/discv5/discover.go
richard-ramos: An empty string means that a in-memory database will be used for storing the enodes. It's not necessary to persist the enode information since these will be converted to multiaddresses and we already have persistent peer storage for those

waku/v2/discv5/discover.go
richard-ramos: Nah, I'll change it to the format you suggested

waku/v2/discv5/discover.go
richard-ramos: Yeah, this part has not been specified yet.

PR #156, opened 2021-11-18
feat: implement rcp relay get messages

alaibe:
The rpc service subscribe to the messages as per the specs here:
https://rfc.vac.dev/spec/16/#get_waku_v2_relay_v1_messages

Let me know if you have a better idea/place on how to implement this. I hesitate to make the relay itself message aware but I feel it is out of his responsabilities. WDYT

### DIFF
--- a/waku/v2/rpc/relay.go
+++ b/waku/v2/rpc/relay.go
@@ -3,14 +3,22 @@ package rpc
 import (
 	"fmt"
 	"net/http"
+	"sync"
 
 	"github.com/status-im/go-waku/waku/v2/node"
+	"github.com/status-im/go-waku/waku/v2/protocol"
 	"github.com/status-im/go-waku/waku/v2/protocol/pb"
 	"github.com/status-im/go-waku/waku/v2/protocol/relay"
 )
 
 type RelayService struct {
 	node *node.WakuNode
+
+	messages      map[string][]*pb.WakuMessage
+	messagesMutex sync.RWMutex
+
+	ch   chan *protocol.Envelope
+	quit chan bool
 }
 
 type RelayMessageArgs struct {
@@ -20,6 +28,53 @@ type RelayMessageArgs struct {
 
 type TopicsArgs struct {
 	Topics []string `json:"topics,omitempty"`
+}
+
+type TopicArgs struct {
+	Topic string `json:"topic,omitempty"`
+}
+
+type MessagesReply struct {
+	Messages []*pb.WakuMessage `json:"messages,omitempty"`
+}
+
+func NewRelayService(node *node.WakuNode) *RelayService {
+	return &RelayService{
+		node:     node,
+		messages: make(map[string][]*pb.WakuMessage),
+		quit:     make(chan bool),
+	}
+}
+
+func (r *RelayService) addEnvelope(envelope *protocol.Envelope) {
+	r.messagesMutex.Lock()
+	defer r.messagesMutex.Unlock()
+
+	if _, ok := r.messages[envelope.PubsubTopic()]; !ok {
+		return
+	}
+
+	r.messages[envelope.PubsubTopic()] = append(r.messages[envelope.PubsubTopic()], envelope.Message())
+}
+
+func (r *RelayService) Start() {
+	r.ch = make(chan *protocol.Envelope, 1024)
+	r.node.Broadcaster().Register(r.ch)
+
+	for {
+		select {
+		case <-r.quit:
+			return
+		case envelope := <-r.ch:
+			r.addEnvelope(envelope)
+		}
+	}
+}
+
+func (r *RelayService) Stop() {
+	r.quit <- true
+	r.node.Broadcaster().Unregister(r.ch)
+	close(r.ch)
 }
 
 func (r *RelayService) PostV1Message(req *http.Request, args *RelayMessageArgs, reply *SuccessReply) error {
@@ -44,6 +99,7 @@ func (r *RelayService) PostV1Subscription(req *http.Request, args *TopicsArgs, r
 			reply.Error = err.Error()
 			return nil
 		}
+		r.messages[topic] = make([]*pb.WakuMessage, 0)
 	}
 	reply.Success = true
 	return nil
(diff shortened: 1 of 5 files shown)


waku/v2/rpc/relay.go
```diff
@@ -5,12 +5,16 @@ import (
...
 
 	"github.com/status-im/go-waku/waku/v2/node"
+	"github.com/status-im/go-waku/waku/v2/protocol"
 	"github.com/status-im/go-waku/waku/v2/protocol/pb"
 	"github.com/status-im/go-waku/waku/v2/protocol/relay"
 )
 
 type RelayService struct {
 	node *node.WakuNode
+
```
richard-ramos, line 15:
```suggestion
	quit chan struct{}
```

waku/v2/rpc/waku_rpc.go
```diff
@@ -28,7 +32,13 @@ func NewWakuRpc(node *node.WakuNode, address string, port int) *WakuRpc {
 		log.Error(err)
 	}
 
-	err = s.RegisterService(&RelayService{node}, "Relay")
+	relayService := &RelayService{
+		node,
+		make(map[string][]*pb.WakuMessage),
+		make(chan *protocol.Envelope, 1024),
```
richard-ramos, line 38:
```suggestion
		make(chan *protocol.Envelope, 1024),
		make(chan struct{}),
```

waku/v2/rpc/waku_rpc.go
```diff
@@ -28,7 +32,13 @@ func NewWakuRpc(node *node.WakuNode, address string, port int) *WakuRpc {
 		log.Error(err)
 	}
 
-	err = s.RegisterService(&RelayService{node}, "Relay")
+	relayService := &RelayService{
```
richard-ramos, line 35: What do you think of creating a `NewRelayService` function that creates a `*RelayService`, so `WakuRpc` does not have to be aware of the internals of `RelayService` struct

waku/v2/rpc/relay_test.go
```diff
@@ -14,7 +19,11 @@ func makeRelayService(t *testing.T) *RelayService {
 	require.NoError(t, err)
 	err = n.Start()
 	require.NoError(t, err)
-	return &RelayService{n}
+	return &RelayService{
+		n,
+		make(map[string][]*pb.WakuMessage),
```
richard-ramos, line 24: If we end up closing `ch` in `Stop()` then `ch` should be initialized in `Start`, so it's possible to restart the RelayService by calling `Start` again  (maybe we'll add an rpc method or api function to start/stop services and protocols on demand)

waku/v2/rpc/relay.go
```diff
@@ -22,6 +26,30 @@ type TopicsArgs struct {
...
+func (r *RelayService) Start() {
+	r.node.Broadcaster().Register(r.ch)
+
+	for envelope := range r.ch {
+		if _, ok := r.messages[envelope.PubsubTopic()]; !ok {
+			continue
+		}
+
+		r.messages[envelope.PubsubTopic()] = append(r.messages[envelope.PubsubTopic()], envelope.Message())
+	}
```
richard-ramos, line 46:
```suggestion
    for {
       select {
           case <- r.quit:
              return
           case envelope := <- r.ch:
              // Extract this to a separate function because defer does not work correctly in a loop
              // BEGIN ==============================================================
              r.messagesMutex.Lock()   // TODO: define mutex in struct
              defer r.messagesMutex.Unlock()
              if _, ok := r.messages[envelope.PubsubTopic()]; !ok {
	         continue   // if extracted to a separate function, replace `continue` by `return`
	      }
              r.messages[envelope.PubsubTopic()] = append(r.messages[envelope.PubsubTopic()], envelope.Message())
              // END ===========================================
       }
    }
```

waku/v2/rpc/relay.go
```diff
@@ -44,6 +72,7 @@ func (r *RelayService) PostV1Subscription(req *http.Request, args *TopicsArgs, r
 			reply.Error = err.Error()
 			return nil
 		}
+		r.messages[topic] = make([]*pb.WakuMessage, 0)
```
richard-ramos, line 75: Shouldn't this happen before the `.Publish()`?, also lock r.messages
```suggestion
                r.messagesMutex.Lock()
                defer r.messagesMutex.Unlock()
		r.messages[topic] = make([]*pb.WakuMessage, 0)
```

waku/v2/rpc/relay.go
```diff
@@ -22,6 +26,30 @@ type TopicsArgs struct {
...
+		if _, ok := r.messages[envelope.PubsubTopic()]; !ok {
+			continue
+		}
+
+		r.messages[envelope.PubsubTopic()] = append(r.messages[envelope.PubsubTopic()], envelope.Message())
+	}
+}
+
+func (r *RelayService) Stop() {
+	r.node.Broadcaster().Unregister(r.ch)
```
richard-ramos, line 50:
```suggestion
	r.node.Broadcaster().Unregister(r.ch)
	close(r.ch)
```

waku/v2/rpc/relay.go
```diff
@@ -59,11 +88,19 @@ func (r *RelayService) DeleteV1Subscription(req *http.Request, args *TopicsArgs,
...
+		delete(r.messages, topic)
 	}
 	reply.Success = true
 	return nil
 }
 
-func (r *RelayService) GetV1Messages(req *http.Request, args *Empty, reply *Empty) error {
-	return fmt.Errorf("not implemented")
+func (r *RelayService) GetV1Messages(req *http.Request, args *TopicArgs, reply *MessagesReply) error {
+	if _, ok := r.messages[args.Topic]; !ok {
```
richard-ramos, line 99:
```suggestion
        r.messagesMutex.Lock()
        defer r.messagesMutex.Unlock()
	if _, ok := r.messages[args.Topic]; !ok {
```

richard-ramos (commented):
Good work!. I just left few comments

waku/v2/rpc/relay.go
alaibe: What if publish fail? I am adding at the point I am sure that the call will be successful if that makes sense

waku/v2/rpc/waku_rpc.go
alaibe: Agree, added

waku/v2/rpc/relay_test.go
alaibe: I added the init of ch in start

richard-ramos (approved):

PR #157, opened 2021-11-19
feat: Add rpc method for filter#getmessages

alaibe:


waku/v2/rpc/waku_rpc.go
```diff
@@ -72,15 +73,20 @@ func NewWakuRpc(node *node.WakuNode, address string, port int) *WakuRpc {
...
 func (r *WakuRpc) Start() {
 	go r.relayService.Start()
 	defer r.relayService.Stop()
+
+	go r.filterService.Start()
+	defer r.filterService.Stop()
+
 	log.Info("Rpc server started at ", r.server.Addr)
 	log.Info("server stopped ", r.server.ListenAndServe())
 }
```
richard-ramos, line 92: Should the `r.__Service.Stop()` functios be called in a separate `func(r *WakuRpc) Stop()` function that is invoked in node.go just before `wakuNode.Stop()` ?

alaibe: The start of the RPC server is trigger via go coroutine, this call:
```
 	log.Info("server stopped ", r.server.ListenAndServe()) 
``
Is going to block it until it is stopped, so it is working I think as expected but maybe I can make it better with a proper stop and replace the ListenAndServe with something non blocking. What do u prefer?

alaibe: @richard-ramos let me know if you prefer the new implementation

richard-ramos: Looks cool!

richard-ramos (approved):

PR #158, opened 2021-11-19
refactor: remove topic type

richard-ramos:
since go-libp2p uses string for topics

### DIFF
--- a/waku/node.go
+++ b/waku/node.go
@@ -232,7 +232,7 @@ func Execute(options Options) {
 
 	if !options.Relay.Disable {
 		for _, nodeTopic := range options.Relay.Topics {
-			_, err := wakuNode.Relay().SubscribeWithTopic(ctx, nodeTopic)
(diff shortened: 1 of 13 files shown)


waku/node.go
```diff
@@ -231,9 +231,8 @@ func Execute(options Options) {
 	}
 
 	if !options.Relay.Disable {
-		for _, t := range options.Relay.Topics {
-			nodeTopic := relay.Topic(t)
-			_, err := wakuNode.Relay().Subscribe(ctx, &nodeTopic)
+		for _, nodeTopic := range options.Relay.Topics {
+			_, err := wakuNode.Relay().SubscribeWithTopic(ctx, nodeTopic)
```
alaibe, line 235: nit: what about renaming SubscribeWithTopic to SubscribeToTopic

alaibe (approved):
Simple is better :) nice one.
In general you created function WithTopic, I think the ToTopic might read better.
WDYT

PR #164, opened 2021-11-22
feat: add NAT for DiscV5 UDP port

richard-ramos:


waku/v2/discv5/discover.go
```diff
@@ -184,13 +194,28 @@ func (d *DiscoveryV5) listen() error {
 		return err
 	}
 
+	d.udpAddr = conn.LocalAddr().(*net.UDPAddr)
+
+	if d.NAT != nil {
+		if !d.udpAddr.IP.IsLoopback() {
```
alaibe, line 200: nit: `if d.NAT != nil && !d.udpAddr.IP.IsLoopback() {`

alaibe (approved):

PR #173, opened 2021-12-06
feat: build swap and attach it to the store

alaibe:


waku/v2/node/wakunode2.go
```diff
@@ -241,7 +243,12 @@ func (w *WakuNode) checkForAddressChanges() {
 }
 
 func (w *WakuNode) Start() error {
-	w.store = store.NewWakuStore(w.host, w.opts.messageProvider, w.opts.maxMessages, w.opts.maxDuration)
+	w.swap = swap.NewWakuSwap([]swap.SwapOption{
```
richard-ramos, line 246: Just a question:
Would it be possible to create a store node without enabling Swap?

alaibe: If you build the store node manually, you can pass nil. If you use wakunode2 I didn't allow this option as by default, this is soft mode which does only accounting (inc/dec a counter). I don't think this would be problematic, let me know if you want this option or if you think this is problematic

richard-ramos (approved):

PR #184, opened 2022-01-18
feat: add WithLogger option to wakunode

richard-ramos:


waku/metrics/http.go
```diff
@@ -6,35 +6,38 @@ import (
 	"net/http"
 
 	"contrib.go.opencensus.io/exporter/prometheus"
-	logging "github.com/ipfs/go-log"
```
alaibe, line 9: :100: I wasn't sure why we were using log from ipfs

alaibe (commented):
Code looks good to me, could we just use zap logger and remove the go-ipfs logger dependency?

waku/metrics/http.go
alaibe: This comment was too quick! false joy

waku.go
```diff
@@ -17,11 +19,19 @@ func main() {
...
 
+	// for go-libp2p loggers
 	lvl, err := logging.LevelFromString(options.LogLevel)
 	if err != nil {
 		os.Exit(1)
 	}
 	logging.SetAllLoggers(lvl)
 
+	// go-waku logger
+	fmt.Println(options.LogLevel)
```
alaibe, line 30: leftover?

alaibe (approved):

PR #189, opened 2022-02-18
refactor: use int64 timestamps

richard-ramos:
Matches the changes done in these PRs:
- https://github.com/status-im/nim-waku/pull/842
- https://github.com/vacp2p/rfc/pull/483

However, it seems the test fleet has not been updated yet, because it still assumes the timestamp is float. Also, as pointed out in here https://github.com/vacp2p/rfc/pull/483#issuecomment-1044197233, the .proto for WakuMessages and HistoryQuery might need to be updated to deprecate the existing float fields and create new fields with the new data type.

So, this PR can't be merged until both issues are solved.


alaibe (approved):

richard-ramos:
Should be merged after https://github.com/status-im/nim-waku/pull/861 is merged

PR #192, opened 2022-02-18
chore: upgrade outdated dependencies

richard-ramos:
Since some of the dependencies require go 1.17, do not merge yet


alaibe (approved):

richard-ramos:
To avoid keeping this PR open forever, we're gonna have a branch for status-go (which uses 1.16), and only send fixes to that PR until we upgrade it to run on >= 1.17

PR #194, opened 2022-02-20
fix: flag naming inconsistency

### LINKED ISSUE #80: Inconsistent naming of flags compared to Nim-Waku
While working on the Ansible role for `go-waku` I found a few inconsistencies in flag naming compared to `nim-waku`:

* `--dbpath` instead of `--db-path`
* `--metrics*` instead of `--metrics-server*` (I think the shorter name is better tbh)
* `--port`  but no `--address`
* `--ws-port` but no `--ws-address`
* `--no-relay` instead of `--relay`
* `--store-node` instead of `--storenode` ~~(might be better changed in `nim-waku`)~~
* `--filter-node` instead of `--filternode` ~~(might be better changed in `nim-waku`)~~
* `--lightpush-node` instead of `--lightpushnode` ~~(might be better changed in `nim-waku`)~~

Not a huge issue, but it's good to keep things consistent, especially since we are implementing the same spec.

richard-ramos:
Fixes #80

Note: I had to change github.com/jessevdk/go-flags to github.com/urfave/cli/v2 since the former does not support setting a true/false value to boolean flags


jakubgs (approved):
Very nice, thank you.

PR #195, opened 2022-02-20
fix: remove references to fleets.status.im

### LINKED ISSUE #187: Remove references to fleets.status.im from the code
Please do not use https://fleets.status.im/ in the code:
https://github.com/status-im/go-waku/blob/11d1f8fb0d43a94e2409f3c18525b42880fb05b4/examples/chat2/main.go#L258-L259It is not intended for public use, but only as a convenience for developers.
The only reasonable way of using it is at build time to bake a list of static nodes into a build.

richard-ramos:
Fixes #187

### DIFF
--- a/examples/chat2/Makefile
+++ b/examples/chat2/Makefile
@@ -1,6 +1,18 @@
-.PHONY: all build
-
-build:
-	go build -o build/chat2 .
+.PHONY: all build fleets-remove fleets-update
 
 all: build
+
+FLEETS := fleets.json
+$(FLEETS):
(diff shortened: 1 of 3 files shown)


jakubgs (approved):
Great, thanks! Just to be clear, I don't want `fleets.status.im` to be used at runtime or by external users, but having a commited JSON from it in the repo like `status-go` for examples is fine if it's used at build time. I just don't want users and applications to use it at runtime.

richard-ramos:
I missunderstood the issue. I'll change the code back so it reads the fleets.json from filesystem instead.

richard-ramos:
oops. I closed it by mistake

PR #200, opened 2022-03-02
feat: add pubsub topic to Index pb

richard-ramos:


waku/persistence/store.go
```diff
@@ -92,13 +92,14 @@ func NewDBStore(log *zap.SugaredLogger, options ...DBOption) (*DBStore, error) {
 
 func (d *DBStore) createTable() error {
 	sqlStmt := `CREATE TABLE IF NOT EXISTS message (
-		id BLOB PRIMARY KEY,
+		id BLOB,
```
alaibe, line 95: humm, curious here, what is the value of an ID which is not an ID?

e.g how can 2 messages have send id?

richard-ramos: the `id` here should probably be renamed to `digest` or something. It's a hash of a marshalled WakuMessage. You could receive the same WakuMessage in different pubsub topics, hence the change that introduces a new constraint. I kept the name `id` to match nim-waku

alaibe: Gotcha knowing what is the id it makes sense. Maybe add a little comment ;)

alaibe (approved):

PR #210, opened 2022-03-15
Create pluggable store

neekolas:
## Problem
The current `WakuNode` implementation references the store as a concrete struct type, rather than an interface. This makes it difficult to replace with an alternative store implementation.

Since the store is initialized inside the the `Start()` function of the `WakuNode`, it is even more difficult to replace. While there is a `MessageProvider` interface that can be replaced, this assumes that a store implementation will want to keep a full list of all messages in memory. That drastically limits the capacity of the store.

## Solution
- Create a `Store` interface type that can be used on the `WakuNode` struct, representing the public methods of the store.
- Introduce the concept of a `storeFactory`, which is a function receiving a `WakuNode` instance and returning a `Store` interface.
- Create a default `storeFactory` that replicates the existing store creation behaviour
- Create a new parameter allowing the default `storeFactory` to be replaced.

### DIFF
--- a/waku/v2/node/wakunode2.go
+++ b/waku/v2/node/wakunode2.go
@@ -44,6 +44,8 @@ type Peer struct {
 	Connected bool
 }
 
+type storeFactory func(w *WakuNode) store.Store
+
 type WakuNode struct {
 	host host.Host
 	opts *WakuNodeParameters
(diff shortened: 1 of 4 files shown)


richard-ramos:
This is a very nice contribution! thank you! looks very useful in case anyone wants to use a different store!

richard-ramos (approved):
:100: 

The error on lint has already been fixed in master, so I think we can merge this!

PR #243, opened 2022-05-25
refactor: use db for serving history queries

richard-ramos:


dbutils/insert.go
```diff
@@ -70,6 +74,11 @@ func main() {
 		fmt.Println("Inserting ", N, " records in ", dbName)
 
 		db, err := newdb(dbName)
+		if err != nil {
+			panic(err)
```
alaibe, line 78: why not returning err and handling it upward rather than a panic?

waku/persistence/store.go
```diff
@@ -78,7 +93,31 @@ func NewDBStore(log *zap.SugaredLogger, options ...DBOption) (*DBStore, error) {
...
+	// readers do not block writers and faster i/o operations
+	// https://www.sqlite.org/draft/wal.html
+	// must be set after db is encrypted
+	var mode string
+	err = result.db.QueryRow("PRAGMA journal_mode=WAL").Scan(&mode)
+	if err != nil {
+		return nil, err
+	}
+	if mode != WALMode && file != "" {
+		fmt.Println(mode)
```
alaibe, line 116: use logger?

alaibe (approved):
Awesome, would you consider an orm, what we do with db is simple but error prone, the ORM would prevent use to do simple mistake.

dbutils/insert.go
richard-ramos: Since this is a simple program to generate DB files with 10K, 100K and 1M message records for perfomance testing, I think it's okay to just kill the program on failure. Error handling is not really necessary here (also all this code is running on main())

PR #242, opened 2022-05-25
Structured logging

mkobetic:
This is a somewhat opinionated proposal and I sincerely hope it won't come over the wrong way. I'm putting it forward as an open set of suggestions that can be adopted or rejected in part or as a whole. The burden is on me to provide sufficiently convincing argument for adoption.

To provide a bit of background, we (XMTP Labs) are preparing to run a network of go-waku nodes with production level SLOs. Observability is a critical prerequisite and logs are central part of that. We will possibly want to augment specific entries to suite our needs, but there are some fundamental aspects that would be really helpful to resolve here so that we're not pulling in completely opposite directions.

At the heart of it is adoption of structured logging. There are many ways to go about that, I tried to summarize the gist of the proposal in the included style guide. The guide could be preserved for future reference, but I mainly hope it will make it easier to comment on specific aspects of the proposal.

Central piece of the changes is merciless eradication of the SugaredLogger with a vengeance. Current change set does that in the core packages, leaving the various protocol packages alone for now, but I'd be happy to continue with this quest if that is what we end up agreeing on. The hope is that the current changes give some sense of how the logging changes would look in practice. I'll see if there are any specific bits worth pointing out inline.

PR #248, opened 2022-05-27
Structured logging followup

mkobetic:
This is continuation of #242 where I promised to finish eradicating SugaredLogger. The changes are sprawling through many files but are largely unsurprising. I tried to preserve the status quo as much as possible with few exceptions that I'll highlight inline. I tweaked the log messages to be more in line with the proposed style. I've added few more helpers for values of interest, like *enode.Node or IP addres/port combinations.

The cleanup of the various subsystems is split into separate commits, so we could pick and chose if desired (with some additional effort).

### DIFF
--- a/waku/v2/discv5/discover.go
+++ b/waku/v2/discv5/discover.go
@@ -215,7 +215,7 @@ func (d *DiscoveryV5) listen() error {
 	d.log.Info("started Discovery V5",
 		zap.Stringer("listening", d.udpAddr),
 		logging.TCPAddr("advertising", d.localnode.Node().IP(), d.localnode.Node().TCP()))
-	d.log.Info("Discovery V5: discoverable ENR ", logging.ENode("node", d.localnode.Node()))
+	d.log.Info("Discovery V5: discoverable ENR ", logging.ENode("enr", d.localnode.Node()))
 
 	return nil
 }
@@ -273,7 +273,7 @@ func (d *DiscoveryV5) UpdateAddr(addr *net.TCPAddr) error {
 	d.localnode.SetStaticIP(addr.IP)
 	d.localnode.Set(enr.TCP(uint16(addr.Port))) // lgtm [go/incorrect-integer-conversion]
 	d.log.Info("updated Discovery V5 node address", logging.TCPAddr("address", d.localnode.Node().IP(), d.localnode.Node().TCP()))
-	d.log.Info("Discovery V5", logging.ENode("node", d.localnode.Node()))
+	d.log.Info("Discovery V5", logging.ENode("enr", d.localnode.Node()))
 
 	return nil
 }
@@ -300,7 +300,7 @@ func hasTCPPort(node *enode.Node) bool {
 	enrTCP := new(enr.TCP)
 	if err := node.Record().Load(enr.WithEntry(enrTCP.ENRKey(), enrTCP)); err != nil {
 		if !enr.IsNotFound(err) {
-			utils.Logger().Named("discv5").Error("retrieving port for enr", logging.ENode("node", node))
+			utils.Logger().Named("discv5").Error("retrieving port for enr", logging.ENode("enr", node))
 		}
 		return false
 	}
@@ -321,7 +321,7 @@ func evaluateNode(node *enode.Node) bool {
 	_, err := utils.EnodeToPeerInfo(node)
 
 	if err != nil {
-		utils.Logger().Named("discv5").Error("obtaining peer info from enode", logging.ENode("node", node), zap.Error(err))
+		utils.Logger().Named("discv5").Error("obtaining peer info from enode", logging.ENode("enr", node), zap.Error(err))
 		return false
 	}
 
--- a/waku/v2/discv5/discover.go
+++ b/waku/v2/discv5/discover.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/ecdsa"
 	"errors"
-	"fmt"
 	"math"
 	"math/rand"
 	"net"
@@ -20,6 +19,7 @@ import (
 	"github.com/libp2p/go-libp2p-core/peer"
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/status-im/go-discover/discover"
+	"github.com/status-im/go-waku/logging"
 	"github.com/status-im/go-waku/waku/v2/utils"
 	"go.uber.org/zap"
 )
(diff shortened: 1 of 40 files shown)


waku/v2/protocol/filter/waku_filter.go
```diff
@@ -62,13 +62,13 @@ type (
 // relay protocol.
 const FilterID_v20beta1 = libp2pProtocol.ID("/vac/waku/filter/2.0.0-beta1")
 
-func NewWakuFilter(ctx context.Context, host host.Host, isFullNode bool, log *zap.SugaredLogger, opts ...Option) (*WakuFilter, error) {
+func NewWakuFilter(ctx context.Context, host host.Host, isFullNode bool, log *zap.Logger, opts ...Option) (*WakuFilter, error) {
 	wf := new(WakuFilter)
-	wf.log = log.Named("filter")
+	wf.log = log.Named("filter").With(zap.Bool("fullNode", isFullNode))
```
mkobetic, line 67: The `fullNode` attribute is meant to replace the hints in the various log entries emphasizing full vs light node cases. This attribute should now be present on all entries consistently.

waku/v2/protocol/filter/waku_filter.go
```diff
@@ -92,29 +92,25 @@ func NewWakuFilter(ctx context.Context, host host.Host, isFullNode bool, log *za
...
 	reader := protoio.NewDelimitedReader(s, math.MaxInt32)
 
 	err := reader.ReadMsg(filterRPCRequest)
 	if err != nil {
-		wf.log.Error("error reading request", err)
+		logger.Error("reading request", zap.Error(err))
 		return
 	}
 
-	wf.log.Info(fmt.Sprintf("%s: received request from %s", s.Conn().LocalPeer(), s.Conn().RemotePeer()))
```
mkobetic, line 117: The remote `peer` attribute is attached to the logger above, the local `node` peer ID should be attached to the central logger from the start, so all entries should have it.

waku/v2/protocol/lightpush/waku_lightpush.go
```diff
@@ -66,23 +66,23 @@ func (wakuLp *WakuLightPush) IsClientOnly() bool {
 
 func (wakuLP *WakuLightPush) onRequest(s network.Stream) {
 	defer s.Close()
-
+	logger := wakuLP.log.With(logging.HostID("peer", s.Conn().RemotePeer()))
```
mkobetic, line 69: I've used this patter in most of the `onRequest` methods, just attach the peer ID to the logger right at the top and then you get it consistently in all the following log entries.

waku/v2/protocol/store/waku_store.go
```diff
@@ -328,8 +330,6 @@ func (store *WakuStore) fetchDBRecords(ctx context.Context) {
 	}
 
 	metrics.RecordMessage(ctx, "stored", store.messageQueue.Length())
-
-	store.log.Info(fmt.Sprintf("%d messages available in waku store", store.messageQueue.Length()))
```
mkobetic, line 332: I merged this with the entry from the defered function above, just added the message count there as a field. Subtle difference is that now you get `messages=0` if the function fails out in the middle, but I thought that's fine.

waku/v2/protocol/store/waku_store.go
```diff
@@ -374,31 +374,35 @@ func (store *WakuStore) storeIncomingMessages(ctx context.Context) {
...
 	err := reader.ReadMsg(historyRPCRequest)
 	if err != nil {
-		store.log.Error("error reading request", err)
+		logger.Error("reading request", zap.Error(err))
 		metrics.RecordStoreError(store.ctx, "decodeRPCFailure")
 		return
 	}
-
-	store.log.Info(fmt.Sprintf("%s: Received query from %s", s.Conn().LocalPeer(), s.Conn().RemotePeer()))
+	logger = logger.With(zap.String("id", historyRPCRequest.RequestId))
```
mkobetic, line 389: This is the one that I embellished a bit with some additional parameters that we found useful in our fork. It at least shows how to use the history query field helpers. Happy to rip it out if you don't find it useful, but figured I can at least show how it was meant to be used.

waku/v2/protocol/store/waku_store.go
```diff
@@ -374,31 +374,35 @@ func (store *WakuStore) storeIncomingMessages(ctx context.Context) {
...
 
 	historyResponseRPC := &pb.HistoryRPC{}
 	historyResponseRPC.RequestId = historyRPCRequest.RequestId
 	historyResponseRPC.Response = store.FindMessages(historyRPCRequest.Query)
 
+	logger = logger.With(zap.Int("messages", len(historyResponseRPC.Response.Messages)))
 	err = writer.WriteMsg(historyResponseRPC)
 	if err != nil {
-		store.log.Error("error writing response", err)
+		logger.Error("writing response", zap.Error(err), logging.PagingInfo(historyResponseRPC.Response.PagingInfo))
```
mkobetic, line 402: The paging info is a somewhat complex value, but was intended to help with debugging query results. In our fork we only include it in debug mode (I didn't include that facility in my proposal, since that was a bit more obscure bit), here it would be included with all response write errors, which may not be a great fit. Again happy to drop it, just let me know.

waku/v2/discv5/discover.go
```diff
@@ -212,8 +212,10 @@ func (d *DiscoveryV5) listen() error {
 
 	d.listener = listener
 
-	d.log.Info(fmt.Sprintf("Started Discovery V5 at %s:%d, advertising IP: %s:%d", d.udpAddr.IP, d.udpAddr.Port, d.localnode.Node().IP(), d.localnode.Node().TCP()))
-	d.log.Info("Discv5: discoverable ENR ", d.localnode.Node())
+	d.log.Info("started Discovery V5",
+		zap.Stringer("listening", d.udpAddr),
+		logging.TCPAddr("advertising", d.localnode.Node().IP(), d.localnode.Node().TCP()))
+	d.log.Info("Discovery V5: discoverable ENR ", logging.ENode("node", d.localnode.Node()))
```
mkobetic, line 218: Calling the `ENode` attribute `node` is not good, since we're using it for the peer ID of the local node. But I'm unsure what to call the field. Would `enr` or `nodeRecord` or something be suitable? There are few of these sprinkled around but are easy to find (`logging.Enode(`).

richard-ramos: either `enr` or `nodeRecord` is fine!

mkobetic: ok, `enr` it is then, I'll make the change shortly.

mkobetic: Done

PR #253, opened 2022-06-01
logging: fix timestamp assumption micro -> nano

### LINKED ISSUE #251: logging: waku timestamp precision
In #242 I have probably incorrectly concluded that ["waku timestamp is a microsecond unix time"](https://github.com/status-im/go-waku/blob/master/logging/logging.go#L48-L58). In reality it seems that it may have been true with Waku but Waku2 seems to be  moving to [nanosecond precision](https://rfc.vac.dev/spec/14/#wakumessage).

I made my conclusion based on what I observed in our test suite originally where we were using an older version of js-waku running against a go-waku node. It seems js-waku has since [implemented store 2.0.0 beta4](https://github.com/status-im/js-waku/pull/566) which changed the precision of the sent timestamp that I was logging to nano-seconds.

So, the logging helper is likely wrong and should change to nanoseconds going forward, although it's not entirely clear how much of a residual mismatch should we expect while the various parts of the waku ecosystem are being upgraded to waku2. We can certainly make the helper more defensive and have it check whether the value looks like a nanosecond value or microsecond value and do the right thing. That would however hide the discrepancy, so maybe we want to see the invalid timestamps in the logs so that they can be hunted down and fixed.

I'm happy to make the PR, just need some guidance as to how to approach this.

mkobetic:
The logging helpers wrongly assumed timestamps to be microsecond unix time instead of nanosecond.

Fixes #251

### DIFF
--- a/logging/logging.go
+++ b/logging/logging.go
@@ -45,7 +45,7 @@ func HostID(key string, id peer.ID) zapcore.Field {
 
 func (id hostID) String() string { return peer.Encode(peer.ID(id)) }
 
-// Time - Waku uses Microsecond Unix Time
+// Time - Waku uses Nanosecond Unix Time
 type timestamp int64
 
 // Time creates a field for Waku time value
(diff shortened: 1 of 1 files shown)


richard-ramos:
Thank you!

PR #262, opened 2022-06-14
Fix filter deadlock

mkobetic:
status-im#260 introduced a subtle recursive read lock issue. Even though `SubscriptionHasContentTopic` runs in a different goroutine than the goroutine created bye `Items()`, that's why an RLock was needed there, it creates a deadlock in its interaction with `onRequest` that attempts to acquire a full lock when adding a subscriber. This creates the situation [described in the docs](https://pkg.go.dev/sync#RWMutex):

> If a goroutine holds a RWMutex for reading and another goroutine might call Lock, no goroutine should expect to be able to acquire a read lock until the initial read lock is released. In particular, this prohibits recursive read locking. This is to ensure that the lock eventually becomes available; a blocked Lock call excludes new readers from acquiring the lock.

So Items() grabs the RLock, an incoming request attempts to grab full Lock which makes subsequent RLock calls block until Items() completes. So SubscriptionHasContentTopic starts blocking and prevents Items() from completing (since they are synchronized through an unbuffered channel).

The solution here is what I proposed in the original PR as [maybe a better fix](https://github.com/status-im/go-waku/pull/260#discussion_r893739405). The idea is to extend `Items()` to do the topic filtering for us, that eliminates the need to iterate over subscription filters and consequently the need for RLock in FilterListener. The resulting behavior should be the same with one small caveat that I'll point out inline.

### DIFF
--- a/waku/v2/protocol/filter/filter_subscribers.go
+++ b/waku/v2/protocol/filter/filter_subscribers.go
@@ -14,6 +14,15 @@ type Subscriber struct {
 	filter    pb.FilterRequest // @TODO MAKE THIS A SEQUENCE AGAIN?
 }
 
+func (sub Subscriber) HasContentTopic(topic string) bool {
+	for _, filter := range sub.filter.ContentFilters {
+		if filter.ContentTopic == topic {
+			return true
+		}
+	}
+	return false
+}
+
 type Subscribers struct {
 	sync.RWMutex
 	subscribers []Subscriber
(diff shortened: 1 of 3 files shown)


waku/v2/protocol/filter/waku_filter.go
```diff
@@ -189,7 +189,7 @@ func (wf *WakuFilter) FilterListener() {
 		g := new(errgroup.Group)
 		// Each subscriber is a light node that earlier on invoked
 		// a FilterRequest on this node
-		for subscriber := range wf.subscribers.Items() {
+		for subscriber := range wf.subscribers.Items(&(msg.ContentTopic)) {
 			logger := logger.With(logging.HostID("subscriber", subscriber.peer))
 			subscriber := subscriber // https://golang.org/doc/faq#closures_and_goroutines
 			if subscriber.filter.Topic != "" && subscriber.filter.Topic != topic {
```
mkobetic, line 195: Note that this check will now run only for subscribers that match the content topic, instead for all subscribers. This seems like a defensive check and we're only potentially losing a warning for an unrelated subscription, which would presumably pop up in some other iteration where the content topic matches. That seems acceptable to me.

waku/v2/protocol/filter/filter_subscribers.go
```diff
@@ -36,25 +45,16 @@ func (sub *Subscribers) Append(s Subscriber) int {
...
-	for _, filter := range sub.filter.ContentFilters {
-		if filter.ContentTopic == topic {
-			return true
-		}
-	}
-	return false
-}
-
-func (sub *Subscribers) Items() <-chan Subscriber {
+func (sub *Subscribers) Items(topic *string) <-chan Subscriber {
```
mkobetic, line 48: A pointer to string so that nil can indicate no filtering.

richard-ramos (approved):
:medal_military: , thank you for fixing this

mkobetic:
❤️ FWIW we've already deployed this to our network and things look better.

PR #268, opened 2022-07-13
Fix logging.From panic

mkobetic:
The type assertion in `logging.From` blows up if the `Context` does not have a logger. This PR fixes the assertion and returns nil if there's no logger. Added tests to confirm the fix.


richard-ramos (approved):
LGTM, Thank you!

PR #269, opened 2022-07-15
Fix store error metrics tagging

snormore:
Register view with tag key `ErrorType` instead of `KeyType`, since `ErrorType` is what the metric is [emitted with](https://github.com/status-im/go-waku/blob/master/waku/v2/metrics/metrics.go#L88-L92).


richard-ramos (approved):
Thank you!

PR #276, opened 2022-07-26
Fix unsubscribe from missing content topic

neekolas:
## Summary
We encountered an issue when trying to unsubscribe from a topic that was not already subscribed to. The current implementation will panic with an Index Out Of Range error in that case. This updates the logic to be safer, and simplifies the implementation.

Also adds a few tests specifically for the `filter_subscribers` to ensure no regressions on this issue.


waku/v2/protocol/filter/filter_subscribers.go
```diff
@@ -120,22 +120,23 @@ func (sub *Subscribers) RemoveContentFilters(peerID peer.ID, contentFilters []*p
...
-			for _, cf := range subCfs {
+			for i, cf := range subCfs {
 				if cf.ContentTopic == contentFilter.ContentTopic {
 					l := len(subCfs) - 1
-					subCfs[l], subCfs[i] = subCfs[i], subCfs[l]
+					subCfs[i] = subCfs[l]
 					subscriber.filter.ContentFilters = subCfs[:l]
 				}
 			}
+			sub.subscribers[subIndex] = subscriber
```
neekolas, line 139: This is to deal with the fact that `Subscribers.subscribers` is a list of concrete structs and not pointers, and updates were being lost.

richard-ramos (approved):
good catch! thank you!

PR #279, opened 2022-07-28
Clean up ENR localNode DB

snormore:
Close ENR localnode DB during node tear down and address change, otherwise memory usage accrues pretty quickly if you're creating and tearing down a bunch of nodes in process.


snormore:
Seeing some `close of closed channel` panics with this change, going to close the PR until it's ready

PR #280, opened 2022-07-29
Verify matching `requestId` before removing subscription

elisealix22:
### Summary

This PR fixes the `unsubscribe` function to only remove subscriptions for topics that have a matching `requestId`. This resolves an issue where subscribing to the same content topic in two requests (request id 1 & 2) and calling unsubscribe on request id 2 would also unsubscribe request id 1.

Now a client can subscribe to the same content topic in two requests (request id 1 & 2), unsubscribe from one (request id 2) and request id 1 will stay subscribed.


richard-ramos (approved):
Thank you!

PR #286, opened 2022-08-10
Fix relay goroutine leaks

snormore:
This PR fixes a couple goroutine leaks in the relay, which were causing unbounded memory growth in processes that create/destroy client nodes over time:
1. Allow pubsub `.Next(ctx)` to exit by passing it a cancellable context that's cancelled during relay `Stop()`
2. Allow relay broadcaster `WaitUnregister` and `Unregister` to exit by checking to see if the broadcaster has been closed already, along with the goroutine that would consume from the blocking channel that these 2 methods send to

The PR also adds a `WithoutWakuRelay` node option to support a node client without relay enabled, since `WithWakuRelay` is given to the set of `DefaultWakuNodeOptions`.

Note that there's also a goroutine leak in the libp2p NAT service enabled via [libp2p.EnableNATService()](https://github.com/status-im/go-waku/blob/master/waku/v2/node/wakuoptions.go#L420), but this PR does not address that.

## Before 
goroutine

## After disabling libp2p NAT service

goroutine

## After (1)

goroutine

## After (2)

goroutine

### DIFF
--- a/waku/v2/broadcast.go
+++ b/waku/v2/broadcast.go
@@ -24,6 +24,8 @@ type broadcaster struct {
 
 	outputs         broadcastOutputs
 	outputsPerTopic map[string]broadcastOutputs
+
+	closed bool
 }
 
 // The Broadcaster interface describes the main entry points to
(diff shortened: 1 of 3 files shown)


snormore:
I've discovered a possible issue with this. Going to close to avoid it being merged until I can figure that out.

PR #297, opened 2022-08-25
feat: add flag to set wss port

### LINKED ISSUE #275: Add --wss-port to set an specific port for secure websockets
cc: @jakubgs

richard-ramos:
Fixes #275 
Adds the flag `--websocket-secure-port` and its alias `--wss-port` to allow setting up different ports for websockets and secure websockets.


richard-ramos:
@jm-clius @fryorcraken 
We should probably add the same flag to nwaku, so we can have separate ports for ws and wss!

jakubgs (approved):
Thanks.

PR #302, opened 2022-09-07
feat: zerokit

richard-ramos:


richard-ramos:
Consider looking at this too: https://github.com/wasmerio/wasmer-go  (maybe there's support for mobile devices?)

PR #315, opened 2022-09-15
chore: build .rpm with jenkins

richard-ramos:


richard-ramos:
Not needed anymore, as we're using github actions for building linux packages

PR #348, opened 2022-10-26
chore: set userAgent to `go-waku`

richard-ramos:


alrevuelta:
Great stuff! Didn't mean to be picky with "Go Waku v2 node" vs "go-waku", had no idea that you were already providing a userAgent different from the default one (was not the case in nwaku)

Thanks!

alrevuelta:
Can we add a cli flag to configure it? In case someone doesn't want to reveal the client?

richard-ramos:
> Can we add a cli flag to configure it? In case someone doesn't want to reveal the client?

Sure! will keep the PR open in the meantime, and once we have a name for the flag, will update it to include it.
Thank you!

alrevuelta:
> Sure! will keep the PR open in the meantime, and once we have a name for the flag, will update it to include it.

Sure, I had in mind `agent-string` on nwaku side, to follow nimbus pattern. But no strong opinion on its name.

PR #367, opened 2022-11-22
Enable Hole Punching

rasom:


richard-ramos:
Closing as this option was added in status-go

Issue #363, opened 2022-11-15
Incorrect Usage. flag provided but not defined: -store-message-db-url


fryorcraken:
go-waku v0.2.2 does not support the `--store-message-db-url` for database like nwaku

Note I moved js-waku's CI to use `--store-message-db-url=sqlite://:memory:` with nwaku v0.13.0

Can the same argument be supported in go-waku?

PR #374, opened 2022-11-25
refactor: store

richard-ramos:


richard-ramos:
Note: include https://github.com/waku-org/go-waku/issues/363 as part of this PR

PR #400, opened 2022-12-19
feat: add discV5 to c-bindings

richard-ramos:
cc: @danielSanchezQ

### DIFF
--- a/mobile/api_discovery.go
+++ b/mobile/api_discovery.go
@@ -2,6 +2,7 @@ package gowaku
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/waku-org/go-waku/waku/v2/dnsdisc"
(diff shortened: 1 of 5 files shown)


mobile/api_discovery.go
```diff
@@ -37,3 +38,25 @@ func DnsDiscovery(url string, nameserver string, ms int) string {
...
+}
+
+func StopDiscoveryV5() string {
+	if wakuNode == nil {
+		return MakeJSONResponse(errWakuNodeNotReady)
+	}
+	if wakuNode.DiscV5() == nil {
+		return MakeJSONResponse(errors.New("DiscV5 is not mounted"))
+	}
+	err := wakuNode.DiscV5().Start(context.Background())
```
bacv, line 60: Shouldn't this be `DiscV5().Stop()`?

richard-ramos: :facepalm: yes

richard-ramos: https://github.com/waku-org/go-waku/pull/402
here's the fix

PR #413, opened 2023-01-04
fix(c-bindings): simplify discv5 API

richard-ramos:
Removes the start and stop function for discV5 from the c-bindings


danielSanchezQ (approved):
🎸

PR #418, opened 2023-01-11
feat: use backedoff connector and discovery

richard-ramos:
Uses exponential backoff for connecting to peers and use pointer to enode to simplify code

### DIFF
--- a/waku/v2/discv5/discover.go
+++ b/waku/v2/discv5/discover.go
@@ -52,7 +52,7 @@ type peerCache struct {
 type PeerRecord struct {
 	expire int64
 	Peer   peer.AddrInfo
-	Node   enode.Node
+	Node   *enode.Node
 }
 
 type discV5Parameters struct {
@@ -126,10 +126,6 @@ func NewDiscoveryV5(host host.Host, priv *ecdsa.PrivateKey, localnode *enode.Loc
 		config: discover.Config{
 			PrivateKey: priv,
 			Bootnodes:  params.bootnodes,
-			ValidNodeFn: func(n enode.Node) bool {
-				// TODO: track https://github.com/status-im/nim-waku/issues/770 for improvements over validation func
-				return evaluateNode(&n)
-			},
 			V5Config: discover.V5Config{
 				ProtocolID: &protocolID,
 			},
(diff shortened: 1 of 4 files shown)


waku/v2/discv5/discover.go
```diff
@@ -291,7 +287,7 @@ func (d *DiscoveryV5) iterate(ctx context.Context, iterator enode.Iterator, limi
 
 	for {
 		if len(d.peerCache.recs) >= limit {
-			break
+			time.Sleep(1 * time.Minute)
```
rasom, line 290: I'm not very familiar with the code, thus question:
previously we would break the cycle and now we just delay execution, what's the point?

waku/v2/protocol/peer_exchange/waku_peer_exchange.go
```diff
@@ -122,6 +135,13 @@ func (wakuPX *WakuPeerExchange) handleResponse(ctx context.Context, response *pb
 
 	if len(peers) != 0 {
 		log.Info("connecting to newly discovered peers", zap.Int("count", len(peers)))
+
+		ch := make(chan peer.AddrInfo, len(peers))
+		for _, p := range peers {
+			ch <- p
+		}
+
+		wakuPX.connector.Connect(ctx, ch)
```
rasom, line 144: again, it is likely I'm missing something:
here we connect to peers, but we do not remove any old code which would do the same thing. Did we not connect to peers before? It is likely a stupid question

waku/v2/discv5/discover.go
richard-ramos: Previously, if we reached the maximum number of discovered nodes (600), we would stop discovering new peers, and exit the goroutine.  Calling `FindPeers` would still work for 1hr since peers would still be in the cache, however, after 1 hour the peers would be removed from the cache, and no peers would be discovered since the iterate() function would never be executed

waku/v2/protocol/peer_exchange/waku_peer_exchange.go
richard-ramos: Oops. Indeed. I forgot to remove lines 145 to 154.
Good catch

rasom (approved):

waku/v2/discv5/discover.go
rasom: @richard-ramos btw does this affect this issue https://github.com/status-im/status-go/issues/3063 ? Does it mean that there is no need to restart discovery as it is not stopped now?

richard-ramos: I think it does not. It would help once you discover 600 peers, since discovery would not stop. I have some code in progress that would probably help with 3063

PR #432, opened 2023-01-30
fix: add lock when deleting peer

alaibe:


caybro (approved):
Fixes the crash for me

saledjenic (approved):

PR #440, opened 2023-02-08
feat: add WakuMessage validation in gossipsub

richard-ramos:
This also stores the waku message in a cache to avoid having to decode it twice.
RLN validator is also modified to store the messages in the cache
Matches https://github.com/waku-org/nwaku/pull/1537

### DIFF
--- a/waku/v2/node/wakunode2.go
+++ b/waku/v2/node/wakunode2.go
@@ -408,9 +408,9 @@ func (w *WakuNode) Stop() {
 
 	w.host.Close()
 
-	close(w.enrChangeCh)
-
 	w.wg.Wait()
+
+	close(w.enrChangeCh)
 }
 
 // Host returns the libp2p Host used by the WakuNode
--- a/waku/v2/node/wakunode2_test.go
+++ b/waku/v2/node/wakunode2_test.go
@@ -157,6 +157,7 @@ func Test500(t *testing.T) {
 			if err := wakuNode2.Publish(ctx, msg); err != nil {
 				require.Fail(t, "Could not publish all messages")
 			}
+			time.Sleep(5 * time.Millisecond)
 		}
 	}()
 
--- a/waku/v2/node/wakunode2_test.go
+++ b/waku/v2/node/wakunode2_test.go
@@ -59,8 +59,8 @@ func int2Bytes(i int) []byte {
 	return append(big.NewInt(int64(i)).Bytes(), byte(0))
 }
 
-func Test5000(t *testing.T) {
-	maxMsgs := 5000
+func Test500(t *testing.T) {
+	maxMsgs := 500
 	maxMsgBytes := int2Bytes(maxMsgs)
 
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
--- a/waku/v2/protocol/relay/waku_relay.go
+++ b/waku/v2/protocol/relay/waku_relay.go
@@ -68,6 +68,7 @@ type WakuRelay struct {
 	deleteCacheCh chan string
 	getMsgCh      chan msgReq
 
+	ctx    context.Context
 	cancel context.CancelFunc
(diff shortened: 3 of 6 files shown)


alrevuelta (approved):
lgtm!

I would perhaps add some tests for the cache and the validator (perhaps similar to [this](https://github.com/waku-org/nwaku/blob/ddf5f97e540fbbe066d755621f9345cb71639706/tests/v2/test_wakunode_relay.nim#L211-L241)?). Mainly to protect the code from future changes.

Perhaps a nitpick, but as I see the cache has a size of `1000` right? afaik we should never get the same message twice (gossipsub will prevent this) so even if the cache size is just `1` we should always have a hit? Since what benefits from this cache is immediatly after the validation in the subscribe. But again, just a nitpick.

LNSD:
Hold on! Note my comment on the research issue: https://github.com/vacp2p/research/issues/164#issuecomment-1422920922

richard-ramos:
@LNSD:

> The Waku message integrity validator is only being applied to the default pub-sub topics ([here](https://github.com/waku-org/nwaku/blob/master/waku/v2/protocol/waku_relay.nim#L105-L113)) since the validator is only applied at the WakuRelay instance creation.
>   If any user subscribes to a pub-sub topic via the subscribe API ([here](https://github.com/waku-org/nwaku/blob/master/waku/v2/protocol/waku_relay.nim#L138-L151)) (e.g., via the JSON-RPC API), the message integrity validator won't be added to that topic, leaving those topics out of the claimed "DoS protection".

In go-waku registering a validator is done in [`subscribe` function](https://github.com/waku-org/go-waku/blob/498669a1f2194a1fb2f4fbf1d48f3539b658a3e4/waku/v2/protocol/relay/waku_relay.go#L241-L264) which is called each time a new subscription to a pubsub topic is done (not only on instance creation). The only special case is rln-relay which has its own validator hence why I [replace any validator set for the rln pubsub topic before setting the specific rln relay validator](https://github.com/waku-org/go-waku/blob/498669a1f2194a1fb2f4fbf1d48f3539b658a3e4/waku/v2/protocol/rln/waku_rln_relay.go#L410-L413) (although all validators share the same message cache)

>    Additionally, the message protobuf integrity validator uses the waku_message's decode function that deserializes the protobuf fields. The validation approach is inefficient (and potentially problematic) because the function performs a series of allocations due to the deserialization process. As they are executed on every message's arrival, validators must be fast and efficient to not impact the message latency and overall system performance.

Interesting, have you done some measurement on how many ns in latency are added for doing this decoding? I imagine this a problem too in nwaku RLN Relay. I see here that for RLN: https://github.com/waku-org/nwaku/blob/master/waku/v2/protocol/waku_rln_relay/utils.nim#L663-L719 besides doing the protobuffer decoding, we also validate a zkproof, and check if the message epoch is appropiate and if it's not duplicated (https://github.com/waku-org/nwaku/blob/605cf1c38cad83da31ee171e8d2f815c24efad51/waku/v2/protocol/waku_rln_relay/utils.nim#L384-L467). (Same operations happen in go-waku too). How will this be solved? or message transmission being inefficient is not a problem with rln-relay?

Curiously enough in eth2 some operations with validators are also done: https://github.com/prysmaticlabs/prysm/blob/77d3ccb9ad8b28656452456d2d25437f5fc53d88/beacon-chain/sync/subscriber.go#L163 and I see some mentions of decoding... I'll try to investigate later what techniques do they use

richard-ramos:
> @alrevuelta: Perhaps a nitpick, but as I see the cache has a size of `1000` right?

The 1000 is a hint to how much space will be initially allocated to hold the specified number of elements. If the number of items grows, then the cache will allocate more space. The 1000 does not limit the maximum number of items in the cache.

Later in the code i do track the deletions done in the cache to recreate the map. The reason is because go's maps [do not shrink when elements are deleted](https://github.com/golang/go/issues/20135)

PR #446, opened 2023-02-10
feat(c-bindings): expose store protocol

richard-ramos:
Allows an application using the c-bindings to act as a storenode

Adds the following config items:
```
- `store`: Enable store protocol to persist message history
  Default `false`
- `databaseURL`: url connection string. Accepts SQLite and PostgreSQL connection strings
  Default: `sqlite3://store.db`
- `storeRetentionMaxMessages`: max number of messages to store in the database.
  Default `10000`
- `storeRetentionTimeSeconds`: max number of seconds that a message will be persisted in the database.
  Default `2592000` (30d)
```

And a new function to retrieve data from local database:
```c
// Query historic messages stored in the localDB
// queryJSON must contain a valid json string with the following format:
//
//	{
//		"pubsubTopic": "...", // optional string
//	 "startTime": 1234, // optional, unix epoch time in nanoseconds
//	 "endTime": 1234, // optional, unix epoch time in nanoseconds
//	 "contentFilters": [ // optional
//			{
//		      contentTopic: "contentTopic1"
//			}, ...
//	 ],
//	 "pagingOptions": {// optional pagination information
//	     "pageSize": 40, // number
//			"cursor": { // optional
//				"digest": ...,
//				"receiverTime": ...,
//				"senderTime": ...,
//				"pubsubTopic" ...,
//	     }
//			"forward": true, // sort order
//	 }
//	}
//
// If a non empty cursor is returned, this function should be executed again, setting  the `cursor` attribute with the cursor returned in the response
// Requires the `store` option to be passed when setting up the initial configuration
//
extern char* waku_store_local_query(char* queryJSON);
```

### DIFF
--- a/library/README.md
+++ b/library/README.md
@@ -277,6 +277,10 @@ interface JsonConfig {
     discV5?: boolean;
     discV5BootstrapNodes?: Array<string>;
     discV5UDPPort?: number;
+    store?: boolean;
+    databaseURL?: string;
+    storeRetentionMaxMessages?: number;
+    storeRetentionTimeSeconds?: number;
 }
 ```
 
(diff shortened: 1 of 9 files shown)


danielSanchezQ:
Not to be added to this PR. But, what do you think of having extra calls so we can replicate this behaviour:

- If an in-memory db, being able to dump it entirely to disk.
- recover a disk db file and loading it up to memory

This will allow us to keep using a memory db but having the possibility un storing the state in case of failure for example.

Don't know if it is possible though, but it should somehow 🤔  What do you think @richard-ramos ?

danielSanchezQ (approved):
🎸

richard-ramos:
@danielSanchezQ
Should be possible, because these functions are available in sqlite: `sqlite3_backup_init`, `sqlite3_backup_step` and `sqlite3_backup_finish` https://www.sqlite.org/backup.html.  I'll have a look this week.

Issue #256, opened 2022-06-07
Build go-waku using Nix derivations


jakubgs:
As building different forms for different paltforms using Jenkins CI can be a pain because of need for extra tools like gomobile or extra resources like Android SDK and NDK, it might make sense to build everything in Nix.

As we already use Gomobil to build [status-go](https://github.com/status-im/status-go) in [status-react](https://github.com/status-im/status-react) it shouldn't be too difficult:
https://github.com/status-im/status-react/blob/develop/nix/status-go/mobile/build.nix

jakubgs:
I'm making progress doing an Android AAR build using `gomobile`, but I've stumbled on this error:
```
go: downloading golang.org/x/sync v0.0.0-20220513210516-0976fa681c29
go: downloading github.com/golang-migrate/migrate/v4 v4.15.2
# runtime/cgo
gcc_android.c:6:10: fatal error: android/log.h: No such file or directory
    6 | #include <android/log.h>
      |          ^~~~~~~~~~~~~~~
compilation terminated.
```
There are some references to this in Go repos:

* https://github.com/golang/go/issues/33859
* https://github.com/golang/go/issues/39549
* https://github.com/golang/go/issues/42725

jakubgs:
I ran `gomobile bind` via `strace -f` and this is what I found:
```
 > grep 'android/log.h' /nix/store/aqf11zrgdzh0i4zr8gvdzk5b6hpiwrix-go-waku-devel/strace.log 
535   openat(AT_FDCWD, "/nix/store/4mz4navg6bwavsfkwzpphbsm284ns087-go-1.17.10/share/go/src/runtime/cgo/android/log.h", O_RDONLY|O_NOCTTY) = -1 ENOENT (No such file or directory)
535   openat(AT_FDCWD, "/build/go-build2729256319/b082/android/log.h", O_RDONLY|O_NOCTTY) = -1 ENOENT (No such file or directory)
535   openat(AT_FDCWD, "/nix/store/j5vhrywqmz1ixwhsmmjjxa85fpwryzh0-gcc-11.3.0/lib/gcc/x86_64-unknown-linux-gnu/11.3.0/include/android/log.h", O_RDONLY|O_NOCTTY) = -1 ENOENT (No such file or directory)
[...]

PR #483, opened 2023-03-05
nix: add Nix flake to build node and library

jakubgs:
This way we can build stuff locally using:
```sh
nix build
nix build .#node
nix build .#library
```
Or just start a shell with Go `1.19.x` using:
```
nix develop
```
Which simply has the same environment as the build shell for the node.

One known snag is that there is currently no simple way to keep `vendorSha256` updated to match the contents of `go.mod` and `go.sum`. For more details see:
https://discourse.nixos.org/t/how-should-i-build-a-go-package-from-local-source/19490/8

One way around this would be to have our own `vendor` folder, but that's also a pretty ugly solution that requires manual updating.

Resolves:
* https://github.com/waku-org/go-waku/issues/256

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,6 +53,8 @@ Gowaku.xcframework
 # Icon must end with two \r
 Icon
 
+# Nix
+result
 
 # Thumbnails
 ._*
--- a/README.md
+++ b/README.md
@@ -21,6 +21,19 @@ make
 # See the available command line options with
 ./build/waku --help
 ```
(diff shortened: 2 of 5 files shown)


richard-ramos (approved):
Very cool!

jakubgs:
We still can't build mobile versions, but as I updated in the issue, this seems like a bigger problem.

And it's good to have this boilerplate in place anyway.

PR #498, opened 2023-04-02
refactor: setting nodeKey via GOWAKU-NODEKEY env

harsh-98:
Since issue https://github.com/urfave/cli/issues/1272 is closed, and go-waku is already using version greater than 2.14.1, extra env logic can be removed.


richard-ramos (approved):
Thank you! :100:

PR #502, opened 2023-04-04
refactor: fetch membership events

harsh-98:
- WatchMembershipRegistered func either returns nil, err or Subscription, nil. 
image

So, subs.Subscribe https://github.com/waku-org/go-waku/blob/master/waku/v2/protocol/rln/web3.go#L197 will fail when there is an err. subs will be nil.

- I think the purpose of `errCh` and `done` in watchNewEvents func is to get the first error while creating a subscription (like if the node doesn't support ws notifications). On such err, application should terminate. But  https://github.com/waku-org/go-waku/blob/master/waku/v2/protocol/rln/web3.go#L205 done might be closed before  `Resubscribe` creates the first subscription internally. Hence, the first subscription error might not be captured.

### DIFF
--- a/waku/node.go
+++ b/waku/node.go
@@ -112,7 +112,6 @@ func Execute(options Options) {
 		node.WithHostAddress(hostAddr),
 		node.WithKeepAlive(options.KeepAlive),
 	}
-
 	if len(options.AdvertiseAddresses) != 0 {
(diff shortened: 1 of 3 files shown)


richard-ramos (approved):
thank you!

PR #528, opened 2023-05-02
chore: add unit test for signed validator and --protected-topic-flag

richard-ramos:


waku/node.go
```diff
@@ -300,6 +300,11 @@ func Execute(options Options) {
 			failOnErr(err, "Error subscring to topic")
 			wakuNode.Broadcaster().Unregister(&nodeTopic, sub.C)
 		}
+
+		for _, protectedTopic := range options.Relay.ProtectedTopics {
```
alrevuelta, line 304: not sure if this is the right place, but can we add some logs informing that said topics are "protected"? I think its very important to inform on this, so that the user is aware.

Also, do we have any check that we are subscribed for the protected topics? I mean, it wouldn't make sense (perhaps signals a configuration error) that we have protected-topic "a" but we are not subscribed to that topic "a"

richard-ramos: Since protecting a topic makes no sense if you're not subscribed to it, what do you think of automatically subscribing any topic specified in `--protected-topic` flag?

alrevuelta: > Since protecting a topic makes no sense if you're not subscribed to it, what do you think of automatically subscribing any topic specified in --protected-topic flag?

Not sure its a good idea. imho this feature (protected topics) should be totally isolated from the rest and do only one thing, which is protecting a topic. Note also that this feature is not part of the protocol, but something more specific to the application layer, so imho its better to keep it separate. I think its better to be explicit and on one side, subcribe to the topic (one cli flag), and on the other side "protect" it (another cli flag, --protected-topic)

PR #535, opened 2023-05-04
feat: validate message timestamp in signed topic validator

richard-ramos:


waku/v2/protocol/relay/validators.go
```diff
@@ -13,19 +15,46 @@ import (
...
 	"go.uber.org/zap"
 	proto "google.golang.org/protobuf/proto"
 )
 
 // Application level message hash
 func MsgHash(pubSubTopic string, msg *pb.WakuMessage) []byte {
-	// TODO: Other fields?
-	return hash.SHA256([]byte(pubSubTopic), msg.Payload, []byte(msg.ContentTopic))
+	timestampBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(timestampBytes, uint64(msg.Timestamp))
```
alrevuelta, line 26: the test vector is working fine, but `Timestamp` is a `int64` not `uint64`.

alrevuelta (approved):
lgtm! minor comment

waku/v2/protocol/relay/validators.go
alrevuelta: well i guess it doesnt matter as long as the number is possitive (which would be). guess that a positive uint64 is represented in bytes just like an int64.

richard-ramos: `PutUint64` expects an `uint64` hence why I cast the timestamp to that type

PR #538, opened 2023-05-05
Safe broadcaster

harsh-98:
## New design:

The previous design of the broadcaster sometimes suffered from deadlocks. New broadcaster is thread-safe, lock-free, and guarantees closure of all channels. It only returns read-only channels. The new broadcaster is under relay folder.

- When the broadcaster is not set in waku_relay, then relay.SubscribeToTopic returns NoopSubscription, i.e. closed channel. With this, waku_relay doesn't maintain subscription channels. 
As I understand, the purpose of maintaining channels for a topic was to broadcast all messages from the same topic on all these channels(i.e. cloned channels for a topic). This can also be achieved by the broadcaster.

If the broadcaster is not set in waku_relay, then the waku_relay doesn't create pubsub.Subscription. Basically not much purpose of `relay` w/o the broadcaster (is this wrong)?
Now there is only one pubsub.Next loop, earlier there were multiple pubsub.Next loop per topic, the same functionality with less overhead.

## Test Changes(doubts):

I tried to make most of the test pass, without relying on the buffered channel from relay.SubscribeToTopic, check commit [e4a6a1d1](https://github.com/harsh-98/go-waku/commit/e4a6a1d15f5f5e96583bed61ed3b7f734ed545b7). 
Though some legacy filter and filter test relies on the buffered channel so added back buffered channel to relay.SubscribeToTopic in commit [aa02e745db](https://github.com/harsh-98/go-waku/commit/aa02e745dbb9573ef851f4d9cba9bb8391870941).(there is scope for future PRs :)

In the constraint environment, TestRelaySubscription in /waku/v2/rest was failing, as the node didn't have enough time to start, added some wait [939f1](https://github.com/harsh-98/go-waku/commit/939f1ed675e453f657fe13c6fd1a11ffff172029).

Question:
what to do if pubsub.subscription for a topic terminates in waku_relay? resubscribe to that topic, or don't do anything?
Currently, this is not handled. In this case, since relaySubs already has a pubsub.subscription entry, a new pubsub.subscription is not created on another call to relay.SubscribeToTopic. No new message will be received for this topic from pubsub.subscription(this is same as before).


richard-ramos (approved):
:medal_military: cool PR! thank you!

richard-ramos:
> What to do if pubsub.subscription for a topic terminates in waku_relay? resubscribe to that topic, or don't do anything?
Currently, this is not handled. In this case, since relaySubs already has a pubsub.subscription entry, a new pubsub.subscription is not created on another call to relay.SubscribeToTopic. No new message will be received for this topic from pubsub.subscription(this is same as before).

I think for now it should be okay for it to not do nothing. We can revisit this in the future if the need for resubscribing to a pubsub topic comes up!

harsh-98:
Thanks. One more point:
In new design,  if the broadcaster is not set in waku_relay, then the waku_relay doesn't create pubsub.Subscription. And the returned channel/subscription is also closed one. 

```
relay := NewWakuRelay(nil, 0, timesource.NewDefaultClock(), utils.Logger())
sub, err := relay.SubscribeToTopic(context.Background(), topic)
require.NoError(t, err)
sub // closed subscription, internally relaySubs has an created pubsub not that is not listened to via pubsub.Next.
```

PR #543, opened 2023-05-09
feat: use addresses in signed topic validator

richard-ramos:
(This PR is still missing the reference values from the RFC)


alrevuelta:
Unfortunately we wont be moving forward with that spec, so we can continue with the public key as it was before. Feel free to close :)

PR #550, opened 2023-05-15
refactor: add `--ext-ip` and `--dns4-domain-name` values to the list of node multiaddressess

richard-ramos:
This allows using both flags without one overriding the other


jakubgs (approved):
Thanks.

PR #555, opened 2023-05-19
chore: upgrade deps

richard-ramos:
For usage within status-go, we should rebase status-im/go-ethereum fork to use a go-ethereum version that is less than 6 months old. (Ideally 1.11.6, since it's 1 month old already)

cc: @cammellos @rasom @siphiuel @alaibe @saledjenic @jakubgs


go.mod
```diff
@@ -2,24 +2,21 @@ module github.com/waku-org/go-waku
...
-	github.com/ethereum/go-ethereum v1.10.26
+	github.com/ethereum/go-ethereum v1.11.6
 	github.com/golang-migrate/migrate/v4 v4.15.2
-	github.com/golang/protobuf v1.5.2 // indirect
+	github.com/golang/protobuf v1.5.3 // indirect
 	github.com/gorilla/rpc v1.2.0
 	github.com/ipfs/go-ds-sql v0.3.0
 	github.com/ipfs/go-log/v2 v2.5.1
-	github.com/libp2p/go-libp2p v0.26.4
+	github.com/libp2p/go-libp2p v0.27.3
```
chaitanyaprem, line 14: I am wondering if we should upgrade to latest version of 0.32 of go-libp2p. 
They seem to have added webTRansport support which allows browsers to connect without TLS certificates, which i remember seems to be a pain point for js-waku.
Some new features like smart dialing etc seem to have been introduced as well.

chaitanyaprem (approved):

go.mod
richard-ramos: Yes, we can do that on a separate PR. The main reason this PR hasn't been merged is because a lot of these dependencies make go-waku incompatible with status-go due to them using a very old version of go-ethereum.

But maybe there's some availability this week or the next for someone to take a look into that? :thinking: 
cc: @cammellos @iurimatias

jakubgs (approved):

chaitanyaprem:
@richard-ramos , I think this is outdated now since we have another PR for updating libp2p right. 
If so, let us close this one unless you think there is still some value in keeping this around.

richard-ramos:
yeah, let's close this

PR #577, opened 2023-06-22
feat(c-bindings): filterv2

richard-ramos:
- Renames existing `waku_filter_*` functions to `waku_legacy_filter_*`
- Adds FilterV2 functions `waku_filter_*`


library/api_legacy_filter.go
```diff
@@ -0,0 +1,50 @@
+package main
+
+import (
+	"C"
+
+	mobile "github.com/waku-org/go-waku/mobile"
+)
+
+// Creates a subscription to a light node matching a content filter and, optionally, a pubSub topic.
```
chaitanyaprem, line 9: Better to start marking legacyFilter API's as deprecated indicating that they would be removed in upcoming releases.

library/README.md
```diff
@@ -93,6 +93,23 @@ Fields:
...
+
+Fields:
+
+- `contentTopics`: Array of content topics.
+- `topic`: pubsub topic.
+
+
+### `LegacyFilterSubscription` type
+
+The criteria to create subscription to a light node in JSON Format:
```
chaitanyaprem, line 111: Confusing wording: The create subscription would be from a light node not to it. Wherever we are referring to creation of subscription, this can be modified to avoid confusion.

chaitanyaprem (commented):
Minor comment wrt documentation which confused me. 
Other than that, LGTM.

PR #582, opened 2023-06-23
refactor: use backoff strategy to connect to random rendezvous points

richard-ramos:


waku/v2/rendezvous/rendezvous.go
```diff
@@ -99,8 +111,40 @@ func (r *Rendezvous) Start(ctx context.Context) error {
...
+			dialableRP = append(dialableRP, rp)
+		}
+	}
+
+	result := make(chan *rendezvousPoint, 1)
+
+	if len(dialableRP) > 0 {
+		result <- r.rendezvousPoints[rand.Intn(len(r.rendezvousPoints))] // nolint: gosec
+	} else {
+		if len(r.rendezvousPoints) > 0 {
```
chaitanyaprem, line 128: If there is no dialable RP, choosing one randomly is fine if discover is done by Waku layer. 
Due to #581 , discovery triggers are moved to application layer..in that case, it would be possible for an application to spam the RP even if by mistake by constantly invoking discover. By having this backoff strategy, in a way we are preventing same RP not being dialed constantly.
Maybe, instead of choosing to dial if no dialable RPs are there, it would be better to return a wait-retry to the application based on which the app can redial again.

Another thought that comes to me, is do we even need to define this backoff strategy or just let application handle it?

richard-ramos: > Maybe, instead of choosing to dial if no dialable RPs are there, it would be better to return a wait-retry to the application based on which the app can redial again.

Hmm, not sure I follow. The code is set so when there is no dialable RP, it will then wait until the RP with the shortest wait time is available. Dialing will be only attempted when there is a RP 

> Another thought that comes to me, is do we even need to define this backoff strategy or just let application handle it?

Good point. I'm thinking that i could refactor this code to extracted part of the logic into a function that an application could use to avoid having to implement their own backoff strategy from scratch. What do you think?

chaitanyaprem: > > Maybe, instead of choosing to dial if no dialable RPs are there, it would be better to return a wait-retry to the application based on which the app can redial again.
> 
> Hmm, not sure I follow. The code is set so when there is no dialable RP, it will then wait until the RP with the shortest wait time is available. Dialing will be only attempted when there is a RP
> 
My bad, got confused when i was looking at diff only in github. When i looked in vsCode, i understood.
> > Another thought that comes to me, is do we even need to define this backoff strategy or just let application handle it?
> 
> Good point. I'm thinking that i could refactor this code to extracted part of the logic into a function that an application could use to avoid having to implement their own backoff strategy from scratch. What do you think?
Yeah, that's a good idea since anyways you have coded the backoff strategy logic.

waku/v2/rendezvous/rendezvous.go
```diff
@@ -162,23 +130,38 @@ func (r *Rendezvous) callRegister(ctx context.Context, rendezvousClient rvs.Rend
 	return t, retries
 }
 
-func (r *Rendezvous) Register(ctx context.Context, topic string) {
-	for _, m := range r.rendezvousPoints {
+func (r *Rendezvous) Register(ctx context.Context, rendezvousPoints []*RendezvousPoint) {
+	r.RegisterWithTopic(ctx, protocol.DefaultPubsubTopic().String(), rendezvousPoints)
+}
+
+func (r *Rendezvous) RegisterShard(ctx context.Context, cluster uint16, shard uint16, rendezvousPoints []*RendezvousPoint) {
```
chaitanyaprem, line 137: Improvement: Wondering if we should create a type called shard which contains cluster and shard, and abstracted out from all function calls. 
Later when we move to auto-sharding, then we don't need to change all functions..but just modify the type and add a function to it to compute the pubSubTopic.

chaitanyaprem (commented):
I could not figure out what are the new changes made after my previous review, probably due to a force push. 
Would be good to indicate changes in the commit message. It took more time for me to figure out during review. :)

waku/v2/rendezvous/iterator.go
```diff
@@ -0,0 +1,65 @@
...
+	return &RendezvousPointIterator{
+		rendezvousPoints: rendevousPoints,
+	}
+}
+
+func (r *RendezvousPointIterator) RendezvousPoints() []*RendezvousPoint {
+	return r.rendezvousPoints
+}
+
+func (r *RendezvousPointIterator) Next(ctx context.Context) <-chan *RendezvousPoint {
```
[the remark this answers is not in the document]
richard-ramos: The tests are not run with -race. I created this issue to track this: https://github.com/waku-org/go-waku/issues/618

waku/v2/node/wakuoptions.go
```diff
@@ -80,10 +80,8 @@ type WakuNodeParameters struct {
 	enableStore     bool
 	messageProvider store.MessageProvider
 
-	rendezvousNodes        []multiaddr.Multiaddr
-	enableRendezvousServer bool
-
-	rendezvousDB *rendezvous.DB
+	enableRendezvous bool
```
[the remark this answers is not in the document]
richard-ramos: Interesting thought. If we do that for rendezvous, we should probably do the same for the other discovery protocols: discv5 and peer exchange. If we decide to follow this route, it would make sense to ensure that nwaku's defaults for these protocols also match go-waku

waku/v2/rendezvous/rendezvous.go
richard-ramos: `RegisterRelayShards(ctx context.Context, rs protocol.RelayShards, rendezvousPoints []*RendezvousPoint)` receives this kind of structure you suggest. Perhaps later we can remove this `RegisterShard` function and leave the function that receives a `RelayShard`. I'm not sure if we should remove now as we haven't really used these functions and might end up deciding that passing the specific cluster and shard ends up being a better dev experience :thinking:

PR #587, opened 2023-06-28
fix: set wakurelay at index 0 in gossipsub protocol list

richard-ramos:


richard-ramos:
I noticed there was a difference in the protocol being used in gossipsub for transmitting the messages. Changing the order of the protocol to set wakurelay at index 0 makes it behave like nwaku

cc: @chaitanyaprem

PR #588, opened 2023-06-28
chore: bump go-libp2p and deps

richard-ramos:


chaitanyaprem (commented):
Changes look good, but there seems to be a failure on the nix-flake CI task which will have to fixed.

PR #590, opened 2023-06-30
fix: p2p circuit and discv5 conns

richard-ramos:
This PR fixes some issues found while doing a comparison between go-waku and nwaku.
1. Adds two artificial delays to discv5 iterators: When `Next()` is called, if it takes less than 50ms, it wait `50ms - elapsed time` before requesting another peer from DiscV5; and adds a 5s delay every 15 peers.
2. Disables concurrent dialing in peer connector. This should be re-enabled once a performance comparison is done again with PR https://github.com/waku-org/go-waku/pull/597
3. Fixes the multiaddresses sent to the discovery connector since they were missing the `p2p` component

PR made in collaboration with @alrevuelta

### DIFF
--- a/waku/v2/discv5/discover.go
+++ b/waku/v2/discv5/discover.go
@@ -330,10 +330,43 @@ func (d *DiscoveryV5) FindPeersWithShard(ctx context.Context, cluster, index uin
 	return enode.Filter(iterator, predicate), nil
 }
 
+const peerDelay = 100 * time.Millisecond
+const bucketSize = 16
+
 func (d *DiscoveryV5) Iterate(ctx context.Context, iterator enode.Iterator, onNode func(*enode.Node, peer.AddrInfo) error) {
 	defer iterator.Close()
 
-	for iterator.Next() { // while next exists, run for loop
+	peerCnt := 0
+	for {
+		// Delay if .Next() is too fast
+		start := time.Now()
+		hasNext := iterator.Next()
+		if !hasNext {
+			break
+		}
+		elapsed := time.Since(start)
+		if elapsed < peerDelay {
+			t := time.NewTimer(peerDelay - elapsed)
+			select {
+			case <-ctx.Done():
+				return
+			case <-t.C:
+				t.Stop()
+			}
+		}
+
+		// Delay every 15 peers being returned
+		peerCnt++
+		if peerCnt == bucketSize {
+			t := time.NewTimer(5 * time.Second)
+			select {
+			case <-ctx.Done():
+				return
+			case <-t.C:
+				t.Stop()
+			}
+		}
+
 		_, addresses, err := enr.Multiaddress(iterator.Node())
 		if err != nil {
 			metrics.RecordDiscV5Error(context.Background(), "peer_info_failure")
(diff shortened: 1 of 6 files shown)


waku/v2/discv5/discover.go
```diff
@@ -322,10 +322,43 @@ func (d *DiscoveryV5) FindPeersWithShard(ctx context.Context, cluster, index uin
...
+				return
+			case <-t.C:
+				t.Stop()
+			}
+		}
+
+		// Delay every 15 peers being returned
+		peerCnt++
+		if peerCnt == bucketSize {
+			t := time.NewTimer(5 * time.Second)
```
chaitanyaprem, line 353: Better to use constants rather than magic numbers to improve readability.

waku/v2/node/wakunode2.go
```diff
@@ -246,9 +247,9 @@ func New(opts ...WakuNodeOption) (*WakuNode, error) {
 	// Setup peer connection strategy
 	cacheSize := 600
 	rngSrc := rand.NewSource(rand.Int63())
-	minBackoff, maxBackoff := time.Second*30, time.Hour
+	minBackoff, maxBackoff := time.Minute, time.Hour
 	bkf := backoff.NewExponentialBackoff(minBackoff, maxBackoff, backoff.FullJitter, time.Second, 5.0, 0, rand.New(rngSrc))
-	w.peerConnector, err = v2.NewPeerConnectionStrategy(cacheSize, w.opts.discoveryMinPeers, network.DialPeerTimeout, bkf, w.log)
+	w.peerConnector, err = v2.NewPeerConnectionStrategy(cacheSize, w.opts.discoveryMinPeers, 20*time.Second, bkf, w.log)
```
chaitanyaprem, line 252: Use a constant or a config.

waku/v2/node/wakunode2.go
```diff
@@ -246,9 +247,9 @@ func New(opts ...WakuNodeOption) (*WakuNode, error) {
 	// Setup peer connection strategy
 	cacheSize := 600
 	rngSrc := rand.NewSource(rand.Int63())
-	minBackoff, maxBackoff := time.Second*30, time.Hour
+	minBackoff, maxBackoff := time.Minute, time.Hour
```
chaitanyaprem, line 250: Improvement: We should start moving these values to config, so that it helps in tuning these params while testing without code changes. We can have a default config read from a file, so that tuning can be done there. Will take it up as part of peer management enhancements.

waku/v2/node/wakunode2.go
```diff
@@ -882,6 +883,11 @@ func (w *WakuNode) Peers() ([]*Peer, error) {
 		}
 
 		addrs := w.host.Peerstore().Addrs(peerId)
+		hostInfo, _ := multiaddr.NewMultiaddr(fmt.Sprintf("/p2p/%s", peerId.Pretty()))
```
chaitanyaprem, line 886: Suggestion: Better to make this a multiaddr utility function that can be used across.

chaitanyaprem (approved):
LGTM

PR #592, opened 2023-07-03
Persistence: Addressing code-climate issues to increase maintainability score

chaitanyaprem:
Addressing code-climate issues in persistence module.


waku/persistence/store.go
```diff
@@ -392,19 +404,41 @@ func (d *DBStore) Query(query *pb.HistoryQuery) (*pb.Index, []StoredMessage, err
...
 
+	d.log.Info(fmt.Sprintf("conditionstr: %s", conditionStr))
+	d.log.Info(fmt.Sprintf("parameters: %s", parameters))
+
 	paramCnt++
+	d.log.Info(fmt.Sprintf("paramCnt: %d", paramCnt))
+
 	sqlQuery += fmt.Sprintf("LIMIT $%d", paramCnt)
 	sqlQuery = fmt.Sprintf(sqlQuery, conditionStr, orderDirection, orderDirection, orderDirection, orderDirection)
+	d.log.Info(fmt.Sprintf("sqlQuery: %s", sqlQuery))
```
richard-ramos, line 415: I'd change all these logs from lines 407 to 415, to `Debug` if they're really necessary, otherwise we can remove them. Also the logger uses zap, so instead of using `fmt.Sprintf`, you should be able to do something like:
```
d.log.Debug("some message",  zap.String("conditionStr", conditionStr), zap.String("parameters", parameters), etc etc) 
```

waku/persistence/store.go
```diff
@@ -354,34 +331,69 @@ func (d *DBStore) Query(query *pb.HistoryQuery) (*pb.Index, []StoredMessage, err
...
 
+func (d *DBStore) prepareQuerySQL(query *pb.HistoryQuery) (string, []interface{}, error) {
+	sqlQuery := `SELECT id, receiverTimestamp, senderTimestamp, contentTopic, pubsubTopic, payload, version 
+	FROM message 
+	%s
+	ORDER BY senderTimestamp %s, id %s, pubsubTopic %s, receiverTimestamp %s `
+
+	var conditions []string
+	var parameters []interface{}
+	//parameters := make([]interface{}, 0) //Allocating size of 1 as pageSize is gauranteed param to be present.
```
richard-ramos, line 372:
```suggestion
```

richard-ramos (changes requested):
Cool PR! It will surely improve our codeclimate score :rocket: 
I just left some minor observations.

I'm wondering why CI did not run tho? cc: @jakubgs

jakubgs:
Because of this:
```json
{
  "message":"`waku-org` forbids access via a personal access token (classic). Please use a GitHub App, OAuth App, or a personal access token with fine-grained permissions.",
  "documentation_url":"https://docs.github.com/rest/reference/repos#get-a-repository"
}
```
Someone must have changed the org policy.

jakubgs:
Oh, but it needs wider permissions to do updates of checks:
```
Could not update commit status. Message:
{"message":"Resource not accessible by personal access token","documentation_url":"https://docs.github.com/rest/commits/statuses#create-a-commit-status"}
```

jakubgs:
I thought this might be it:

image

But it doesn't seem to be, and I see no separate settings for checks:
https://docs.github.com/en/rest/overview/permissions-required-for-fine-grained-personal-access-tokens?apiVersion=2022-11-28#repository-permissions-for-checks

jakubgs:
I have opened a support ticket with GitHub support to clarify if fine-grained tokens are even capable of updating checks.

waku/persistence/store.go
chaitanyaprem: Oh man, missed removing these debug statements i added for debugging some issue i ran into. Will remove them.

waku/persistence/sql_queries.go
```diff
@@ -0,0 +1,93 @@
...
+
+// CreateTable creates the table that will persist the peers
+func CreateTable(db *sql.DB, tableName string) error {
+	sqlStmt := fmt.Sprintf("CREATE TABLE IF NOT EXISTS %s (key TEXT NOT NULL UNIQUE, data BYTEA);", tableName)
+	_, err := db.Exec(sqlStmt)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
chaitanyaprem: Good catch! 
I remember diffing the 2 files to see the differences, might have missed this. Will move this to DB specific one.

waku/persistence/store.go
chaitanyaprem: Was running into some SQL issue when i changed it to slice. 
But now i think that might have been probably due to other issues i fixed later. Will change it and test.

jakubgs:
I have re-enabled classic tokens since GitHub Support has not gotten back to me about use of fine-grained tokens.

The jobs are running again.

jakubgs:
According to GitHub support:

>Check is only [available](https://docs.github.com/en/rest/guides/using-the-rest-api-to-interact-with-checks?apiVersion=2022-11-28#overview) for GitHub apps. https://docs.github.com/en/rest/guides/using-the-rest-api-to-interact-with-checks?apiVersion=2022-11-28#overview
>
>>REST API to manage checks is available exclusively to your GitHub Apps.
>
>https://docs.github.com/en/rest/guides/using-the-rest-api-to-interact-with-checks?apiVersion=2022-11-28#about-check-suites
>
>>Write permission for the REST API to interact with checks is only available to GitHub Apps.
>
>Hope this clarifies things.

Not sure how to use a GitHub App with jenkins, but lets leave it as it is for now.

chaitanyaprem:
@jakubgs , for some reason the nix-flake task seems to fail in multiple commits with some vendor related error eventhough we don't have any vendor folder for the repo. Can you please take a look once, as merging is blocked because of this.

richard-ramos:
@chaitanyaprem the issue with `go mod vendor` seems to be 'fixed' now. I re-ran the CI build and now the response is: 
```
error: hash mismatch in fixed-output derivation '/nix/store/6c0765vk50pzpxn9rdsnzpjq5864bxga-go-waku-go-modules.drv':
         specified: sha256-6VBZ0ilGcXhTXCoUmGdrRQqgnRwVJOtAe4JIMUCVw8Y=
            got:    sha256-TU/jog0MZNC4g13gaGm88gsKTRvmlcKkMeXZbaVf3fc=
```
to solve this, replace the hash here https://github.com/waku-org/go-waku/blob/master/flake.nix#L31 by `sha256-TU/jog0MZNC4g13gaGm88gsKTRvmlcKkMeXZbaVf3fc=`

PR #595, opened 2023-07-05
feat: subscribe to shards with --topic and refactors

richard-ramos:
- Shards in same cluster are advertised on ENR
- Store().Resume() was moved to app layer
- NoDefaultWakuTopic() was removed since it's the app that must determine whether it subscribes to the default waku topic or not
- Removed `Publish` from WakuNode (not really used and easy to implement in app layer if needed)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,7 +28,7 @@
           ];
           doCheck = false;
           # FIXME: This needs to be manually changed when updating modules.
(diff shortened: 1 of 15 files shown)


chaitanyaprem (approved):
LGTM

Had few optimization suggestions, but can be taken up at later stage.

PR #596, opened 2023-07-06
Feat/peer manager

chaitanyaprem:
Introduce peer manager as per #621

1.  connection pruning for inRelay peers as per max default.
2.  add new flag --max-connections to set max number of peers to be allowed
3. Control number of outRelayPeers to be connected based on max connections as per https://github.com/waku-org/nwaku/pull/1813.

Tested pruning of incoming relay peer connections based on configuration. 
Tested outRelay Peers limit based on configuration.

Not sure if this will impact service peer connectivity, since that has not yet been tested.
If so, then will implement service slots next which should take care of that.


waku/v2/peermanager/peer_manager.go
```diff
@@ -0,0 +1,121 @@
...
+	"github.com/libp2p/go-libp2p/core/protocol"
+	wps "github.com/waku-org/go-waku/waku/v2/peerstore"
+	"golang.org/x/exp/slices"
+
+	"go.uber.org/zap"
+)
+
+// TODO: Move all the protocol IDs to a common location.
+// WakuRelayIDv200 is protocol ID for Waku v2 relay protocol
+const WakuRelayIDv200 = protocol.ID("/vac/waku/relay/2.0.0")
```
richard-ramos, line 18: this constant can be obtained from `relay` package

waku/v2/peermanager/peer_manager.go
```diff
@@ -0,0 +1,121 @@
...
+	go pm.connectivityLoop()
+}
+
+// This is a connectivity loop, which currently checks and prunes inbound connections.
+func (pm *PeerManager) connectivityLoop() {
+
+	for {
+		pm.pruneInRelayConns()
+		time.Sleep(peerConnectivityLoopSecs * time.Second)
+	}
```
richard-ramos, line 68: I'd suggest replacing this code to use a context, so the connectivity loop goroutine can be stopped when the node is closed. i.e.:
```suggestion
func (pm *PeerManager) Start(ctx context.Context) {
	go pm.connectivityLoop(ctx)
}

// This is a connectivity loop, which currently checks and prunes inbound connections.
func (pm *PeerManager) connectivityLoop(ctx context.Context) {
        t := time.NewTicker(peerConnectivityLoopSecs * time.Second)
	for {
	       select {
	               case <-ctx.Done():
	                     return
	               case <-t.C:
	                          pm.pruneInRelayConns()
	       }
	}
```

The wakunode should have a cancellable context somewhere that can be passed to this start function.

waku/v2/peermanager/peer_manager.go
```diff
@@ -0,0 +1,121 @@
...
+		supportedProtocols, err := pm.host.Peerstore().GetProtocols(p)
+		if err != nil {
+			pm.logger.Warn("Failed to get supported protocols for peer", zap.String("peerID", p.String()))
+		}
+		if slices.Contains(supportedProtocols, proto) {
+			filteredPeers = append(filteredPeers, p)
+		}
+	}
+	return filteredPeers
+}
```
richard-ramos, line 85: The peerstore has a function that can be used for checking supported protocols. It's cool because then we can change this `filterPeersByProto` function to check for peers that support more than one protocol, in case we need this in the future.
Notice i also added a `continue` in the `err` verification.
```suggestion
func (pm *PeerManager) filterPeersByProto(peers peer.IDSlice, proto ...protocol.ID) peer.IDSlice {
	var filteredPeers peer.IDSlice
	//TODO: This can be optimized once we have waku's own peerStore

	for _, p := range peers {
		supportedProtocols, err := pm.host.Peerstore().SupportsProtocols(p, proto...)
		if err != nil {
			pm.logger.Warn("Failed to get supported protocols for peer", zap.String("peerID", p.String()))
		        continue
		}
		if len(supportedProtocols) != 0 {
			filteredPeers = append(filteredPeers, p)
		}
	}
	return filteredPeers
}
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -0,0 +1,121 @@
...
+			p := inRelayPeers[pruningStartIndex]
+			err := pm.host.Network().ClosePeer(p)
+			if err != nil {
+				pm.logger.Warn("Failed to disconnect connection towards peer", zap.String("peerID", p.String()))
+			}
+			pm.host.Peerstore().RemovePeer(p) //TODO: Should we remove the peer immediately?
+			err = pm.host.Peerstore().(wps.WakuPeerstore).SetDirection(p, network.DirUnknown)
+			if err != nil {
+				pm.logger.Warn("Failed to remove metadata for peer", zap.String("peerID", p.String()))
+			}
```
richard-ramos, line 117: Looking at the implementation of `RemovePeer` in go-libp2p, it remove the metadata for the peers. Then i think it is not necessary necessary to set the direction to unknown in the next line as the record would have been removed from the DB anyways

waku/v2/peermanager/peer_manager.go
```diff
@@ -0,0 +1,121 @@
...
+
+	//Need to filter peers to check if they support relay
+	inRelayPeers = pm.filterPeersByProto(inPeers, WakuRelayIDv200)
+	outRelayPeers := pm.filterPeersByProto(outPeers, WakuRelayIDv200)
+	pm.logger.Info("Number of Relay peers connected", zap.Int("inRelayPeers", inRelayPeers.Len()), zap.Int("outRelayPeers", outRelayPeers.Len()))
+
+	if inRelayPeers.Len() > int(pm.InRelayPeersTarget) {
+		//Start disconnecting peers, based on what?
+		//For now, just disconnect most recently connected peers
+		//TODO: Need to have more intelligent way of doing this, maybe peer scores.
```
richard-ramos, line 105: I like the idea of using peer score for choosing which peer to remove!

richard-ramos (commented):
Nice PR!

waku/v2/peermanager/peer_manager.go
chaitanyaprem: If i do that, it was causing some circular dependency. 
I think we will have to move all protocol constants to a common file so that they can be imported wherever we want.

waku/v2/peermanager/peer_manager.go
chaitanyaprem: Thanks for catching that. Will change it.

richard-ramos (approved):

PR #599, opened 2023-07-06
fix: warning message format when using static shards and named shards at same time

richard-ramos:


waku/v2/node/localnode.go
```diff
@@ -305,17 +305,29 @@ func (w *WakuNode) watchTopicShards(ctx context.Context) error {
...
 			case <-evtRelaySubscribed.Out():
-				rs, err := protocol.TopicsToRelayShards(w.Relay().Topics()...)
+				topics := w.Relay().Topics()
+				rs, err := protocol.TopicsToRelayShards(topics...)
 				if err != nil {
 					w.log.Warn("could not set ENR shard info", zap.Error(err))
 					continue
 				}
 
 				if len(rs) > 1 {
```
chaitanyaprem, line 315:
```suggestion

				if len(rs) > 0 {
				    if len(rs) > 1 {
					     w.log.Warn("use sharded topics within the same cluster")
					    w.log.Warn("could not set ENR shard info", zap.String("error", "use sharded topics within the same cluster"))
					    continue
				    }

```

chaitanyaprem (commented):
Minor suggestion.
Otherwise LGTM

PR #600, opened 2023-07-06
feat(c-bindings): websockets support

richard-ramos:


chaitanyaprem (commented):
LGTM.

Would be good to address the minor code-Climate issues - adding comments for exported types and functions. Let me see if we can add this check as part of pre-commit hook itself, so that we don't have to wait for CI to report such minor issues.

richard-ramos:
> Let me see if we can add this check as part of pre-commit hook itself, so that we don't have to wait for CI to report such minor issues.

Sounds like a good idea!

PR #603, opened 2023-07-07
fix: set discovery concurrent dials to 5

richard-ramos:


waku/v2/discovery_connector.go
```diff
@@ -245,8 +245,18 @@ func (c *PeerConnectionStrategy) workPublisher(ctx context.Context) {
...
+
 func (c *PeerConnectionStrategy) dialPeers(ctx context.Context) {
 	defer c.wg.Done()
+
+	maxGoRoutines := c.minPeers
+	if maxGoRoutines > maxActiveDials {
+		maxGoRoutines = maxActiveDials
+	}
+
+	sem := make(chan struct{}, maxGoRoutines)
```
chaitanyaprem, line 258: Why not use a channel of int?

richard-ramos: This is a buffered signal channel used as a semaphore to limit the number of go-routines. I used struct{} because an empty struct has no memory cost.

I could have used "golang.org/x/sync/semaphore" instead but using this is IMO simpler and achieves the same result

chaitanyaprem: This is the better option then.

chaitanyaprem (approved):
LGTM

PR #604, opened 2023-07-12
chore: add back lock to discv5

richard-ramos:
In status-go, there are different go-routines concurrently handling the `Start()` and `Stop()` of discoveryV5, and I noticed that sometimes a race condition happened with both of these function being executed at the same time. The usage of the atomic's `CompareAndSwap` was not protecting the channel `peerCh` nor the `cancel` function, so we ended up with an inconsistent state.

### DIFF
--- a/waku/v2/discv5/discover.go
+++ b/waku/v2/discv5/discover.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/libp2p/go-libp2p/core/host"
@@ -33,6 +32,8 @@ type PeerConnector interface {
 }
 
 type DiscoveryV5 struct {
+	sync.RWMutex
+
 	params    *discV5Parameters
 	host      host.Host
 	config    discover.Config
@@ -46,9 +47,8 @@ type DiscoveryV5 struct {
 
 	log *zap.Logger
 
-	started atomic.Bool
-	cancel  context.CancelFunc
-	wg      *sync.WaitGroup
+	cancel context.CancelFunc
+	wg     *sync.WaitGroup
 }
 
 type discV5Parameters struct {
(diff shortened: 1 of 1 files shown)


chaitanyaprem:
Would be good to have some context in the PR. 
Also, finding it difficult to figure out which data in the discv5 struct the lock is supposed to be for apart from the cancel field. Would be helpful if that info is indicated as comments .

waku/v2/discv5/discover.go
```diff
@@ -192,9 +192,10 @@ func (d *DiscoveryV5) SetHost(h host.Host) {
 
 // only works if the discovery v5 hasn't been started yet.
 func (d *DiscoveryV5) Start(ctx context.Context) error {
-	// compare and swap sets the discovery v5 to `started` state
-	// and prevents multiple calls to the start method by being atomic.
-	if !d.started.CompareAndSwap(false, true) {
+	d.Lock()
+	defer d.Unlock()
+
+	if d.cancel != nil {
```
chaitanyaprem, line 198: It would be better to release the Lock once the cancel has been assigned, otherwise we are holding the lock for too long as we are invoking listen which is a networking call. 
This way any other routine waiting on the lock would not get blocked for too long only until the context is created.

waku/v2/discv5/discover.go
```diff
@@ -291,6 +300,9 @@ func evaluateNode(node *enode.Node) bool {
 // used for caching enr address in peerExchange
 // used for connecting to peers in discovery_connector
 func (d *DiscoveryV5) Iterator() (enode.Iterator, error) {
+	d.Lock()
+	defer d.Unlock()
+
```
chaitanyaprem, line 305: Did not get the purpose of this lock. 
Let's assume that this is to lock when someone is iterating, stop is not called. 

But since we are returning the iterator itself, this lock is not enough to protect caller from using the iterator after stop is called which may lead to invalid access.

richard-ramos:
Closing as https://github.com/waku-org/go-waku/pull/687 solves the problem!

PR #610, opened 2023-07-16
ci: docker image

dozyio:
- Builds and pushes an `amd64` docker image when there is a push to master.
- Bumps build image to Go 1.20 / Alpine 3.18
- Bumps final image to Alpine 3.18

I also wanted to also build an `arm64` image, but builds were timing out. I'll try to revisit this some point in the future.


.github/workflows/docker.yml
```diff
@@ -0,0 +1,51 @@
...
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
```
chaitanyaprem, line 49: I think we would want to define a latest and stable tags for images. 
Images that are automatically generated by PR merge would be called latest.
@richard-ramos WDYT?

Stable tag would be used only when a release is made. Which i am guessing can be identified in the github action by using some sort of version tag.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,51 @@
+name: Docker image
+
+on:
+  push:
+    branches: ["master"]
+
+env:
+  REGISTRY: ghcr.io
```
chaitanyaprem, line 8: Last i checked, we had been uploading go-waku images to docker-hub registry under statusteam. 
Here the images shall be uploaded to github's container registery.

@richard-ramos , @fryorcraken  Need your input here as to if we will be continuing to store docker images in dockerhub under status or use github. This would apply for all Waku images and not just go-waku as it makes sense to store all of them at one location.

chaitanyaprem (commented):
Thanks a lot for the PR @dozyio.
PR looks good to me except few details which we want to iron out.

.github/workflows/docker.yml
fryorcraken: I would engage the discussion on our discord (infra channel) for this.

chaitanyaprem:
Looks like we are going to take the Jenkins route and upload docker images to dockerhub where our team is already uploading other images. 
@dozyio Closing this PR for now. In future if we move to github actions then we can reuse this one.

PR #611, opened 2023-07-19
chore: add missing comments to some functions

richard-ramos:


waku/v2/protocol/rln/common.go
```diff
@@ -6,23 +6,23 @@ import (
...
 
-type MessageValidationResult int
+type messageValidationResult int
 
 const (
-	MessageValidationResult_Unknown MessageValidationResult = iota
-	MessageValidationResult_Valid
-	MessageValidationResult_Invalid
-	MessageValidationResult_Spam
+	validationError messageValidationResult = iota
```
chaitanyaprem, line 12: Should we use all caps for constants to keep them consistent Or is this what is recommended by codeclimate?

chaitanyaprem (approved):

waku/v2/protocol/rln/common.go
richard-ramos: It's what code climate recommended

PR #615, opened 2023-07-27
feat(rln-relay): pass in index to keystore credentials

richard-ramos:
Allows users/operators to pass in the index of credentials to use. This is done with the `--rln-relay-membership-group-index` flag


waku/v2/protocol/rln/onchain_test.go
```diff
@@ -309,7 +309,7 @@ func (s *WakuRLNRelayDynamicSuite) TestCorrectRegistrationOfPeers() {
 	defer sub2.Unsubscribe()
 
 	// mount the rln relay protocol in the on-chain/dynamic mode
-	gm2, err := dynamic.NewDynamicGroupManager(s.clientAddr, s.u2PrivKey, s.rlnAddr, "./test_onchain.json", "", false, nil, utils.Logger())
+	gm2, err := dynamic.NewDynamicGroupManager(s.clientAddr, s.u2PrivKey, s.rlnAddr, 0, "./test_onchain.json", "", 0, false, nil, utils.Logger())
```
chaitanyaprem, line 312: Would be good to add a test to verify at a different index than 0.

richard-ramos: I changed the code in https://github.com/waku-org/go-waku/pull/615/commits/81aea259659a7563aece6adc515fe95102b1cd12 so Start() would return an error instead and not panic with an index out of range if the indexes passed are incorrect. The datatype for the index was changed to uint to avoid negative values as well. I'm thinking we could create a test for this could be part of a coverage improvement PR for RLN  :thinking:

chaitanyaprem: Sounds good! Do add it to the relevant issue ticket so that we don't miss it.

chaitanyaprem (approved):
LGTM

PR #616, opened 2023-07-27
feat: verify proofs based on bandwidth usage

richard-ramos:
Runs rln verification only if bandwidth exceeds 1mbps

### DIFF
--- a/cmd/waku/flags_rln.go
+++ b/cmd/waku/flags_rln.go
@@ -73,7 +73,7 @@ func rlnFlags() []cli.Flag {
 				Value: &options.RLNRelay.MembershipContractAddress,
 			},
 		},
-		&cli.IntFlag{
+		&cli.Int64Flag{
 			Name:        "rln-relay-bandwidth-threshold",
 			Value:       0,
 			Usage:       "Message rate in bytes/sec after which verification of proofs should happen. Use 0 to disable bandwidth rate limits",
--- a/cmd/waku/flags_rln.go
+++ b/cmd/waku/flags_rln.go
@@ -73,5 +73,11 @@ func rlnFlags() []cli.Flag {
 				Value: &options.RLNRelay.MembershipContractAddress,
 			},
 		},
+		&cli.IntFlag{
+			Name:        "rln-relay-bandwidth-threshold",
+			Value:       0,
+			Usage:       "Message rate in bytes/sec after which verification of proofs should happen. Use 0 to disable bandwidth rate limits",
+			Destination: &options.RLNRelay.BandwidthThreshold,
+		},
 	}
 }
--- a/cmd/waku/options.go
+++ b/cmd/waku/options.go
@@ -44,6 +44,7 @@ type RLNRelayOptions struct {
 	ETHPrivateKey             *ecdsa.PrivateKey
 	ETHClientAddress          string
 	MembershipContractAddress common.Address
+	BandwidthThreshold        int
 }
 
 // FilterOptions are settings used to enable filter protocol. This is a protocol
--- a/cmd/waku/options.go
+++ b/cmd/waku/options.go
@@ -44,7 +44,7 @@ type RLNRelayOptions struct {
 	ETHPrivateKey             *ecdsa.PrivateKey
 	ETHClientAddress          string
 	MembershipContractAddress common.Address
-	BandwidthThreshold        int
+	BandwidthThreshold        int64
 }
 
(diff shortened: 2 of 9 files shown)


cmd/waku/flags_rln.go
```diff
@@ -73,5 +73,11 @@ func rlnFlags() []cli.Flag {
 				Value: &options.RLNRelay.MembershipContractAddress,
 			},
 		},
+		&cli.IntFlag{
+			Name:        "rln-relay-bandwidth-threshold",
+			Value:       0,
+			Usage:       "Message rate in bytes/sec after which verification of proofs should happen. Use 0 to disable bandwidth rate limits",
```
chaitanyaprem, line 79: I think you meant Message rage in bits/sec as bandwidth is always represented in bits/sec.

cmd/waku/options.go
```diff
@@ -44,6 +44,7 @@ type RLNRelayOptions struct {
 	ETHPrivateKey             *ecdsa.PrivateKey
 	ETHClientAddress          string
 	MembershipContractAddress common.Address
+	BandwidthThreshold        int
```
chaitanyaprem, line 47: Maybe better to use int64.

waku/v2/node/wakunode2_rln.go
```diff
@@ -73,7 +74,12 @@ func (w *WakuNode) mountRlnRelay(ctx context.Context) error {
 		}
 	}
 
-	rlnRelay, err := rln.New(w.Relay(), groupManager, w.opts.rlnRelayPubsubTopic, w.opts.rlnRelayContentTopic, w.opts.rlnSpamHandler, w.timesource, w.log)
+	var limiter *rate.Limiter
+	if w.opts.rlnRelayBandwidthThreshold != 0 {
+		limiter = rate.NewLimiter(rate.Limit(w.opts.rlnRelayBandwidthThreshold), w.opts.rlnRelayBandwidthThreshold)
```
chaitanyaprem, line 79: Wondering if we should we permit a burst of full bandwidth limit.
Generally bursts are set to a percentage of full-rate especially when rate limits are high like this.

cmd/waku/flags_rln.go
richard-ramos: I copy-pasted this text from nwaku :-P 
But then, checking how the bandwidth is verified, I use bytes there (len of a byte slice)

waku/v2/node/wakunode2_rln.go
richard-ramos: hm. not sure, tho. I set the burst and rate limit to be the same to match nwaku's behavior (since nwaku's rate limiter does not have the concept of bursts)

chaitanyaprem: Interesting, generally rate-limiters define smaller bursts so that all of the limits are not consumed (especially when limits are high). This is to ensure system is not hit with too big of a burst that it can't handle or causes it to slow down, it is a way to throttle. 

Now that i am rethinking, probably in this case since we are using this rate-limiter only to determine whether to enable proof verification, keeping burst same as rate should be fine.

cmd/waku/flags_rln.go
chaitanyaprem: This was more from a user-facing experience point of view. People are using to bits/sec when bandwidth is specified. We should keep the user-facing flag to use bits/sec and internally multiply by 8 to store the rate as it is easy to account. 

Let us ask the nwaku team also to follow the same.

richard-ramos: cc: @rymnc

rymnc: Hmm yes, we have to revert that commit. Based on our offsite discussion, we will have to move the bandwidth constraining into its own validator, before the rln-relay validator - I'll create an issue and link it here for the design, does that work?

And yes, we'll use bits/sec!

richard-ramos: Works for me. I'll keep the PR open and update it accordingly according to the design. Thank you!

alrevuelta: @rymnc i would say we also need to remove that flag and have it as a hardcoded parameter? otherwise if different nodes have different configuration a message can be valid for ones not not for others.

chaitanyaprem:
I think this PR is no longer requried and can be closed. 
If we see the need for this feature at a later point in time, we can take code from this and create a new PR then

PR #619, opened 2023-08-02
refactor: c-bindings

richard-ramos:
Updates the c-bindings to match the style of using success and error callbacks defined by @Ivansete-status in nwaku's c-bindings.

- [X] Use callbacks in c library
- [X] reorg mobile and c library to reuse same code but with different response format
- [ ] update docs
- [ ] update example


Ivansete-status:
It looks very nice! I look forward to becoming a reviewer :)

richard-ramos:
Finally done with this PR :) . It ended up being quite large due to moving a lot of files and extracting functions so they could be used both for the mobile bindings and the c bindings


@Ivansete-status:

Here's the generated header if it makes it simpler to review, instead of focusing in mostly repetitive code changes in go: https://gist.github.com/richard-ramos/bb1a6b6a2ae59cb3949a4627ab0b139b . Notice that some functions in here receive a callback, while they don't in nwaku, (`waku_start`  and `waku_stop` can fail in go-waku). Should we have the same signature in nwaku even if it is not possible for it to fail on start/stop? (the onErrCallback will never be called there)



Something else that I wanted to discuss with you is how are we going to handle functions that do not receive a string result. for example these:
```c
// Get number of connected peers
extern int waku_peer_cnt(WakuCallBack onOkCb, WakuCallBack onErrCb);

// Determine if there are enough peers to publish a message on a topic. Use NULL
// to verify the number of peers in the default pubsub topic
extern int waku_relay_enough_peers(char* topic, WakuCallBack onOkCb, WakuCallBack onErrCb);
```

The peer count is supposed to be a number,  and the second function is supposed to be a boolean. In this PR i'm returning the string representation of a number and a boolean, but IMO it's ugly. Do you think we should have instead different types of callbacks instead? something like this:

```
typedef void (*WakuStringCallBack) (const char* msg, size_t len_0);

typedef void (*WakuIntCallBack) (int value);
```

Then `WakuIntCallBack` could be used both for numbers and booleans (returning 1 or 0).
What do you think?

richard-ramos:
@chaitanyaprem I just noticed the infinite number of codeclimate issues that were generated. I'll try to fix them :)

chaitanyaprem:
Noticed that ReadMe intro section needs to be modified to change nwaku to go-waku. 
https://github.com/waku-org/go-waku/blob/721c71b48c3ac1e357da5d85c51a808432de21b2/library/c/README.md?plain=1#L9

library/signals.go
```diff
@@ -56,8 +56,9 @@ func send(signalType string, event interface{}) {
 		mobileSignalHandler(data)
 	} else {
 		// ...and fallback to C implementation otherwise.
-		str := C.CString(string(data))
-		C.StatusServiceSignalEvent(str)
+		dataStr := string(data)
+		str := C.CString(dataStr)
+		C.StatusServiceSignalEvent(str, C.size_t(len(data)))
```
chaitanyaprem, line 61: Maybe we should rename this to ServiceSignalEvent?

library/mobile/api_store.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+// StoreQuery is used to retrieve historic messages using waku store protocol.
+func StoreQuery(queryJSON string, peerID string, ms int) string {
+	response, err := library.StoreQuery(queryJSON, peerID, ms)
+	return prepareJSONResponse(response, err)
+}
+
+// StoreLocalQuery is used to retrieve historic messages stored in the localDB using waku store protocol.
+func StoreLocalQuery(queryJSON string) string {
```
chaitanyaprem, line 14: Why are 2 different methods required? 
I am thinking, wouldn't it be better to expose to the user a simple method called StoreQuery which checks localDB and if not present fetches from the peer?

chaitanyaprem (approved):
LGTM

Left few minor comments, which are not critical.

library/mobile/api_store.go
richard-ramos: The problem is that you do not know if the data you have locally is complete. For an specific time period you might have `n` messages on your db, and the peer might have `n+m` message, so attempting to retrieve messages locally might return results and still end up missing messages. :thinking:

chaitanyaprem: Ok, it's the users responsibility to identify what is missing and query peer accordingly.

Issue #618, opened 2023-07-31
Run unit tests with race


richard-ramos:
- Run all unit tests with -race flag so that data races can be detected and fixed
- Fix issues observed based on the same.

PR #627, opened 2023-08-04
Fix/data races

chaitanyaprem:
Enabled go data race detector for tests per #618  and fixing the data races one by one.


jakubgs (changes requested):
Honestly, I'd split this into two PRs. One to add the Go changes, and another to add the CI changes.

richard-ramos (approved):
:100:

richard-ramos:
> @jakubgs: Why exactly is this a separate Jenkinsfile and not just part of tests?

this is because for now we want the tests to be required for merging the PRs, however if we enable the race flag, tests might fail for reasons unrelated to the PRs, so, to avoid delaying PRs and to not include changes unrelated to the PRs themselves, we are running the tests twice, one required, that does not use the race flag, and another one that is just informative, with the race flag enabled,  that way if we find a bug with a race condition, we can fix it into a separate issue.

chaitanyaprem:
> Honestly, I'd split this into two PRs. One to add the Go changes, and another to add the CI changes.

Ok, will follow this approach from next time. 
For now, please review only the jenkins file. Others are reviewed by @richard-ramos .

jakubgs (approved):
Honestly, I don't think we need a separate jenkinsfile. We can just use the same Jenkinsfile and just add a conditional in the test stage. We could have a parameter like `ENABLE_RACE_TESTS` or something, and then have it be on by default when job name is `race` or something like that. There's not need to maintain two nearly identical Jenkinsfiles.

jakubgs:
I mean, we could just have something like this:
```groovy
    stage('On-chain tests') {
      environment {
        GANACHE_NETWORK_RPC_URL = "ws://localhost:${env.GANACHE_RPC_PORT}"
      }
      steps { script { dir(env.REPO) {
        if (param.TEST_WITH_RACE) {
          nix.develop('make test-onchain-with-race', pure: false)
        } else {
          nix.develop('make test-onchain', pure: false)
        }
      } } }
```

chaitanyaprem:
@jakubgs , merging this.
Please create a new job with name "race" in jenkins to run the tests target. 

@richard-ramos , After that i guess we just need to change settings in the repo to make this job as optional in PR checks

jakubgs:
There you go: https://ci.status.im/job/go-waku/job/prs/job/race/

Though I thought a parameter would be better, but ah well.

chaitanyaprem:
Cool, will verify this in next PR.

PR #628, opened 2023-08-04
chore: remove v1 code and legacy filter from chat2

richard-ramos:
Also removes the private API from RPC. C-Bindings are not affected, but future revisions might remove the asymmetric/symmetric functions for lightpush and relay and replace them instead by separate encoding functions

PR roughly equivalent to https://github.com/waku-org/nwaku/pull/1874


jm-clius (approved):
Seems like we're doing it! :D

chaitanyaprem (approved):
LGTM

Issue #605, opened 2023-07-15
[Milestone] RLN: Post-Testnet3 Improvements


richard-ramos:
**Planned start date**:
**Due date**:

# Summary

go-waku's RLN feature has fallen behind compared to nwaku. Code equivalent to the following PRs from nwaku must be implemented for go-waku's to be up to date with RLN functionality and also allowing it to be used with the simulation tools

- [x] https://github.com/waku-org/nwaku/issues/1733
- [x] https://github.com/waku-org/nwaku/issues/1777
- [x] https://github.com/waku-org/nwaku/issues/1805
- [x] https://github.com/waku-org/nwaku/issues/1838 
- [x] https://github.com/waku-org/nwaku/issues/1844

# Acceptance Criteria

- [x] Feature parity
- [x] Interop is achieved between nwaku and go-waku with messages going back and forth between chat2 clients
  - [x] go-waku chat2 -> go-waku relay node -> nwaku chat2
  - [x] go-waku chat2 -> nwaku relay node -> go-waku chat2
  - [x] [ensure credentials generated in nwaku work in go-waku and viceversa](https://github.com/waku-org/go-waku/pull/640)

## Tasks

# RAID (Risks, Assumptions, Issues and Dependencies)

richard-ramos:
#1777 implemented in https://github.com/waku-org/go-waku/pull/615

richard-ramos:
#1838 implemented in https://github.com/waku-org/go-waku/pull/612

richard-ramos:
#1844 implemented in https://github.com/waku-org/go-waku/pull/616

fryorcraken:
@rymnc are all the functionality listed the one you delivered for testnet 3? I wonder if we can scope this issue to reach testnet 3 feature equivalence.

rymnc:
Yes, and the following are the issues sorted from the retrospective post-testnet 3

https://github.com/vacp2p/research/issues/179#issuecomment-1579202451

richard-ramos:
nwaku#1733 implemented in #617

richard-ramos:
**Weekly Update**

_achieved_: use zerokit batch insert/delete for members, exposed function to retrieve data from merkle tree, modified zerokit and go-zerokit-rln to pass merkle tree persistance configuration settings
_next_: resume onchain sync from persisted tree db
_blocker_: none

richard-ramos:
**Weekly Update**

[...]

PR #630, opened 2023-08-07
feat: merkle tree persistence

richard-ramos:
# Description
Persist Merkle Tree in sled db
This is part of Milestone: https://github.com/waku-org/go-waku/issues/605

# Changes

- Resume onchain sync from persisted tree db
- Close eth client and DB connection appropriately
- Pass in the path to the tree db


richard-ramos:
:thinking: For some reason, when a tree config is passed in https://github.com/waku-org/go-waku/pull/630/files#diff-f9f1366bf48973e91af5a769c566b2662c33177e31388325c7f9307b4e7cca55R72 RLN refuses to initialize. Tomorrow I'll figure out what's going on here so I can properly test interop between nwaku and go-waku

richard-ramos:
The issue was that I was using the same database for all nodes in the test units, so RLN was complaining about not being able to acquire locks on the DB

waku/v2/protocol/rln/group_manager/dynamic/metadata.go
```diff
@@ -0,0 +1,39 @@
+package dynamic
+
+import (
+	"encoding/binary"
+	"errors"
+)
+
+type RLNMetadata struct {
```
chaitanyaprem, line 8: Need to add comments for all the exported types.

waku/v2/protocol/rln/group_manager/dynamic/dynamic.go
```diff
@@ -278,17 +297,28 @@ func (gm *DynamicGroupManager) persistCredentials() error {
...
+			idCommitments = append(idCommitments, rln.Bytes32(evt.Pubkey.Bytes()))
+		}
+
+		if len(idCommitments) == 0 {
+			continue
+		}
+
+		// TODO: should we track indexes to identify missing?
+		startIndex := rln.MembershipIndex(uint(oldestIndexInBlock.Int64()))
+		err := gm.rln.InsertMembers(startIndex, idCommitments)
```
chaitanyaprem, line 315: Does the RLN API take care of a scenario where few members are already inserted and not throw error? If not, this has to be handled at this level.

e.g: While inserting members, let's say member insertion succeeded whereas before metadata got updated, node went down. Next time when node comes up, it will try to insert some set of members which are already inserted and probably a new set if they are part of a later block. In this case, atomicOp should not fail. If it fails, then we may have to handle it at this layer(identify if members are already present and not insert them again).

richard-ramos: > Does the RLN API take care of a scenario where few members are already inserted and not throw error? I

The RLN API handles it by acting as an upsert, overwriting whatever is stored at some index in the merkle tree, so atomicOp should not fail as it will just replace the same members at the same position with the same values

waku/v2/protocol/rln/group_manager/dynamic/web3.go
```diff
@@ -112,7 +112,17 @@ func (gm *DynamicGroupManager) HandleGroupUpdates(ctx context.Context, handler R
 }
 
 func (gm *DynamicGroupManager) loadOldEvents(ctx context.Context, rlnContract *contracts.RLN, handler RegistrationEventHandler) error {
-	events, err := gm.getEvents(ctx, 0, nil)
+
```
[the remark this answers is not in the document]
richard-ramos: It should be possible to retrieve events with a full node because these type of nodes contain all transaction receipts/logs. A full node can still be costly so I agree that we'll need to come up with a solution for obtaining the merkle tree changes, instead of relying only on the event logs (although for the time being, it should be fine with using infura/pocket)

richard-ramos:
@chaitanyaprem I updated this PR with fixes for code review observations, most code climate issues, and also fixed some interop issues I found while attempting to get messages going back and forth between nwaku's chat2 and go-waku's

cmd/waku/flags_rln.go
```diff
@@ -30,7 +31,7 @@ func rlnFlags() []cli.Flag {
 		},
 		&cli.StringFlag{
 			Name:        "rln-relay-content-topic",
-			Value:       "/toy-chat/2/luzhou/proto",
+			Value:       protocol.NewContentTopic("toy-chat", 3, "mingde", "proto").String(),
```
chaitanyaprem, line 34: Does it make sense to have a default value for content-topic? Rather we should throw an error if content-topic is not mentioned. 
We can always inlcude in docs which content-topic can be used if someone wants to just test.

examples/chat2/chat.go
```diff
@@ -77,16 +82,19 @@ func NewChat(ctx context.Context, node *node.WakuNode, connNotifier <-chan node.
...
+			if err != nil {
+				chat.ui.ErrorMessage(err)
+			} else {
+				chat.C = make(chan *protocol.Envelope)
+				go func() {
+					for e := range sub.Ch {
+						chat.C <- e
+					}
+				}()
+			}
```
chaitanyaprem, line 97: Shouldn't the wait-group be based on number of times we loop?
It is hardcoded to 7 in line 101.

chaitanyaprem (approved):
LGTM other than few minor comments.

cmd/waku/flags_rln.go
alrevuelta: Note that we plan to remove rln pubsub topics and content topics, so that rln is applied in all the subscribed ones if enabled.

See 1. from https://github.com/waku-org/nwaku/issues/1906
and https://discord.com/channels/1110799176264056863/1111540684575477821/1141378004841410640

richard-ramos: For this PR let's keep it with the hardcoded default content topic to match nwaku's flag: https://github.com/waku-org/nwaku/blob/master/apps/wakunode2/external_config.nim#L165

examples/chat2/chat.go
richard-ramos: That 7 refers to the goroutines created from line 102 to 113:
```
go chat.parseInput()
go chat.receiveMessages()
go chat.welcomeMessage()
go chat.connectionWatcher(&connectionWg, connNotifier)
go chat.staticNodes(&connectionWg)
go chat.discoverNodes(&connectionWg)
go chat.retrieveHistory(&connectionWg)
```

chaitanyaprem: Ah..got it.

PR #632, opened 2023-08-08
chore: keepAlive peer disconnecting peers only on ping failure

richard-ramos:
# Description
@cammellos noticed that sometimes in mobile, go-waku is too eager to disconnect all peers, with this happening even when peers are connected and working fine. This PR changes the behavior to not immediatly disconnect peers when waking up from sleep, but do it only if pinging the peer fails

### DIFF
--- a/waku/v2/node/keepalive.go
+++ b/waku/v2/node/keepalive.go
@@ -5,7 +5,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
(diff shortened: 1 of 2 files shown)


chaitanyaprem (commented):
PR looks fine. 

But if we want this to work on wake-up of a device, we should have some overall time-limit that we want this detection to occur, not sure how much would that be. 

Take a scenario when a node is connected to 30peers, then it may take 30*7=more than 3 minutes to detect and disconnect peers which isn't deterministic.

richard-ramos:
> PR looks fine.
> 
> But if we want this to work on wake-up of a device, we should have some overall time-limit that we want this detection to occur, not sure how much would that be.
> 
> Take a scenario when a node is connected to 30peers, then it may take 30*7=more than 3 minutes to detect and disconnect peers which isn't deterministic.

Should take less since pinging is done in separate go-routines for each peer https://github.com/waku-org/go-waku/blob/master/waku/v2/node/keepalive.go#L58. (which also makes me think that perhaps we should limit the maximum number of goroutines created in this step)

chaitanyaprem:
> > PR looks fine.
> > But if we want this to work on wake-up of a device, we should have some overall time-limit that we want this detection to occur, not sure how much would that be.
> > Take a scenario when a node is connected to 30peers, then it may take 30*7=more than 3 minutes to detect and disconnect peers which isn't deterministic.
> 
> Should take less since pinging is done in separate go-routines for each peer https://github.com/waku-org/go-waku/blob/master/waku/v2/node/keepalive.go#L58. (which also makes me think that perhaps we should limit the maximum number of goroutines created in this step)

Ah...missed that. Do we need to limit go-routines? 
Default max peers is 50, which should be fine i guess since all are only going to generate ping.

PR #633, opened 2023-08-08
feat: sqlite vacuum and optional migrations

richard-ramos:
# Description
This adds new flags already available in nwaku to allow users to execute the vacuum procedure on sqlite databases and choose whether they want to execute migrations or not

cc: @apentori 

# Changes

- [x] Adds new flags `--store-message-db-vacuum` and `--store-message-db-migrate`
- [x] Introduces a struct `DBSettings` that can be used to pass settings required for different RDBMS (might be refactored away in the future)
- [x] DB instantiation requires a logger

### DIFF
--- a/cmd/waku/flags.go
+++ b/cmd/waku/flags.go
@@ -288,6 +288,19 @@ var (
 		Destination: &options.Store.DatabaseURL,
 		EnvVars:     []string{"WAKUNODE2_STORE_MESSAGE_DB_URL"},
 	})
+	StoreMessageDBVacuum = altsrc.NewBoolFlag(&cli.BoolFlag{
+		Name:        "store-message-db-vacuum",
+		Usage:       "Enable database vacuuming at start. Only supported by SQLite database engine.",
+		Destination: &options.Store.Vacuum,
+		EnvVars:     []string{"WAKUNODE2_STORE_MESSAGE_DB_VACUUM"},
+	})
+	StoreMessageDBMigration = altsrc.NewBoolFlag(&cli.BoolFlag{
+		Name:        "store-message-db-migration",
+		Usage:       "Enable database migration at start.",
+		Destination: &options.Store.Migration,
+		Value:       true,
+		EnvVars:     []string{"WAKUNODE2_STORE_MESSAGE_DB_MIGRATION"},
+	})
 	StoreResumePeer = cliutils.NewGenericFlagMultiValue(&cli.GenericFlag{
 		Name:  "store-resume-peer",
 		Usage: "Peer multiaddress to resume the message store at boot. Option may be repeated",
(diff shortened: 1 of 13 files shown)


cmd/waku/flags.go
```diff
@@ -298,6 +298,19 @@ var (
 		Destination: &options.Store.DatabaseURL,
 		EnvVars:     []string{"WAKUNODE2_STORE_MESSAGE_DB_URL"},
 	})
+	StoreMessageDBVacuum = altsrc.NewBoolFlag(&cli.BoolFlag{
+		Name:        "store-message-db-vacuum",
+		Usage:       "Enable database vacuuming at start. Only supported by SQLite database engine.",
```
chaitanyaprem, line 303: Looks like postgres also supports vacuum https://www.postgresql.org/docs/current/sql-vacuum.html. 
Maybe we should allow this for postgres as well.

apentori: Yes it could be good to have it also on Postgres.

waku/persistence/postgres/postgres.go
```diff
@@ -35,13 +36,30 @@ func WithDB(dburl string, migrate bool) persistence.DBOption {
...
 // NewDB connects to postgres DB in the specified path
-func NewDB(dburl string) (*sql.DB, func(*sql.DB) error, error) {
+func NewDB(dburl string, shouldVacuum bool, logger *zap.Logger) (*sql.DB, func(*sql.DB) error, error) {
 	db, err := sql.Open("pgx", dburl)
 	if err != nil {
 		return nil, nil, err
 	}
 
+	if shouldVacuum {
+		err := executeVacuum(db, logger)
```
chaitanyaprem, line 57: Just wondering, since vacuuming of db can take time upto minutes. 
Should this be a bocking call or should we start this process in a go-routine and report the result back in async way? 
Do we have any test indicating how much time a vacuum takes for our store db?

richard-ramos: Reading how SQLite `VACUUM` documentation, it will prevent writes to be executed, so it should be a blocking call.
No tests, but this command works by creating a temporary copy of the db, so the time it will take depends on the amount of data stored in the DB.

For postgresql, I modified the PR to use `VACUUM FULL`, just so the behavior of the command is the same as SQLite, but I think we shouldnt expect people to use vacuum with PostgreSQL since it is possible to setup routine vacuuming in it https://www.postgresql.org/docs/current/routine-vacuuming.html#AUTOVACUUM

chaitanyaprem: Ok, maybe the caller can do an async status to the user then. Rather than us doing it. 
I am just concerned since this is an API call by the user, blocking it for soo long may not be a good idea.

But its ok for now, we can change it if we get user feedback.

chaitanyaprem (approved):
LGTM

PR #635, opened 2023-08-09
refactor: remove enc_publish functions and create separate encoding functions instead

richard-ramos:
# Description
Current c-bindings have encoding functions for both lightpush and relay that while at first view they appear to simplify the process of encoding messages using v1 encryption, in practice, they don't introduce benefits that make their existence worthy of the duplication of code and coupling lightpush and relay bindings to encoding functionality.

 This PR refactors away this issue by removing the encryption functions for lightpush and relay, and instead create two functions to perform this operation, similar to how decoding is done.

# Changes

- [x] Delete `waku_relay_publish_enc_asymmetric`, `waku_relay_publish_enc_symmetric`, `waku_lightpush_publish_enc_asymmetric` and `waku_lightpush_publish_enc_symmetric`
- [x] Add 2 new functions: `waku_encode_symmetric` and `waku_encode_asymmetric` whose output can be used directly in the publish functions of relay and lightpush
- [x] Update c-bindings example
- [x] Update the bindings README.md

### DIFF
--- a/library/encoding.go
+++ b/library/encoding.go
@@ -1,9 +1,14 @@
 package library
 
 import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
 	"encoding/json"
+	"errors"
 
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/crypto/secp256k1"
 	"github.com/waku-org/go-waku/waku/v2/payload"
 	"github.com/waku-org/go-waku/waku/v2/protocol/pb"
 	"github.com/waku-org/go-waku/waku/v2/utils"
@@ -12,14 +17,13 @@ import (
 func wakuMessage(messageJSON string) (*pb.WakuMessage, error) {
 	var msg *pb.WakuMessage
 	err := json.Unmarshal([]byte(messageJSON), &msg)
-	msg.Version = 0
 	return msg, err
 }
 
-func wakuMessageSymmetricEncoding(messageJSON string, symmetricKey string, optionalSigningKey string) (*pb.WakuMessage, error) {
+func EncodeSymmetric(messageJSON string, symmetricKey string, optionalSigningKey string) (string, error) {
 	msg, err := wakuMessage(messageJSON)
 	if err != nil {
-		return msg, err
+		return "", err
 	}
 
 	payload := payload.Payload{
@@ -31,7 +35,7 @@ func wakuMessageSymmetricEncoding(messageJSON string, symmetricKey string, optio
 
 	keyBytes, err := utils.DecodeHexString(symmetricKey)
 	if err != nil {
-		return msg, err
+		return "", err
 	}
 
 	payload.Key.SymKey = keyBytes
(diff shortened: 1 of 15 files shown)


library/encoding.go
```diff
@@ -69,28 +81,152 @@ func wakuMessageAsymmetricEncoding(messageJSON string, publicKey string, optiona
...
+	if err != nil {
+		return "", err
+	}
+
+	payload, err := payload.DecodePayload(&msg, keyInfo)
+	if err != nil {
+		return "", err
+	}
+
+	pubkey, signature := extractPubKeyAndSignature(payload)
```
Ivansete-status, line 159: I wonder if we could return better feedback in case of an error

library/encoding.go
```diff
@@ -69,28 +81,152 @@ func wakuMessageAsymmetricEncoding(messageJSON string, publicKey string, optiona
...
```
Ivansete-status, line 209: Likewise as per the previous comment

Ivansete-status (approved):
Hey ! Thanks for inviting me to the party ! I like it very much ! 

Is a great API simplification where the encryption and signing are protocol-agnostic, and also the error cases are well considered within the `library/encoding.go`.

Not a big deal at all but maybe in the future we could also split the encryption and signing a bit more.

Thanks for that !

chaitanyaprem:
Nice PR!

But I am wondering, this change would make it a 2 step API to the caller/user who want to send/receive encrypted messages. Wouldn't it be better to do encrypt/decrypt under the hood?

richard-ramos:
> @chaitanyaprem: But I am wondering, this change would make it a 2 step API to the caller/user who want to send/receive encrypted messages. Wouldn't it be better to do encrypt/decrypt under the hood when the user invokes enc_publish(relay/lightpush)?

Good question. I do have the following concerns if we follow that approach:
- the function interface is less elegant as you'd have to specify the pubsub topic, content topic, and timeout for publishing the message, besides the parameters required for the encryption.
- `enc_publish` would break the principle of single responsibility as it would deal with both encryption and publishing, and an error happening during the function execution would need to handle both publishing and encryption errors, while error handling with separate functions would arguably be easier to implement as you'd have a section for encryption errors, and a separate section for publishing errors.

chaitanyaprem:
> > @chaitanyaprem: But I am wondering, this change would make it a 2 step API to the caller/user who want to send/receive encrypted messages. Wouldn't it be better to do encrypt/decrypt under the hood when the user invokes enc_publish(relay/lightpush)?
> 
> Good question. I do have the following concerns if we follow that approach:
> 
>     * the function interface is less elegant as you'd have to specify the pubsub topic, content topic, and timeout for publishing the message, besides the parameters required for the encryption.
> 
>     * `enc_publish` would break the principle of single responsibility as it would deal with both encryption and publishing, and an error happening during the function execution would need to handle both publishing and encryption errors, while error handling with separate functions would arguably be easier to implement as you'd have a section for encryption errors, and a separate section for publishing errors.

I do agree that it makes the interface less elegant, can we have them pass a struct for the encryption params and another struct for publish params? That would make this cleaner. 
Since it is a cross-language API call, i am wondering if it is a good idea to have multiple calls across. Otherwise i agree with your single responsibility principle. 

If we don't have a good solution for your concerns, then  we can go ahead with this approach of splitting.

chaitanyaprem (approved):
LGTM apart from the comment.

PR #636, opened 2023-08-09
refactor: remove unused function and simplify code related to creating db and migrations

richard-ramos:
# Description
Minor refactor that removes the functions `WithDB` specific to each RDBMS, that are not used at all, and whose behavior can still be achieved by using `persistence.WithDB`; and reduce the number of returned parameters from `NewDB` while also reducing the responsibility of this function, as there is already a `Migrations` function that can be called, in case we want to obtain the migrations for postgresql/sqlite

# Changes

- [x] Removes `postgre.WithDB`
- [x] Removes `sqlite.WithDB`
- [x] `NewDB` will not return the migrations associated to the RDBMS
- [x] Renames `Migrate` to `Migrations` so the name of the functions reads better. i.e: `persistence.WithMigrations(sqlite.Migrations)` vs `persistence.WithMigrations(sqlite.Migrate)`

### DIFF
--- a/cmd/waku/rest/utils_test.go
+++ b/cmd/waku/rest/utils_test.go
@@ -12,10 +12,10 @@ import (
 
 func MemoryDB(t *testing.T) *persistence.DBStore {
 	var db *sql.DB
-	db, migration, err := sqlite.NewDB(":memory:", false, utils.Logger())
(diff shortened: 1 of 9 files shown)


chaitanyaprem (approved):
LGTM

PR #639, opened 2023-08-10
chore(filter2): test updates

siphiuel:
# Description
Includes a refactor of filter v2 tests, plus some minor changes taken from js-waku (https://github.com/waku-org/js-waku/blob/master/packages/tests/tests/filter.node.spec.ts)

Many changed files are due to function renamings done for consistency - not sure if this was the right call.

### DIFF
--- a/waku/v2/protocol/filter/server.go
+++ b/waku/v2/protocol/filter/server.go
@@ -45,8 +45,8 @@ type (
 	}
 )
 
-// NewWakuFilterFullnode returns a new instance of Waku Filter struct setup according to the chosen parameter and options
-func NewWakuFilterFullnode(timesource timesource.Timesource, log *zap.Logger, opts ...Option) *WakuFilterFullNode {
+// NewWakuFilterFullNode returns a new instance of Waku Filter struct setup according to the chosen parameter and options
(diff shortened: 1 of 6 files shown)


waku/v2/protocol/filter/server.go
```diff
@@ -69,6 +69,10 @@ func (wf *WakuFilterFullNode) SetHost(h host.Host) {
 	wf.h = h
 }
 
+func (wf *WakuFilterFullNode) Host() host.Host {
```
richard-ramos, line 72: The `Host` function in here, lightnode and relay aren't necessary as you would have access to the host via the WakuNode.

richard-ramos (commented):
Thank you for the PR. Left some minor comments:

richard-ramos:
@siphiuel, lint and build steps are failing.

waku/v2/protocol/filter/server.go
siphiuel: Understood, done.

siphiuel:
@richard-ramos fixed lint and build

richard-ramos (approved):
:rocket:

PR #640, opened 2023-08-10
fix: nwaku / go-waku rlnKeystore compat

richard-ramos:
# Description
Makes go-waku rln implementation compatible with nwaku's

# Changes
- [ ] Stores chainID in uppercase hex
- [ ] Renames `rlnCredentials.json` to `rlnKeystore.json`
- [ ] Changes credential applicationName to `nwaku-rln-relay"`

cc: @rymnc What do you think if we change in nwaku and go-waku the application name to `waku-rln-relay` to make it implementation agnostic?


chaitanyaprem (approved):
LGTM

chaitanyaprem:
> cc: @rymnc What do you think if we change in nwaku and go-waku the application name to waku-rln-relay to make it implementation agnostic?
This should be changed.

@richard-ramos , looks like nix build failed..that also needs to be addressed.

rymnc:
Good idea @richard-ramos, I'll create an issue for it

PR #648, opened 2023-08-16
ci: add RACE parameter to fix conditional

jakubgs:
We are checking if `JOB_NAME` contains a token `race` to set the default value for `RACE` parameter which decides which tests run.


chaitanyaprem (approved):
LGTM!

PR #656, opened 2023-08-18
fix: log output and encoding

### LINKED ISSUE #649: bug: CLI flag for log-output is not working with file as option
**Describe the bug**
If we run waku node with --log-output as file or file:<filename>, logs are still printed only on stdout. 

**To Reproduce**
Build and run waku node with following options

./build/waku -l DEBUG --log-output file 

./build/waku -l DEBUG --log-output file:<filename>

**Expected behavior**
Logs should be written to file provided or in some default filename in current path.

richard-ramos:
# Description
Adds back code that I deleted by mistake in https://github.com/waku-org/go-waku/commit/61cba076bb0a585a33b8fc3187c3b780b0e611b1#diff-1365cc75829d0357d9c3adbd9fbc213f47e9ad86aa02e9e24f1513d84e7de037L109

## Issue

closes #649

### DIFF
--- a/cmd/waku/node.go
+++ b/cmd/waku/node.go
@@ -83,6 +83,10 @@ const dialTimeout = 7 * time.Second
 
 // Execute starts a go-waku node with settings determined by the Options parameter
 func Execute(options NodeOptions) {
+	// Set encoding for logs (console, json, ...)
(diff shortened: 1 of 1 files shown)


chaitanyaprem (approved):
LGTM

PR #669, opened 2023-08-23
feat: autosharding core logic

chaitanyaprem:
# Description
Implemented basic autosharding logic based on current nwaku master. 
Note that algo mentioned as per https://github.com/waku-org/nwaku/pull/1854 has been removed probably to enhance in the future(which talks about bias etc). 

# Changes

- [x] Updated content topic and pubsub topic related code
- [x] Hashing algo to arrive at pubsubTopic from a content-topic as per rfc 51


waku/v2/protocol/content_topic.go
```diff
@@ -0,0 +1,110 @@
...
+func DefaultOptions() []ContentTopicOption {
+	return []ContentTopicOption{
+		WithGeneration(0),
+	}
+}
+
+func (ct ContentTopic) Equal(ct2 ContentTopic) bool {
+	return ct.ApplicationName == ct2.ApplicationName && ct.ApplicationVersion == ct2.ApplicationVersion &&
+		ct.ContentTopicName == ct2.ContentTopicName && ct.Encoding == ct2.Encoding
+}
```
richard-ramos, line 70: Should the `ContentTopicParams` be taken into account in this function?

richard-ramos (approved):
LGTM!

waku/v2/protocol/content_topic.go
chaitanyaprem: Yup, we should include that as well. Although for now it won't make a difference. 
Will update.

Issue #655, opened 2023-08-18
RLN enabled by default


richard-ramos:
**Planned start date**:
**Due date**:

# Summary

RLN validation should be enabled by default for relay nodes

# Acceptance Criteria

## Tasks
- [x] Run rln-relay on all configured pubsub topics and content topics (passed in w/ `--pubsub-topic`)
- [ ] Separate out bandwidth constraint validator 
- [x] Ordered validator execution
  1. WakuMessage decoding
  2. WakuRelay
  3. WakuRlnRelay
- [x] Review metrics (few too many one-time metrics, like startup time etc)
- [x] Clear up the nullifier table (clear the table once in 10 epochs or so)
- [x] Health check api (`/health` should denote if the nwaku node is ready for use)
- [x] Sync from the deployment block instead of from genesis. 
- [x] The `rln` feature should be available in the non-experimental context
- [x] Add metric to show the amout of registered memberships in the tree.
- [x] Rename rlnkeystore application to `waku-rln-relay`
- [x] Remove registration capability from go-waku
- [x] RLN DB should be aware of chain and contract address
- [x] Command to generate keystores
- [x] Add metric with the total registered memberships in the contract 
- [x] `s/flush_interval/flush_every_ms` misconfiguration on rln config 
- [x] DB should be aware of chain and contract address 
- [x] DB dissapears (use zerokit v0.3.2)
- [x] Sync valid roots to metadata
- [x] Non deterministic sorting applied to credentials, and contract
- [ ] Mobile support for RLN https://github.com/waku-org/go-zerokit-rln/issues/5 
- [ ] C-Bindings support for RLN #732 
- [x] Nix build support for RLN #731 

# RAID (Risks, Assumptions, Issues and Dependencies)

richard-ramos:
**Weekly Update**

- _achieved_: removed registration capability from the wakunode and created a separate subcommand to do the registration
- _next_: run rln-relay on all configured pubsub topics and content topics

richard-ramos:
**Weekly Update**

- _achieved_: run rln-relay on all configured pubsub topics and content topics, added metrics, made RLN database aware of chainID and contract address, refactored keystore.
- _next_:  test keystore interop with nwaku, integrate waku rln registry, and restore valid roots from DB

richard-ramos:
**Weekly Update**

- _achieved_: test keystore interop with nwaku, integrate waku rln registry, and restore valid roots from DB
[...]

PR #670, opened 2023-08-23
fix(rln-relay): RLN DB should be aware of chain and contract address

richard-ramos:
# Description
See https://github.com/waku-org/nwaku/issues/1921 for reasoning behind this fix

# Changes

- Adds chainID and contract address to RLN metadata
- Adds test unit to verify metadata serialization 
- Returns an error during startup if metadata does not match configuration settings

## Issue

part of #655


rymnc:
wow, this was really quick! Kudos!

chaitanyaprem (approved):
LGTM

PR #671, opened 2023-08-23
fix(rln-relay): flush_interval incorrectly set

richard-ramos:
# Description
See https://github.com/waku-org/nwaku/pull/1933 for motivation behind this change

# Changes

- [ ] Updates go-zerokit-rln to a version that uses `flush_every_ms` instead of `flush_interval`
- [ ] Uses a `time.Duration` to set the interval instead of an `uint`


chaitanyaprem (approved):
LGTM

PR #675, opened 2023-08-23
refactor: keystore as a data type and store credentials into a map

richard-ramos:
# Description
Extracts the rln keystore to be its own data type so the dynamic group manager does not have to worry about the keystore path and its existence. Also, uses the structure proposed in https://github.com/waku-org/nwaku/issues/1922#issuecomment-1692137198 and https://github.com/waku-org/nwaku/issues/1922#issuecomment-1692192077

# Changes
- [x] Extracts keystore functionality to an `AppKeystore struct`
- [x] Uses a map instead of an array to store credentials
- [x] Removes the flag `rln-relay-membership-index` from wakunode and chat2
- [x] Simplifies API of `GetMembershipCredentials` function
- [x] Ensures credentials can be used in both nwaku and go-waku

### DIFF
--- a/waku/v2/protocol/rln/keystore/keystore.go
+++ b/waku/v2/protocol/rln/keystore/keystore.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"sort"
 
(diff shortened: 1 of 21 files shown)


richard-ramos:
Some things that are pending or need to be verified:
1 - Does the hash used as key have a 0x prefix?
2 - In `AddMembershipCredential`, is it possible to have different identity credentials within the same SHA? (probably not)
3 - In `AddMembershipCredential`,  replace `newIdentityCredential` and `newMembershipGroup` by a single `MembershipCredentials` parameter

richard-ramos:
I updated this PR with the format for the key `sha256(chainId|contractAddress|treeIndex)` from https://github.com/waku-org/nwaku/pull/1956 and it's now ready for review!

waku/v2/protocol/rln/keystore/keystore.go
```diff
@@ -2,324 +2,171 @@ package keystore
...
 	"github.com/ethereum/go-ethereum/accounts/keystore"
+	"github.com/waku-org/go-waku/waku/v2/hash"
 	"github.com/waku-org/go-zerokit-rln/rln"
 	"go.uber.org/zap"
 )
 
-const RLN_CREDENTIALS_FILENAME = "rlnKeystore.json"
-const RLN_CREDENTIALS_PASSWORD = "password"
+// DefaultCredentialsFilename is the default filename for the rln credentials keystore
+const DefaultCredentialsFilename = "rlnKeystore.json"
```
chaitanyaprem, line 19: Maybe we should prefix it with `./`

waku/v2/protocol/rln/keystore/types.go
```diff
@@ -0,0 +1,118 @@
...
+
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/waku-org/go-zerokit-rln/rln"
+	"go.uber.org/zap"
+)
+
+// MembershipContract contains information about a membership smart contract address and the chain in which it is deployed
+type MembershipContract struct {
```
chaitanyaprem, line 16: Maybe better to rename it to MembershipContractInfo

chaitanyaprem (approved):
LGTM

PR #683, opened 2023-08-25
refactor: peerManager test

harsh-98:
# Description
`AddPeerToServiceSlot` is exposed to prevent cyclic dependency bcz peerManager is used in [test/utils](https://github.com/waku-org/go-waku/blob/master/tests/utils.go#L20) for creating TestPeerDiscover and test/utils is used in test of peerManager in [subfolder](https://github.com/waku-org/go-waku/tree/master/waku/v2/peermanager/test). 

# Changes

- add mock peer discoverer to peermanager folder. 

# Tests

Issue #668, opened 2023-08-23
feat: PeerManager scoring peers and handling connections.


harsh-98:
### Background
PeerManager's responsibility is to handle peers  for "/vac/waku/relay/2.0.0" topic and otherTopics in serviceSlots. It uses peerConnector for connecting to new Peers via dialPeer on basic.basichost in libp2p. 
Whenever there are more inward connections than the limit, pm closes those connections via pruneInRelayConns(There is also removePeer fn in peerManager that's not used in go-waku repo, i think we missed calling it from pruneInRelayConns to remove these peers from serviceSlots and peerStore?). 
Currently there is no intelligent way of selecting InPeers to close. There is a `TODO` in peer_manager.go to add score. 

### Details
host.Network() returns [libp2p-swarm](https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/swarm.go#L206), which has ConnsToPeer to get all active connections to peer. Each connection has flags like `Transient`(closing soon), `Direct`(transport is proxy).
For each peer we can calculate a score such as : [betterConnectionFlags](https://github.com/libp2p/go-libp2p/blob/master/p2p/net/swarm/swarm.go#L497)
```
Direct: 1 InDirect:0
Transient:0  Lasting:1

                               Score: sigma(Conn.Flags)
```
This calculation doesn't actively fetch the transient or direct status for each connection. But uses the previously stored values for that connection, reducing computation time. 
##### Other scoring criteria:
- There is also `Opened` flag for timestamp when the connection was opened( if needed we can also add this to score calc). 
- I think transient/lasting flags' score can be more than direct/indirect flags.

Improvements for serviceSlots:

- ServiceSlot isn't designed to be thread safe. 
- Whenever [a peer is removed from serviceSlots](https://github.com/waku-org/go-waku/blob/master/waku/v2/peermanager/peer_manager.go#L280), we unset the array slot from peerId to empty string. This leads to a bug where [pm.SelectPeer](https://github.com/waku-org/go-waku/blob/master/waku/v2/peermanager/peer_manager.go#L327) can return empty string as [utils.SelectRandomPeer](https://github.com/waku-org/go-waku/blob/master/waku/v2/utils/peer.go#L60) doesn't filter for empty string.

### Acceptance criteria
- add calculateScoreFn to calculate score for all inRelayPeers.
- ServiceSlot can be data struct with `map[proto]struct{sync.mutex, ids: map[peerId]bool}`.
[...]

PR #684, opened 2023-08-25
feat: add serviceSlot ds

harsh-98:
# Description
Related to #668

# Changes

- fix empty string peer that can be return when that peer is removed from serviceSlot.
- add mutexes for thread safe ds.
- add comments for servicePeer/relayPeers and inRelayPeer/outRelayPeers.

# Tests

- service_slot_Test added .
- solution for failing test #683 TestAdditionAndRemovalOfPeer

### DIFF
--- a/waku/v2/protocol/filter/subscriptions_map.go
+++ b/waku/v2/protocol/filter/subscriptions_map.go
@@ -225,7 +225,7 @@ func iterateSubscriptionSet(logger *zap.Logger, subscriptions SubscriptionSet, e
 			defer subscription.RUnlock()
 
 			_, ok := subscription.ContentTopics[envelope.Message().ContentTopic]
-			if !ok && len(subscription.ContentTopics) != 0 {
+			if !ok { // only send the msg to subscriptions that have matching contentTopic
 				return
 			}
 


waku/v2/peermanager/peer_manager.go
```diff
@@ -24,47 +24,49 @@ const WakuRelayIDv200 = protocol.ID("/vac/waku/relay/2.0.0")
...
 type PeerManager struct {
 	peerConnector       *PeerConnectionStrategy
-	maxConnections      int
 	maxRelayPeers       int
 	logger              *zap.Logger
 	InRelayPeersTarget  int
 	OutRelayPeersTarget int
 	host                host.Host
-	serviceSlots        map[protocol.ID][]peer.ID
+	serviceSlots        *ServiceSlot
```
chaitanyaprem, line 32: Better to rename the type itself to ServiceSlots.

waku/v2/peermanager/peer_manager.go
```diff
@@ -24,47 +24,49 @@ const WakuRelayIDv200 = protocol.ID("/vac/waku/relay/2.0.0")
...
 
-const maxRelayPeersShare = 5
-
-// const defaultMaxOutRelayPeersTarget = 10
-const outRelayPeersShare = 3
 const peerConnectivityLoopSecs = 15
-const minOutRelayConns = 10
+
+// 80% relay peers 20% service peers
+func relayAndServicePeers(maxConnections int) (int, int) {
```
chaitanyaprem, line 39: Should rename to ServiceAndRelayPeers as first return is for servicePeers. 
Also add comments or have named returns to make code readable.

waku/v2/peermanager/peer_manager.go
```diff
@@ -296,7 +292,7 @@ func (pm *PeerManager) AddPeerToServiceSlot(proto protocol.ID, peerID peer.ID) {
 	//TODO: Ideally we should sort the peers per service and return best peer based on peer score or RTT etc.
```
chaitanyaprem, line 292: Move this comment into getPeers method in serviceSlot.

waku/v2/peermanager/peer_manager.go
```diff
@@ -310,19 +306,15 @@ func (pm *PeerManager) SelectPeer(proto protocol.ID, specificPeers []peer.ID, lo
 	//  - which topics they track
 	//  - latency?
 
+	//Try to fetch from serviceSlot
+	if peerID, err := pm.serviceSlots.getPeers(proto).getRandom(); err == nil {
```
chaitanyaprem, line 310: First we should not try to fetch from serviceSlot.
Because relay peers are not added to serviceSlots. So, first we should check if proto is relay then return a random peer from relayPeers. If it is not relay, then proceed with service slot selection.

waku/v2/peermanager/service_slot.go
```diff
@@ -0,0 +1,75 @@
...
+	m  map[peer.ID]struct{}
+}
+
+func newPeerMap() *peerMap {
+	return &peerMap{
+		m: map[peer.ID]struct{}{},
+	}
+}
+
+func (pm *peerMap) getRandom() (peer.ID, error) {
```
chaitanyaprem, line 22: This is not using randomselection logic. 
It is just returning always first one. Need to fix this to return a random one from list of peers in serviceSlots.

waku/v2/peermanager/service_slot.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"sync"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/core/protocol"
+	"github.com/waku-org/go-waku/waku/v2/utils"
+)
+
+type peerMap struct {
```
chaitanyaprem, line 11: Any specific reason for choosing a map over an array here? 
Having an array makes it easy to choose a peer randomly. With a map selecting at random would be difficult.

waku/v2/peermanager/service_slot.go
```diff
@@ -0,0 +1,75 @@
...
+	slots.mu.Lock()
+	defer slots.mu.Unlock()
+	if slots.m[proto] == nil {
+		slots.m[proto] = newPeerMap()
+	}
+	return slots.m[proto]
+}
+
+// RemovePeer for removing peer ID for a given protocol
+func (slots *ServiceSlot) RemovePeer(peerID peer.ID) {
```
chaitanyaprem, line 69: Better to limit this function also to package scope?

waku/v2/peermanager/service_slot.go
chaitanyaprem: Interesting, was not aware of this part. 
Interesting they enforce randomness in iterator.

waku/v2/peermanager/peer_manager.go
chaitanyaprem: But why to do a map lookup unnecessarily which is costly, rather a check for relay would be less costlier. 
Hence i was suggesting the other way around.

chaitanyaprem: These are 2 flows i am looking up for selectPeer 
1. with relay -> where a random peer has to be selected from peerStore
2. for service-protocols -> lookup in service-slot to get a peer randomly, if none found then lookup in peer-table(this case should almost never happen). 

If we just check for relayProto first and proceed in case of relay, it would just be a string compare. 
Not sure why FilterPeersByProto would come into picture here. 
And for servicePeer selection, the overhead is just 1 string-compare. 

Now that i look at the code further, in case of relay a call to `getPeers` would end up creating a `serviceSlot` for relay which is not what we want. 
Also, the cost of acquiring a lock on serviceSlot is un-necessary for relay. Instead of all this overhead shouldn't it be easier to check for relay first which is a stringCompare. And if proto is not relay, then lookup in service slot?

Also, i don't think it is a good idea to insert when a GetMethod is called on serviceSlots.

5lliot: This is the previous excerpt of code. 
https://github.com/waku-org/go-waku/blob/master/waku/v2/utils/peer.go#L46-L55. In the SupportedProtocols, we filter protocols for the peer that are needed. Which is filterProtocol*noOfPeers.
image
Sorry , I not able to find the single StringCompare that you are talking about.

getPeers create a serviceSlot, though that slot will not have peers for DefaultRelay. I can return nil from getPeers for DefaultRelay but that will external check.

chaitanyaprem: OK, i see what you are getting at. 
I was referring to line "proto == WakuRelayIDv200" but we do this check after filtering for peers. 
But now since we are looking into serviceSlots first , for servicePeers we don't need to do this check.

Let us change this behaviour of not creating the serviceSlot while calling getPeers. Especially if it is going to create one for relay, would lead to confusion. 

@5lliot are you @harsh-98 using a different username? Got confused suddenly :)

chaitanyaprem (approved):
LGTM apart from minor unresolved comment.

PR #686, opened 2023-08-25
fix(filterV2): requestID and log request type

richard-ramos:
# Description
The requestID was not being populated when unsubscribing from filter.
I also added the request type to the logs since it's useful for debugging


chaitanyaprem (commented):
LGTM

Issue #634, opened 2023-08-08
[Epic] Maintenance


richard-ramos:
# Summary

Maintenance items that do not directly fit under any other milestones, yet need to be done to avoid problems.
Typically:
- dependencies update (so that we do not wait to _need_ the upgrade to do it)
- bugs reported by users of go-waku
- Releasing

# RAID (Risks, Assumptions, Issues and Dependencies)

- This may become a dumping ground, let's thrive to not let this happen.

richard-ramos:
**Weekly Update**

- _achieved_: improved keep alive logic for identifying if machine is waking up; added vacuum feature to sqlite and postgresql; made migrations optional; refactored db and migration code, extracted code to generate node key to its own separate subcommand
- _next_: ?

richard-ramos:
**Weekly Update**

- _achieved_: refactored wakuv2 metrics to make each protocol responsable for registering and defining its own metrics

richard-ramos:
**Weekly Update**

- _achieved_: allow mixing named and static shards, logs succesful message pushes, concurrency fixes for filterv2

richard-ramos:
**Weekly Update**

- _achieved_: add tls/ws to address factory and log ENRs only after they have been setup

chaitanyaprem:
**Weekly updates**
- _achieved_: fix panic observed in peer-manager, update filter protocol as per rfc.

fryorcraken:
Maintenance work does not need to be tracked with an Epic. Ref: https://github.com/waku-org/pm/pull/91

PR #687, opened 2023-08-25
fix(discv5): threadsafe peerCh

richard-ramos:
# Description
Fixes https://github.com/status-im/status-mobile/issues/17115 and CI error on https://github.com/status-im/status-go/pull/3928

h/t @cammellos for the code :) 

Part of https://github.com/waku-org/go-waku/issues/634

### DIFF
--- a/waku/v2/discv5/discover.go
+++ b/waku/v2/discv5/discover.go
@@ -34,16 +34,16 @@ type PeerConnector interface {
 }
 
 type DiscoveryV5 struct {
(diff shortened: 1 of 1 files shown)


chaitanyaprem (approved):
LGTM.
Nice pattern. Was not aware of this recover function in go, very useful especially in current use-case.
Since there are multiple places where we have peerChannel, i am wondering if we should also reuse this everywhere and make this peerChannel move into peerManager package.

PR #688, opened 2023-08-28
chore: fix frequent logging from poll

chaitanyaprem:
# Description
Change logLevel of log during to 1 second polling. 
Moved info log to relevant location.

@harsh-98 , let's keep in mind number of info logs we generate.


chaitanyaprem:
Merged this without review as change is trivial.

PR #692, opened 2023-08-28
refactor(filter): unsubscribe waitgroup and async

richard-ramos:


richard-ramos:
@chaitanyaprem, @harsh-98 if this pattern looks fine, in future PRs i'll add the same logic to other protocols that could run into the same issue.

cammellos (commented):
Looks good, I would maybe add some tests to verify the behavior?

richard-ramos:
Added some tests to verify behaviour!

chaitanyaprem (approved):
LGTM apart from minor comments.

chaitanyaprem:
> @chaitanyaprem, @harsh-98 if this pattern looks fine, in future PRs i'll add the same logic to other protocols that could run into the same issue.

Yup, makes sense. 
If so, rather than copying this pattern..would it be possible to have this common pattern applied as part of some decorator rather than duplicating it completely.

cammellos:
@richard-ramos with regards of the race condition between `Start`/`Stop`, a panic can still be replicated:

https://github.com/waku-org/go-waku/compare/async-filter-wg-test?expand=1


```
2023-08-29T11:20:23.492+0100    INFO    gowaku.filterv2-lightnode       filter/client.go:85     starting again
2023-08-29T11:20:23.492+0100    INFO    gowaku.filterv2-lightnode       filter/client.go:89     starting
2023-08-29T11:20:23.492+0100    INFO    gowaku.filterv2-lightnode       filter/client.go:113    stopping node
2023-08-29T11:20:23.492+0100    INFO    gowaku.filterv2-lightnode       filter/client.go:92     setting cancel
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xe89bbb]

goroutine 302 [running]:
github.com/waku-org/go-waku/waku/v2/protocol/filter.(*WakuFilterLightNode).Stop(0x0?)
        /home/cammellos/documents/programming/status/go-waku/waku/v2/protocol/filter/client.go:115 +0x5b
github.com/waku-org/go-waku/waku/v2/protocol/filter.(*FilterTestSuite).TestStartStop.func1()
        /home/cammellos/documents/programming/status/go-waku/waku/v2/protocol/filter/filter_test.go:366 +0x38
created by github.com/waku-org/go-waku/waku/v2/protocol/filter.(*FilterTestSuite).TestStartStop
        /home/cammellos/documents/programming/status/go-waku/waku/v2/protocol/filter/filter_test.go:382 +0x119
exit status 2

```

That would happen if you call `Start` for the first time, and before it finishes `Stop` gets called.
Similarly with `Unsubscribe`


```
goroutine 320 [running]:
github.com/waku-org/go-waku/waku/v2/protocol/filter.(*WakuFilterLightNode).Unsubscribe(0xc000410c80, {0x14e0d68?, 0xc000467020}, {{0x1140fb8, 0x9}, {0xc0003f7640, 0x1, 0x1}}, {0xc00047df98, 0x1, ...})
        /home/cammellos/documents/programming/status/go-waku/waku/v2/protocol/filter/client.go:382 +0xdf
github.com/waku-org/go-waku/waku/v2/protocol/filter.(*FilterTestSuite).TestStartStop.func1()
        /home/cammellos/documents/programming/status/go-waku/waku/v2/protocol/filter/filter_test.go:369 +0xd9
created by github.com/waku-org/go-waku/waku/v2/protocol/filter.(*FilterTestSuite).TestStartStop
        /home/cammellos/documents/programming/status/go-waku/waku/v2/protocol/filter/filter_test.go:382 +0x1b7
exit status 2
```

You can replicate using the command:

`go test -run TestFilterSuite -testify.m TestStartStop`

and you can check for detected race conditions using

`go test -run TestFilterSuite -testify.m TestStartStop -race`

The issue is that `CompareAndSwap` only protect the beginning of the function, but after that, all bets are off.
This currently only happens between the first `Start` and `Stop`, if `Start` successfully finishes, then I could not replicate race conditions, since state is not nullled, though there might be some inconsistencies, as the node might not be fully set up once one of these calls are issued.

Probably worth discussing how to address in the call, locks are obviously a solution, but we need to make sure we don't deadlock, but we can discuss later.

richard-ramos:
> That would happen if you call Start for the first time, and before it finishes Stop gets called.

> The issue is that CompareAndSwap only protect the beginning of the function, but after that, all bets are off.
This currently only happens between the first Start and Stop, if Start successfully finishes, then I could not replicate race conditions, since state is not nullled, though there might be some inconsistencies, as the node might not be fully set up once one of these calls are issued.

:thinking: interesting. We do use CompareAndSwap in discv5, so I wonder if this scenario could happen there too. 
cc: @harsh-98

richard-ramos:
@cammellos with latest commit, I copied your test and replaced the instances of CompareAndSwap in filter client and it is not panicking anymore!

waku/v2/protocol/filter/client.go
```diff
@@ -206,6 +236,13 @@ func (wf *WakuFilterLightNode) request(ctx context.Context, params *FilterSubscr
 
 // Subscribe setups a subscription to receive messages that match a specific content filter
 func (wf *WakuFilterLightNode) Subscribe(ctx context.Context, contentFilter ContentFilter, opts ...FilterSubscribeOption) (*SubscriptionDetails, error) {
+	wf.Lock()
```
chaitanyaprem, line 239: I am wondering having a lock at this high level would slow down the API calls if multiple routines want to invoke them. Because ideally we would want the app to use these API's in multiple routines and not wait for 1 API to finish (which includes sending message out), rather only lock parts of code which we feel would cause concurrency issues.
We should have fine-grained locking as below to address this:
1. To protect only start/stop 
2. Have a separate lock to protect only subscriptionsMap access. Similar to what @harsh-98 has done for peerMap in serviceSlots #684 
I don't think we have any other data that we want to protect other than these 2.

chaitanyaprem (approved):
LGTM. 
A comment regarding probable impact on performance due to this lock.

waku/v2/protocol/filter/client.go
richard-ramos: Hm, the problem I see is what happens if there is a race condition when someone calls Subscribe/Unsubscribe and Stop at same time, which was the initial issue that originated this PR. 

I'm thinking that perhaps I should change the lock type to an RLock with the idea of allowing concurrent reads, since the lock of `WakuFilterLightNode` is intended to protect only the `started` attribute

chaitanyaprem (approved):
LGTM

PR #694, opened 2023-08-29
chore: log succesful message pushes

richard-ramos:
# Description
one suggestion that was made during today's message reliability call was to add this information to help debug cases where a message appears to have been pushed succesfully to all subscribers yet the message appears to not have been delivered.

To see this change in action, it would require the gowaku fleet to be deployed 

Notice that might add some noise on the logs for full filter nodes, due to these messages being INFO level, but it should be an acceptable tradeoff while dogfooding for filter is in progress. We can do a separate PR in the future to change the log level to DEBUG later

### DIFF
--- a/waku/v2/protocol/filter/server.go
+++ b/waku/v2/protocol/filter/server.go
@@ -267,7 +267,12 @@ func (wf *WakuFilterFullNode) filterListener(ctx context.Context) {
 }
 
 func (wf *WakuFilterFullNode) pushMessage(ctx context.Context, peerID peer.ID, env *protocol.Envelope) error {
-	logger := wf.log.With(logging.HostID("peer", peerID))
+	logger := wf.log.With(
+		logging.HostID("peer", peerID),
+		logging.HexBytes("envelopeHash", env.Hash()),
+		zap.String("pubsubTopic", env.PubsubTopic()),
+		zap.String("contentTopic", env.Message().ContentTopic),
+	)
 
 	messagePush := &pb.MessagePushV2{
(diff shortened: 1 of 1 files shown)


waku/v2/protocol/filter/server.go
```diff
@@ -298,12 +303,15 @@ func (wf *WakuFilterFullNode) pushMessage(ctx context.Context, peerID peer.ID, e
 		} else {
 			wf.metrics.RecordError(writeResponseFailure)
 		}
-		logger.Error("pushing messages to peer", logging.HexBytes("envelopeHash", env.Hash()), zap.String("pubsubTopic", env.PubsubTopic()), zap.String("contentTopic", env.Message().ContentTopic), zap.Error(err))
+		logger.Error("pushing messages to peer", zap.Error(err))
```
chaitanyaprem, line 306: We should also log message specific details in error if we want to trace the message isn't it? 
Otherwise logs will be hard to analyze as there maybe interleaving from multiple calls to pushMessage.
Infact we should always log message details in error logs. That way anytime debugging becomes easier as message metadata is logged.

richard-ramos: Besides the fields added to the `logger` in lines 270 to 275, what other metadata do you think it would be useful to add?

chaitanyaprem: Those fields should be sufficient. 
Ah, my bad. I mistook logger being appended with info log.

chaitanyaprem (approved):
LGTM

PR #695, opened 2023-08-30
fix: dont send to subscriptions that don't have any ContentTopic

harsh-98:
# Description
dont send to subscriptions that don't have any ContentTopic


chaitanyaprem (approved):
LGTM

PR #700, opened 2023-08-31
chore: exclude directories from code climate

richard-ramos:
Excluding some examples from code climate.


richard-ramos:
Insta-merging, as this is a trivial change, and running code climate in some examples adds noise to the report.
Having said that, probably at some point of the future, we should fix the problems in the c# and swift examples

PR #701, opened 2023-08-31
fix(rln-relay): sync from deployed block number

richard-ramos:
# Description
Sync events from the deployment block instead of doing it from block 0

## Issue
Part of #655


chaitanyaprem (approved):
LGTM

PR #702, opened 2023-09-01
feat: store and retrieve valid merkle roots in RLN db

richard-ramos:
# Description
When a node is restarted, and has access to a persisted tree, the window of acceptable roots is not
persisted, and may result in an empty deque. To mitigate this bug, we are persisting the window of acceptable roots in
the rln metadata.

## Issue

Part of #655

### DIFF
--- a/waku/v2/protocol/rln/group_manager/root_tracker.go
+++ b/waku/v2/protocol/rln/group_manager/root_tracker.go
@@ -7,11 +7,14 @@ import (
 	"github.com/waku-org/go-zerokit-rln/rln"
 )
 
+// RootsPerBlock stores the merkle root generated at N block number
 type RootsPerBlock struct {
-	root        rln.MerkleNode
-	blockNumber uint64
+	Root        rln.MerkleNode
+	BlockNumber uint64
 }
 
+// MerkleRootTracker keeps track of the latest N merkle roots considered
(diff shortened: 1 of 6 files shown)


chaitanyaprem (approved):
LGTM

waku/v2/protocol/rln/group_manager/root_tracker.go
```diff
@@ -125,29 +132,50 @@ func (m *MerkleRootTracker) pushRoot(blockNumber uint64, root [32]byte) {
...
 	}
 
 	return result
 }
 
+// Buffer is used as a repository of older merkle roots that although
+// they were valid once, they have left the acceptable window of
+// merkle roots. We keep track of them in case a chain fork occurs
+// and we need to restore the valid merkle roots to a previous point
+// of time
```
chaitanyaprem, line 154: I am wondering if the following scenario would cause an issue.
Since we are not persisting buffer, what if a node only persists latest roots and restarts and there is a chain fork in between restart. how will the node recover its validRoots? 
Should we persist buffer or identify if there was a chain fork after restart and resync from deployed block? Not sure which is costly

richard-ramos: cc: @rymnc

PR #703, opened 2023-09-01
fix: add back to peerstore dns4 addresses removed by go-libp2p identify

richard-ramos:
# Description
While playing with circuit relay, I noticed that I wasn't getting assigned any circuit relay address. Exploring the source code of go-libp2p, i noticed that Identify is filtering out dns4 addresses. Adding back an address in this case just means adding the address with a `ConnectedAddrTTL` 
See https://github.com/libp2p/go-libp2p/issues/2550 for more details

### DIFF
--- a/waku/v2/node/wakunode2.go
+++ b/waku/v2/node/wakunode2.go
@@ -740,6 +740,20 @@ func (w *WakuNode) connect(ctx context.Context, info peer.AddrInfo) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


richard-ramos:
I'm merging this PR immediatly, as this is very problematic since clients will not be able to obtain circuit relay addresses, and this solution works:tm: even tho it can probably be improved.

PR #704, opened 2023-09-01
fix: loop counter for writing multiaddresses in enr

richard-ramos:
# Description
Writing multiaddresses on the ENR `multiaddr` key worked only when there was more than one multiaddress field that should be stored on that field. Most of the time that wasn't a problem, specially since desktop does not use websocket addresses and always has at least 2 circuit relay addresses

### DIFF
--- a/waku/v2/protocol/enr/localnode.go
+++ b/waku/v2/protocol/enr/localnode.go
@@ -35,7 +35,7 @@ func WithMultiaddress(multiaddrs ...multiaddr.Multiaddr) ENROption {
 		failedOnceWritingENR := false
 		couldWriteENRatLeastOnce := false
 		successIdx := -1
-		for i := len(multiaddrs) - 1; i >= 0; i-- {
+		for i := len(multiaddrs); i > 0; i-- {
 			err = writeMultiaddressField(localnode, multiaddrs[0:i])
 			if err == nil {
 				couldWriteENRatLeastOnce = true


chaitanyaprem (approved):
LGTM

richard-ramos:
> * if the successInd is found, we are trying to write the multiAddress field to the local node again.

The reason why this is done this way is because a failure to write the ENR will still modify the `multiaddrs` fields, making the ENR unusable, so `successInd` is used to revert the `multiaddrs` field back to the last usable state. This could probably be improved by having `writeMultiaddressField` do the revert instead of having it done in this file. 

I can do this in a separate PR:

i.e. - instead of doing it this way:
```go
defer func(){
	if e := recover(); e != nil {
		// Deleting the multiaddr entry, as we could not write it succesfully
		localnode.Delete(enr.WithEntry(MultiaddrENRField, struct{}{}))
		err = errors.New("could not write enr record")
	}
}()
```

to instead do:
```go
var ogMultiaddrValue []byte
if err := node.Record().Load(enr.WithEntry(MultiaddrENRField, &ogMultiaddrValue)); err != nil {
	if !enr.IsNotFound(err) {
		return err
	}
}

defer func(){
	if e := recover(); e != nil {
                // restoring to original value
		localnode.Set(enr.WithEntry(MultiaddrENRField, ogMultiaddrValue))
		err = errors.New("could not write enr record")
	}
}()
```

richard-ramos:
> @harsh-98: Secondly, i goes from n to 0, shouldn't it go from n to 1. The last iteration will provide 0 len array to writemultiAddressField.

sounds good. I'll fix this

PR #705, opened 2023-09-01
test: circuit relay + peer exchange

richard-ramos:
Test PR just to demonstrate how circuit relay works. To see it in action go to 

```
cd ./examples/circuit
make
make run
```

This test program has the following setup:

1. An unreachable node that will obtain a circuit relay address from a node from the status.prod fleet. This node has discv5
2. A bootnode that has discv5 and peer exchange enabled
3. A peer exchange client.

The unreachable node uses the bootnode for advertising its circuit relay addr stored in the ENR.
The peer exchange client will ask the bootnode via peer exchange for peers every 3 seconds.

After 15-20s you should see some circuit relay multiaddresses obtained via peer exchange

This example is super raw, has a lot of repeated code and should not be merged in master. It's just to demonstrate functionality


waku/v2/peermanager/peer_manager.go
```diff
@@ -231,7 +231,7 @@ func (pm *PeerManager) AddDiscoveredPeer(p PeerData) {
 // addPeer adds peer to only the peerStore.
 // It also sets additional metadata such as origin, ENR and supported protocols
 func (pm *PeerManager) addPeer(ID peer.ID, addrs []ma.Multiaddr, origin wps.Origin, protocols ...protocol.ID) error {
-	pm.logger.Info("adding peer to peerstore", logging.HostID("peer", ID))
+	pm.logger.Debug("adding peer to peerstore", logging.HostID("peer", ID))
```
richard-ramos, line 234: This log line was too noisy with Info

waku/v2/node/wakunode2.go
```diff
@@ -236,6 +236,7 @@ func New(opts ...WakuNodeOption) (*WakuNode, error) {
 			}()
 			return r
 		},
+		autorelay.WithBootDelay(2*time.Second),
```
richard-ramos, line 239: By default, go-libp2p will wait 3 minutes before negotiating circuit relay addresses. See that function documentation to understand the reason behind that!

PR #706, opened 2023-09-04
fix: and optimising fetching membership events

harsh-98:
# Description
Bug and other optimisation found for membership fetching logic while checking https://github.com/waku-org/go-waku/pull/701.

# Changes
- errCh can be closed multiple times. It will result in error.
- Some blocks might be missed btw loadOldEvents and watchNewEvents.
- Use binary split of work for gettting events. 

# Tests

### DIFF
--- a/waku/v2/protocol/rln/group_manager/dynamic/membership_fetcher.go
+++ b/waku/v2/protocol/rln/group_manager/dynamic/membership_fetcher.go
@@ -21,6 +21,7 @@ import (
 // the types of inputs to this handler matches the MemberRegistered event/proc defined in the MembershipContract interface
 type RegistrationEventHandler = func([]*contracts.RLNMemberRegistered) error
 
+// for getting membershipRegsitered Events from the eth rpc
 type MembershipFetcher struct {
 	web3Config  *web3.Config
 	rln         *rln.RLN
@@ -41,46 +42,46 @@ func NewMembershipFetcher(web3Config *web3.Config, rln *rln.RLN, rootTracker *gr
 // HandleGroupUpdates mounts the supplied handler for the registration events emitting from the membership contract
 // It connects to the eth client, subscribes to the `MemberRegistered` event emitted from the `MembershipContract`
 // and collects all the events, for every received event, it calls the `handler`
-func (gm *MembershipFetcher) HandleGroupUpdates(ctx context.Context, handler RegistrationEventHandler) error {
-	fromBlock := gm.web3Config.RLNContract.DeployedBlockNumber
-	metadata, err := gm.GetMetadata()
+func (mf *MembershipFetcher) HandleGroupUpdates(ctx context.Context, handler RegistrationEventHandler) error {
+	fromBlock := mf.web3Config.RLNContract.DeployedBlockNumber
+	metadata, err := mf.GetMetadata()
 	if err != nil {
-		gm.log.Warn("could not load last processed block from metadata. Starting onchain sync from deployment block", zap.Error(err), zap.Uint64("deploymentBlock", gm.web3Config.RLNContract.DeployedBlockNumber))
+		mf.log.Warn("could not load last processed block from metadata. Starting onchain sync from deployment block", zap.Error(err), zap.Uint64("deploymentBlock", mf.web3Config.RLNContract.DeployedBlockNumber))
 	} else {
-		if gm.web3Config.ChainID.Cmp(metadata.ChainID) != 0 {
+		if mf.web3Config.ChainID.Cmp(metadata.ChainID) != 0 {
 			return errors.New("persisted data: chain id mismatch")
 		}
 
-		if !bytes.Equal(gm.web3Config.RegistryContract.Address.Bytes(), metadata.ContractAddress.Bytes()) {
+		if !bytes.Equal(mf.web3Config.RegistryContract.Address.Bytes(), metadata.ContractAddress.Bytes()) {
 			return errors.New("persisted data: contract address mismatch")
 		}
 
 		fromBlock = metadata.LastProcessedBlock + 1
-		gm.log.Info("resuming onchain sync", zap.Uint64("fromBlock", fromBlock))
+		mf.log.Info("resuming onchain sync", zap.Uint64("fromBlock", fromBlock))
 	}
 
-	gm.rootTracker.SetValidRootsPerBlock(metadata.ValidRootsPerBlock)
+	mf.rootTracker.SetValidRootsPerBlock(metadata.ValidRootsPerBlock)
 	//
-	latestBlockNumber, err := gm.latestBlockNumber(ctx)
+	latestBlockNumber, err := mf.latestBlockNumber(ctx)
 	if err != nil {
 		return err
 	}
 	//
-	err = gm.loadOldEvents(ctx, fromBlock, latestBlockNumber, handler)
+	err = mf.loadOldEvents(ctx, fromBlock, latestBlockNumber, handler)
 	if err != nil {
 		return err
 	}
 
 	errCh := make(chan error)
 
-	gm.wg.Add(1)
-	go gm.watchNewEvents(ctx, latestBlockNumber+1, handler, errCh) // we have already fetched the events for latestBlocNumber in oldEvents
+	mf.wg.Add(1)
+	go mf.watchNewEvents(ctx, latestBlockNumber+1, handler, errCh) // we have already fetched the events for latestBlocNumber in oldEvents
 	return <-errCh
 }
 
-func (gm *MembershipFetcher) loadOldEvents(ctx context.Context, fromBlock, toBlock uint64, handler RegistrationEventHandler) error {
+func (mf *MembershipFetcher) loadOldEvents(ctx context.Context, fromBlock, toBlock uint64, handler RegistrationEventHandler) error {
 	for ; fromBlock+maxBatchSize < toBlock; fromBlock += maxBatchSize + 1 { // check if the end of the batch is within the toBlock range
-		events, err := gm.getEvents(ctx, fromBlock, fromBlock+maxBatchSize)
+		events, err := mf.getEvents(ctx, fromBlock, fromBlock+maxBatchSize)
 		if err != nil {
 			return err
 		}
@@ -90,7 +91,7 @@ func (gm *MembershipFetcher) loadOldEvents(ctx context.Context, fromBlock, toBlo
 	}
 
 	//
-	events, err := gm.getEvents(ctx, fromBlock, toBlock)
+	events, err := mf.getEvents(ctx, fromBlock, toBlock)
 	if err != nil {
 		return err
 	}
@@ -98,19 +99,19 @@ func (gm *MembershipFetcher) loadOldEvents(ctx context.Context, fromBlock, toBlo
 	return handler(events)
 }
 
-func (gm *MembershipFetcher) watchNewEvents(ctx context.Context, fromBlock uint64, handler RegistrationEventHandler, errCh chan<- error) {
-	defer gm.wg.Done()
+func (mf *MembershipFetcher) watchNewEvents(ctx context.Context, fromBlock uint64, handler RegistrationEventHandler, errCh chan<- error) {
+	defer mf.wg.Done()
 
 	// Watch for new events
 	firstErr := true
 	headerCh := make(chan *types.Header)
 	subs := event.Resubscribe(2*time.Second, func(ctx context.Context) (event.Subscription, error) {
-		s, err := gm.web3Config.ETHClient.SubscribeNewHead(ctx, headerCh)
+		s, err := mf.web3Config.ETHClient.SubscribeNewHead(ctx, headerCh)
 		if err != nil {
 			if err == rpc.ErrNotificationsUnsupported {
 				err = errors.New("notifications not supported. The node must support websockets")
 			}
-			gm.log.Error("subscribing to rln events", zap.Error(err))
+			mf.log.Error("subscribing to rln events", zap.Error(err))
 		}
 		if firstErr { // errCh can be closed only once
 			errCh <- err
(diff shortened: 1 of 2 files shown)


richard-ramos:
I like the changes related to:
```
- errCh can be closed multiple times. It will result in error.
- Some blocks might be missed btw loadOldEvents and watchNewEvents.
```

However I think the current mechanism for loading events (using chunks) is more appropiate compared to using the proposed binary split, which was one of the initial approaches I had considered for loading the events.

Following this approach, while it works for Infura because it uses an offchain index for event logs, it might overwhelm a node you're running locally. The approach we're currently follow in go-waku is the same as eth2 nodes like prysm in which we fetch block ranges in chunks of 5000 blocks (prysm goes even lower at 1000 blocks per chunk), with some special handling for cases where we exceed the number of events, which is using the `multiplicativeDecreaseDivisor` to reduce the range to a smaller number of blocks, and then slowly increase the range using the `additiveFactorMultiplier`, until we get again back to the max number of blocks (5000).

While I agree that using a binary split might be faster, it causes more load over a local node, and also, syncing the chain in 5K blocks is still a fast operation that takes only some 10s of seconds and it's only executed once at the beginning, and since we already store the merkle tree in the RLN database,

richard-ramos:
Yes but do notice that subsequent requests will still try to load the 5000 requests and split it into separate jobs which on a period of heavy usage of the smart contract will mean that the tooMuchData error should appear frequently for some range periods while trying to load the events, while current approach will not use 5000, but the last valid chunk size + 10% in an attempt to slowly get back to to the original 5000.

Do you think the additive factor + chunk size could be kept with the approach you propose?

richard-ramos:
I see. Not sure either what decision to take, I tend to prefer the original approach but I agree with the argument you present! 
let's ask @chaitanyaprem for feedback!

chaitanyaprem:
> I see. Not sure either what decision to take, I tend to prefer the original approach but I agree with the argument you present! let's ask @chaitanyaprem for feedback!

Had a quick discussion with Harsh regarding both the approaches. 
Wrt the approach of halving only in case of error and reverting back to default blockFetchSize seems simple to me and would make the code simple and more maintainable. 

I would suggest, we go ahead with the approach suggested by @harsh-98.

waku/v2/protocol/rln/group_manager/dynamic/web3.go
```diff
@@ -108,78 +128,43 @@ func (gm *DynamicGroupManager) watchNewEvents(ctx context.Context, rlnContract *
 }
 
 const maxBatchSize = uint64(5000)
-const additiveFactorMultiplier = 0.10
-const multiplicativeDecreaseDivisor = 2
 
 func tooMuchDataRequestedError(err error) bool {
 	// this error is only infura specific (other providers might have different error messages)
 	return err.Error() == "query returned more than 10000 results"
```
chaitanyaprem, line 134: Shouldn't this be an error defined in eth JSON-RPC or am i missing something? 
As of now, it looks like we cannot handle errors from other providers or even if we run a local node.

richard-ramos: this error is infura specific, and it's not standardized. Looking into geth I could not find the error generated when too many records are returned.

waku/v2/protocol/rln/waku_rln_relay.go
```diff
@@ -69,31 +50,39 @@ func New(
...
+}
+func New(
+	gmDetails group_manager.GMDetails,
+	timesource timesource.Timesource,
+	reg prometheus.Registerer,
+	log *zap.Logger) *WakuRLNRelay {
+
+	start := time.Now()
+	metrics := newMetrics(reg)
+	metrics.RecordInstanceCreation(time.Since(start))
```
richard-ramos, line 70: This metric is intended to measure the time it takes to create a `rlnInstance`. Here it is not doing anything useful

waku/v2/protocol/rln/web3/web3.go
```diff
@@ -26,12 +28,21 @@ type RLNContract struct {
...
 
+// EthClientI is an interface for the ethclient.Client, so that we can pass mock client for testing
+type EthClientI interface {
+	bind.ContractBackend
+	TransactionReceipt(ctx context.Context, txHash common.Hash) (*types.Receipt, error)
+	BlockByNumber(ctx context.Context, number *big.Int) (*types.Block, error)
+	SubscribeNewHead(ctx context.Context, ch chan<- *types.Header) (ethereum.Subscription, error)
+	Close()
+}
+
```
richard-ramos, line 39: This naming convention of adding `I` suffix looks too java/csharp-ish. Usually in go we use `-er`. But maybe in this case could it be renamed to `ETHClient` instead?

waku/v2/protocol/rln/group_manager/group_manager.go
```diff
@@ -1 +1,21 @@
...
+
+	"github.com/waku-org/go-zerokit-rln/rln"
+)
+
+type GroupManagerI interface {
+	Start(ctx context.Context) error
+	IdentityCredentials() (rln.IdentityCredential, error)
+	MembershipIndex() rln.MembershipIndex
+	Stop() error
+}
```
richard-ramos, line 14: This can be renamed to `GroupManager`. No need for `I` suffix

waku/v2/protocol/rln/group_manager/dynamic/membership_fetcher_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/waku-org/go-waku/waku/v2/utils"
+	"github.com/waku-org/go-zerokit-rln/rln"
+)
+
+var NULL_ADDR common.Address = common.HexToAddress("0x0000000000000000000000000000000000000000")
+
+func TestFetchingLogic(t *testing.T) {
+	client := NewMockClient(t, "membership_fetcher.json")
+
+	rlnContract, err := contracts.NewRLN(NULL_ADDR, client)
```
richard-ramos, line 22:
```suggestion
func TestFetchingLogic(t *testing.T) {
	client := NewMockClient(t, "membership_fetcher.json")

	rlnContract, err := contracts.NewRLN(common.Address{}, client)
```

waku/v2/protocol/rln/group_manager/dynamic/membership_fetcher_test.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
```
richard-ramos, line 25: Use this to not have to check for nil, and have the testify do it for you. (Same suggestion applies to line 31, 56, etc
```suggestion
	require.NoError(t, err)
```

waku/v2/protocol/rln/group_manager/dynamic/membership_fetcher_test.go
```diff
@@ -0,0 +1,77 @@
...
+	// watchNewEvents will check till 10012
+	if err := dgm.HandleGroupUpdates(context.TODO(), mockFn); err != nil {
+		t.Fatal(err)
+	}
+	// sleep so that watchNewEvents can finish
+	time.Sleep(time.Second)
+	// check whether all the events are fetched or not.
+	if !sameArr(counts, []int{1, 3, 2, 1, 1}) {
+		t.Fatal("wrong no of events fetched per cycle", counts)
+	}
```
richard-ramos, line 64:
```suggestion
	require.Equal(t, counts, []int{1, 3, 2, 1, 1})
```

waku/v2/protocol/rln/onchain_test.go
```diff
@@ -265,11 +273,17 @@ func (s *WakuRLNRelayDynamicSuite) TestCorrectRegistrationOfPeers() {
 	membershipGroupIndex := s.register(appKeystore, credentials1, s.u1PrivKey)
 
 	// mount the rln relay protocol in the on-chain/dynamic mode
-	gm1, err := dynamic.NewDynamicGroupManager(s.web3Config.ETHClientAddress, s.web3Config.RegistryContract.Address, membershipGroupIndex, appKeystore, keystorePassword, prometheus.DefaultRegisterer, utils.Logger())
+	rootIntance, rootTracker, err := GetRLNInstanceAndRootTracker(s.tmpRLNDBPath())
```
richard-ramos, line 276:
```suggestion
	rootInstance, rootTracker, err := GetRLNInstanceAndRootTracker(s.tmpRLNDBPath())
```

waku/v2/protocol/rln/group_manager/group_manager.go
```diff
@@ -1 +1,21 @@
...
+
+type GMDetails struct {
```
richard-ramos, line 16: to avoid stuttering since the package is already named `group_manager`, this can be renamed to `Details`. i.e.:   `group_manager.Details`

richard-ramos (changes requested):
Left some observations. 
Sorry for the rebase issues! but having said that, these will keep happening as the RLN feature is under heavy development. Even tho i'm one who always breaks this rule, Let's try to keep the PRs small!

waku/v2/protocol/rln/group_manager/dynamic/web3.go
```diff
@@ -86,11 +104,13 @@ func (gm *DynamicGroupManager) watchNewEvents(ctx context.Context, rlnContract *
...
 		case h := <-headerCh:
-			blk := h.Number.Uint64()
-			events, err := gm.getEvents(ctx, blk, &blk)
+			toBlock := h.Number.Uint64()
+			events, err := gm.getEvents(ctx, fromBlock, toBlock)
 			if err != nil {
 				gm.log.Error("obtaining rln events", zap.Error(err))
 			}
+			// update the last processed block
+			fromBlock = toBlock + 1
```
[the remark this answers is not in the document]
chaitanyaprem: We can avoid ignoring the block and retry may not be required if this is already part of an outer loop.

waku/v2/protocol/rln/group_manager/dynamic/membership_fetcher.go
```diff
@@ -0,0 +1,226 @@
...
+	"github.com/waku-org/go-waku/waku/v2/protocol/rln/group_manager"
+	"github.com/waku-org/go-waku/waku/v2/protocol/rln/web3"
+	"github.com/waku-org/go-zerokit-rln/rln"
+	"go.uber.org/zap"
+)
+
+// the types of inputs to this handler matches the MemberRegistered event/proc defined in the MembershipContract interface
+type RegistrationEventHandler = func([]*contracts.RLNMemberRegistered) error
+
+type MembershipFetcher struct {
```
chaitanyaprem, line 24: Lets add comments explaining what this is for.

waku/v2/protocol/rln/group_manager/dynamic/membership_fetcher.go
```diff
@@ -0,0 +1,226 @@
...
+		rln:         rln,
+		log:         log,
+		rootTracker: rootTracker,
+	}
+}
+
+// HandleGroupUpdates mounts the supplied handler for the registration events emitting from the membership contract
+// It connects to the eth client, subscribes to the `MemberRegistered` event emitted from the `MembershipContract`
+// and collects all the events, for every received event, it calls the `handler`
+func (gm *MembershipFetcher) HandleGroupUpdates(ctx context.Context, handler RegistrationEventHandler) error {
```
chaitanyaprem, line 44: Minor comment, better to rename gm to mf or something like that to avoid confusion.

waku/v2/protocol/rln/group_manager/dynamic/membership_fetcher.json
```diff
@@ -0,0 +1,100 @@
+{
```
chaitanyaprem, line 1: Rename file to indicate this is for testing.

waku/v2/protocol/rln/group_manager/dynamic/membership_fetcher_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/require"
+	"github.com/waku-org/go-waku/waku/v2/protocol/rln/contracts"
+	"github.com/waku-org/go-waku/waku/v2/protocol/rln/group_manager"
+	"github.com/waku-org/go-waku/waku/v2/protocol/rln/web3"
+	"github.com/waku-org/go-waku/waku/v2/utils"
+	"github.com/waku-org/go-zerokit-rln/rln"
+)
+
+func TestFetchingLogic(t *testing.T) {
```
chaitanyaprem, line 18: Better to add few more tests to validate membership fetch. 
Could be done in separate PR. Following are some tests i can think of: 
- Chain reorg
- Membership deletes
- simulate some failures from mock client.

waku/v2/protocol/rln/group_manager/dynamic/mock_blockchain.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"math/big"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+type MockBlockChain struct {
```
chaitanyaprem, line 12: Add some comments indicating what it is intended for and any assumptions.

chaitanyaprem (approved):
Do address comments. 
Let's avoid such refactor's combining with fixes and other changes, it makes review very hard. 
LGTM

Do wait for ok from @richard-ramos as he is actively working on this.

richard-ramos (approved):

waku/v2/protocol/rln/group_manager/dynamic/web3.go
```diff
@@ -34,7 +34,7 @@ func (gm *DynamicGroupManager) HandleGroupUpdates(ctx context.Context, handler R
 			return errors.New("persisted data: contract address mismatch")
 		}
 
-		fromBlock = metadata.LastProcessedBlock
+		fromBlock = metadata.LastProcessedBlock + 1
```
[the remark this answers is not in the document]
richard-ramos: Yes, it seems okay to me!
Interestingly enough, with the code as it was before, loading the same block twice even if that block has a membership event, did not cause problems because the RLN merkle tree wasn't affected because it will reinsert the same data in the same leaf index.

PR #707, opened 2023-09-04
chore(rln-relay): confirm that the provided credential is correct using onchain query

richard-ramos:
# Description
Updates tests to use zero-based indexing and also checks that the decrypted credential's commitment does in fact belong to the tree by calling memberExists() on the rln storage contract

## Issue
Part of #655


chaitanyaprem (approved):
LGTM

PR #708, opened 2023-09-04
chore(rln-relay): use the only key from keystore if only 1 exists

richard-ramos:
# Description
See https://github.com/waku-org/nwaku/pull/1984 for reasoning behind this change

# Changes

- [x] Created a custom flag type for optional uint values with no default 0 value
- [x] Load a credential automagically if there is only a single credential available on keystore
- [x] Test chat2

## Issue
Part of #655

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,12 +1,12 @@
 plugins:
-  golint:
-    enabled: true
+#  golint:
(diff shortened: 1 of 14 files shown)


chaitanyaprem (approved):
LGTM

PR #710, opened 2023-09-05
chore: build examples as part of CI build process to capture compile errors

richard-ramos:
# Description
To avoid ending up with broken examples, I added a stage to the CI build process to compile them


chaitanyaprem (approved):
Really need this so that we don't miss out updating examples. 
LGTM

PR #712, opened 2023-09-06
chore: add tls/ws to address factory

richard-ramos:
# Description
Since `wss` is deprecated, this PR adds `tls/ws` to the list of multiaddresses returned by the address factory when using dns4+wss. In the future once js-libp2p adds support to `tls/ws`, we can drop support for `wss` entirely

### DIFF
--- a/waku/v2/node/wakuoptions.go
+++ b/waku/v2/node/wakuoptions.go
@@ -190,8 +190,11 @@ func WithDns4Domain(dns4Domain string) WakuNodeOption {
 
 			if params.enableWS || params.enableWSS {
 				if params.enableWSS {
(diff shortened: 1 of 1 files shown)


waku/v2/node/wakuoptions.go
```diff
@@ -190,8 +190,11 @@ func WithDns4Domain(dns4Domain string) WakuNodeOption {
 
 			if params.enableWS || params.enableWSS {
 				if params.enableWSS {
+					// WSS is deprecated
```
chaitanyaprem, line 193: Probably better to give a reference of deprecation.

chaitanyaprem (approved):
LGTM

PR #663, opened 2023-08-21
chore(rln): run rln in all relay pubsubtopics + remove cli flags

richard-ramos:
# Changes
- [x] Remove `--rln-relay-pubsub-topic` and `--rln-relay-content-topic` cli flags.
- [x] Remove all references to pubsub and content topics in rln
- [x] Decouples relay from RLN relay, by creating a `Validator` function that should be called to configure a default validator
- [x] Splits `mountRlnRelay()` into `setupRlnRelay()` and `startRlnRelay()`
- [x] Update test units
- [x] Verify service node and chat2 work as expected

## Issue

part of #655

PR #718, opened 2023-09-07
feat: append RLN proofs when posting messages in REST/RPC

richard-ramos:
# Description
If RLN is enabled, proofs will be appended automatically to the messages sent while using the REST or RPC servers. 

# Changes
- Append RLN Proofs to messages
- Validate topic subscription before posting a message
- Added some comments and fixed code climate issues for some of the files involved in this change

## Issue

part of #655


waku/v2/node/wakunode2_no_rln.go
```diff
@@ -5,6 +5,10 @@ package node
 
 import "context"
 
+// SupportsRLN indicates if go-waku was compiled with RLN support or not
+const SupportsRLN = false
```
chaitanyaprem, line 9: Since supporting RLN is a property of WakuNode, wouldn't it be better to have this as a variable in the WakuNode struct and set the value from these 2 files rather than a const.

cmd/waku/rest/relay.go
```diff
@@ -187,11 +192,23 @@ func (d *RelayService) postV1Message(w http.ResponseWriter, r *http.Request) {
...
-		_, err = d.node.Relay().PublishToTopic(r.Context(), message, strings.Replace(topic, "\n", "", -1))
+		topic = relay.DefaultWakuTopic
 	}
 
+	if !d.node.Relay().IsSubscribed(topic) {
+		writeErrOrResponse(w, errors.New("not subscribed to pubsubTopic"), nil)
+		return
+	}
+
+	if node.SupportsRLN {
```
chaitanyaprem, line 203: I am wondering if we should the appending of proof at a common place, as right now this code is duplicated between rpc and rest. 
Ideal place for this would be to be part of waku_relay, as essentially RLN is enabled only for relay.

cmd/waku/rpc/relay.go
```diff
@@ -78,18 +85,34 @@ func (r *RelayService) Start() {
...
+	if !r.node.Relay().IsSubscribed(topic) {
+		return errors.New("not subscribed to pubsubTopic")
+	}
+
+	msg := args.Message.toProto()
+
+	if node.SupportsRLN {
+		err = r.node.RLNRelay().AppendRLNProof(msg, r.node.Timesource().Now())
+		if err != nil {
+			return err
```
chaitanyaprem, line 111: It would be better to append this error to something like "unable to publish to topic due to RLN error: <errstring>"

waku/v2/node/wakunode2_no_rln.go
richard-ramos: RLN support in this case means if it's compiled with RLN support, but it does not indicate if it is enabled or not (which happens with `--rln-relay` flag, which should be available via `w.opts.rlnEnabled` iirc).  Having it as a constant would mean no possibility of anyone updating this value by mistake during runtime.

cmd/waku/rest/relay.go
richard-ramos: Perhaps. Having it as part of waku_relay tightly couples RLN to Relay which might not be ideal, specially considering that lightpush can be used to send messages as well and these messages must have proofs for them to be considered valid

chaitanyaprem: Got it, is there any other way to avoid this duplication? 
If not, then fine.

waku/v2/node/wakunode2_no_rln.go
chaitanyaprem: Ok, but do we still want to provide that flag? 
I remember @alrevuelta mentioning that RLN should be enabled by default for all pubSub topics.
Does it mean this flag itself will have to be removed?

richard-ramos: if `--rln-relay` is passed, it means that RLN will be applied to all pubsub topics. Before https://github.com/waku-org/go-waku/pull/663, it was required to pass `--rln-relay-pubsub-topic` and `--rln-relay-content-topic` and rln proof verification would be applied exclusively to messages that were sent on that pubsub topic with that content topic.

cmd/waku/rest/relay.go
richard-ramos: Yeah, I'll try to extract it to a function that can be reused in both RPC and REST. Might be an overkill since RPC will die eventually but code will look cleaner anyway

chaitanyaprem (approved):
LGTM

PR #719, opened 2023-09-07
feat(rln-relay): ensure execution order for pubsub validators

richard-ramos:
# Description
This PR guarantees that validators are executed in the order they're appended, that RLN validator is executed as the last validator, and that unmarshalling the WakuMessage is only done once

## Issue

Part of #655

### DIFF
--- a/waku/v2/protocol/relay/validators.go
+++ b/waku/v2/protocol/relay/validators.go
@@ -10,14 +10,14 @@ import (
 
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/crypto/secp256k1"
-
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/core/peer"
+	proto "google.golang.org/protobuf/proto"
+
 	"github.com/waku-org/go-waku/waku/v2/hash"
 	"github.com/waku-org/go-waku/waku/v2/protocol/pb"
 	"github.com/waku-org/go-waku/waku/v2/timesource"
 	"go.uber.org/zap"
-	proto "google.golang.org/protobuf/proto"
 )
 
 func msgHash(pubSubTopic string, msg *pb.WakuMessage) []byte {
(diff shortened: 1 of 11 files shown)


waku/v2/protocol/relay/validators.go
```diff
@@ -38,54 +38,60 @@ func msgHash(pubSubTopic string, msg *pb.WakuMessage) []byte {
...
-		if !withinTimeWindow(t, msg) {
-			return false
+		w.topicValidatorMutex.RLock()
+		validators, exists := w.topicValidators[topic]
+		validators = append(validators, w.defaultTopicValidators...)
+		w.topicValidatorMutex.RUnlock()
+
+		if exists {
+			for _, v := range validators {
+				if !v(ctx, msg, topic) {
```
chaitanyaprem, line 78: I do get that we are trying to optimize by not unmarshalling message twice.
But we are loosing concurrency advantages provided by libp2p's validation mechanism this way. 
Also, as more validators get added this could cause slowing down of message processing as numbers of msgs/sec handled increases.

chaitanyaprem (commented):
What is the reason to ensure execution order of validators? See below my concern regarding this change.
Is it only to optimize so that message unmarshalling happens only once?

waku/v2/protocol/relay/validators.go
richard-ramos: Ah interesting! i did not know pubsub validators were executed concurrently! 
perhaps this happens in nim-libp2p as well? cc: @rymnc

richard-ramos: @rymnc @alrevuelta, look: https://github.com/status-im/nim-libp2p/blob/unstable/libp2p/protocols/pubsub/pubsub.nim#L523-L526  looking at this, it seems to me that all the validators are executed async, and then after all of them are done, then it is check if any of the validators return a rejection.

So, based on the concerns from @chaitanyaprem, does it still make sense to execute the validators linearly as opposed to having all of them execute concurrently? Thinking thoroughly over this issue, it seems to me that maybe we should keep the sequential validation execution to avoid spending processing power for proof verification when the signature verification itself fails... but happy to close this PR otherwise!

alrevuelta: Yep, nim-libp2p validators are executed in parallel and hence the order doesn't matter. I don't think there is an objective best approach. The tradeoff is:
* in parallel: validation times are faster, hence less delay in the network, but resource consumption will be higher under spam, since validator n+1 will be executed even if validator n fails. i would be concerned if validator n+1 is the rln one.
* in sequence: validation times may be slower, but resource consumption may be lower under spam, since once the first validator fails, then the message is rejected.

In nwaku we have opted for the sequence approach, but both approaches are fine imho. I mean, if we reach a point where we have 10 validators, we will need to really optimize this, but both are ok by now.

chaitanyaprem: Ok, in nwaku since it is a service node this approach makes sense since validators are pre-defined.
But in go-waku applications can add their own validators ontop of the internal ones when using it as a library. 

@richard-ramos , maybe for go-waku we should do sequential execution in case of running as a service node and  in library mode to do parallel execution?

richard-ramos: Found an interesting thing while looking at how validators are implemented in go-libp2p, @chaitanyaprem!
It seems that go-libp2p does not allow multiple validators per topic:
https://github.com/libp2p/go-libp2p-pubsub/blob/master/validation.go#L153-L157
Notice if there's already a validator defined for a topic, you'll end up with an error.

Looks like the only way you can have more than one validator is by using default validators which apply to all topics can only be defined before pubsub is instantiated :thinking: .

chaitanyaprem: Hmm, that seems like a major limitation. Odd that multiple default can be added but only one per topic?
The odd thing is that an existing validator gets overridden without any warning.

richard-ramos: Yeah, seems like an odd behavior. I'll merge this PR in the meantime, since for practical purposes, it does make sense to execute the signature validator first before the rln validator, and we can always think of ways to run these concurrently in the future if it ends up being required.

PR #721, opened 2023-09-08
fix: log enr only after it has been setup

richard-ramos:
# Description
The node's ENR was being displayed logged it had its information setup, rendering it useless. This PR changes that so the enr is shown after setupENR is executed


richard-ramos:
Merging this PR as the change is trivial

PR #722, opened 2023-09-09
Fix/filterv2

chaitanyaprem:
# Description
Updated filterv2 proto file as per rfc.
Note that further changes wrt autosharding will be part of separate PR in order not to clutter this PR.

# Changes

- [x] Updated filterv2 proto file as per rfc
- [x] Made broadcaster optional feature for FilterLightNode

# Tests
- Ran all unit tests


waku/v2/protocol/filter/client.go
```diff
@@ -160,15 +161,17 @@ func (wf *WakuFilterLightNode) onRequest(ctx context.Context) func(s network.Str
 			return
 		}
 
-		if !wf.subscriptions.Has(s.Conn().RemotePeer(), messagePush.PubsubTopic, messagePush.WakuMessage.ContentTopic) {
-			logger.Warn("received messagepush with invalid subscription parameters", logging.HostID("peerID", s.Conn().RemotePeer()), zap.String("topic", messagePush.PubsubTopic), zap.String("contentTopic", messagePush.WakuMessage.ContentTopic))
+		if !wf.subscriptions.Has(s.Conn().RemotePeer(), *messagePush.PubsubTopic, messagePush.WakuMessage.ContentTopic) {
```
richard-ramos, line 164: While I do think that message pushes with optional pubsub topic make no sense, this code here is not safe as it is possible for a peer to send a `nil` pubsub topic.

waku/v2/protocol/filter/client.go
```diff
@@ -160,15 +161,17 @@ func (wf *WakuFilterLightNode) onRequest(ctx context.Context) func(s network.Str
...
+			logger.Warn("received messagepush with invalid subscription parameters",
+				logging.HostID("peerID", s.Conn().RemotePeer()), zap.String("topic", *messagePush.PubsubTopic),
```
richard-ramos, line 166: `messagePush.PubsubTopic` can be nil, so this dereference could fail. 
Perhaps we should add something like this in logging/logging.go:
```go

func NullString(key string, value *string) zap.Field {
	v := "<nil>"
	if value != nil {
		v = *value
	}
	return zap.String(key, v)
}
```
and then use instead  `logging.NullString("topic", messagePush.PubsubTopic)`

waku/v2/protocol/filter/client.go
```diff
@@ -160,15 +161,17 @@ func (wf *WakuFilterLightNode) onRequest(ctx context.Context) func(s network.Str
...
+				zap.String("contentTopic", messagePush.WakuMessage.ContentTopic))
 			wf.metrics.RecordError(invalidSubscriptionMessage)
 			return
 		}
 
 		wf.metrics.RecordMessage()
 
-		wf.notify(s.Conn().RemotePeer(), messagePush.PubsubTopic, messagePush.WakuMessage)
+		wf.notify(s.Conn().RemotePeer(), *messagePush.PubsubTopic, messagePush.WakuMessage)
```
richard-ramos, line 174: Can be `nil`

richard-ramos (commented):
Overall it looks okay, but requires some changes due to MessagePush `pubsub_topic` being optional introducing some unsafety around the handling of that attribute.

But I do think in this case the RFC is not correct, as the message pubsub topic is always known by the full node, since messages in Relay always arrive in a pubsubTopic, so IMO this condition in https://rfc.vac.dev/spec/12/#message-push  `If this message was received on a specific pubsub_topic, it SHOULD be included in the MessagePush` should be removed and make the pubsub_topic required. What do you think, @jm-clius?

chaitanyaprem:
> 

Agreed, since Filter service node will always know on which pubSubTopic the message is received.

jm-clius:
Two comments here:
1. The `optional` in the protobuf does not always indicate that the field is optional in the protocol. It may be conditionally mandatory, as it is in this case.
2. Indeed, if the filter service node is a relay node, this field is indeed mandatory as the pubsub topic will be known. However, the spec does not mandate the service node to be a relay node or to participate in a pub/sub configuration. Of course, we don't really envision any other types of service nodes, but we should imagine a service node as "privy to a superset of messages", by whichever means. Imagine a cascade of filter nodes, for example, where each service a more precise filter client. This may be a silly setup, but I don't think we should build a restriction on the service node types into the filter protocol. That would be a bleeding concern.

Also note that under certain autosharding assumptions, we _might_ change the spec in future to exclude pubsub topics as it will be implicit in the content topic.

jm-clius (changes requested):
See my comment elsewhere. I think the `pubsub_topic` field should remain `optional` for `MessagePush` (as it is conditionally mandatory in the protocol)

waku/v2/protocol/filter/client.go
```diff
@@ -159,16 +160,23 @@ func (wf *WakuFilterLightNode) onRequest(ctx context.Context) func(s network.Str
 			wf.metrics.RecordError(decodeRPCFailure)
 			return
 		}
-
-		if !wf.subscriptions.Has(s.Conn().RemotePeer(), messagePush.PubsubTopic, messagePush.WakuMessage.ContentTopic) {
-			logger.Warn("received messagepush with invalid subscription parameters", logging.HostID("peerID", s.Conn().RemotePeer()), zap.String("topic", messagePush.PubsubTopic), zap.String("contentTopic", messagePush.WakuMessage.ContentTopic))
+		//For now returning failure, this will get addressed with autosharding changes for filter.
+		if messagePush.PubsubTopic == nil {
+			logger.Error("empty pubSub Topic", zap.Error(err))
```
richard-ramos, line 165: No err is generated here
```suggestion
			logger.Error("empty pubsub topic")
```

richard-ramos (approved):

Issue #673, opened 2023-08-23
feat: Autosharding API for req-resp protocols


chair28980:
This tracks the work necessary in each client to provide API(s) to applications using req-resp protocols (store, lightpush, filter, etc.) with optional pubsub topic arguments. Each client should then use autosharding to determine the associated shards on behalf of applications. Note that underlying protocols should not be affected and each client implementation should locally populate pubsub topic fields with the shards hashed from content topics received via the API.

Priority: Critical for launch

Tasks

- [x] #723 
- [x] #774

chaitanyaprem:
**Weekly Update**

- _achieved_: Completed Filter API and lightClient changes for autosharding

chaitanyaprem:
**Weekly Update**

- _achieved_: Updated lightpush API for autosharding

PR #723, opened 2023-09-11
Feat/autoshard filter

chaitanyaprem:
# Description
Update Filter client API to work with Autosharding logic. #673 
A client user can specify the following options when subscribing/unsubscribing from filter.
- A pubSubTopic along with a list of contentTopics (that would not be considered for non-autosharding logic)
- No pubSubTopic, but a list of contentTopics (as per specified format) which would be considered for autosharding. In this case if the contentTopics fall under multiple pubSubTopics , separate Filter subscriptions would be created and a subscribe request would be sent for each derived pubSubTopic.

Note: In order to not bloat this PR, optional pubSubTopic handing in MessagePush shall be done in separate PR.

# Changes

- [x] Changes to FilterClient code to compute pubSubTopic if not specified.
- [x] Multiple FilterSubscriptions to be created in case contentTopics provided span across multiple pubSubTopics.
- [x] Added tests to verify filter with Autosharding

# Tests

Added new Filter tests to validate Autosharding.
Same can be executed by running below command in `waku/v2/protocol/filter directory`
`go test -testify.m TestAutoShard`


chaitanyaprem:
@jakubgs , some CI checks are failing due to 500 error.  Tried re-running it, still seeing failure.
Do take a look when you get time.

```

[2023-09-18T00:08:59.955Z] + nix develop --impure .#default --command bash -c 'which xcodebuild'

[2023-09-18T00:09:00.846Z] warning: error: unable to download 'https://nix-cache.status.im/8my2i0lcli53l5jypnzd3cgf7qr3ls06.narinfo': HTTP error 500

[2023-09-18T00:09:00.846Z] 

[2023-09-18T00:09:00.846Z]        response body:

[2023-09-18T00:09:00.846Z] 

[2023-09-18T00:09:00.846Z]        Internal Server Error; retrying in 260 ms

[2023-09-18T00:09:01.268Z] warning: error: unable to download 'https://nix-cache.status.im/8my2i0lcli53l5jypnzd3cgf7qr3ls06.narinfo': HTTP error 500

[2023-09-18T00:09:01.268Z] 

[2023-09-18T00:09:01.268Z]        response body:

[2023-09-18T00:09:01.268Z] 

[2023-09-18T00:09:01.268Z]        Internal Server Error; retrying in 507 ms

[2023-09-18T00:09:02.166Z] warning: error: unable to download 'https://nix-cache.status.im/8my2i0lcli53l5jypnzd3cgf7qr3ls06.narinfo': HTTP error 500

[2023-09-18T00:09:02.166Z] 

[2023-09-18T00:09:02.166Z]        response body:

[2023-09-18T00:09:02.166Z] 

[2023-09-18T00:09:02.166Z]        Internal Server Error; retrying in 1240 ms

[2023-09-18T00:09:03.592Z] warning: error: unable to download 'https://nix-cache.status.im/8my2i0lcli53l5jypnzd3cgf7qr3ls06.narinfo': HTTP error 500

[2023-09-18T00:09:03.592Z] 

[2023-09-18T00:09:03.592Z]        response body:

[2023-09-18T00:09:03.592Z] 

[2023-09-18T00:09:03.592Z]        Internal Server Error; retrying in 2466 ms

[2023-09-18T00:09:06.314Z] error: unable to download 'https://nix-cache.status.im/8my2i0lcli53l5jypnzd3cgf7qr3ls06.narinfo': HTTP error 500

[2023-09-18T00:09:06.314Z] 

[2023-09-18T00:09:06.314Z]        response body:

[2023-09-18T00:09:06.314Z] 

[2023-09-18T00:09:06.314Z]        Internal Server Error

script returned exit code 1
```

jakubgs:
Fixed.

waku/v2/protocol/filter/client.go
```diff
@@ -128,23 +131,27 @@ func (wf *WakuFilterLightNode) onRequest(ctx context.Context) func(s network.Str
 			wf.metrics.RecordError(decodeRPCFailure)
 			return
 		}
+		pubSubTopic := messagePush.PubsubTopic
 		//For now returning failure, this will get addressed with autosharding changes for filter.
 		if messagePush.PubsubTopic == nil {
-			logger.Error("empty pubsub topic")
-			wf.metrics.RecordError(decodeRPCFailure)
-			return
+			*pubSubTopic, err = getPubSubTopicFromContentTopic(messagePush.WakuMessage.ContentTopic)
```
richard-ramos, line 137: I'm confused, but if `messagePush.PubsubTopic` is nil, that means that `pubSubTopic` on line 134 will be nil, right? if so, wouldnt you end up with an error due to dereferencing `*pubsubTopic` which is nil?

waku/v2/protocol/filter/client.go
```diff
@@ -128,23 +131,27 @@ func (wf *WakuFilterLightNode) onRequest(ctx context.Context) func(s network.Str
...
+			if err != nil {
+				logger.Error("could not derive pubSubTopic from contentTopic", zap.Error(err))
```
richard-ramos, line 139: Should we add the contentTopic causing the error to the log?

waku/v2/protocol/filter/client.go
```diff
@@ -212,18 +218,50 @@ func (wf *WakuFilterLightNode) request(ctx context.Context, params *FilterSubscr
 	return nil
 }
 
+func getPubSubTopicFromContentTopic(cTopicString string) (string, error) {
+	cTopic, err := protocol.StringToContentTopic(cTopicString)
+	if err != nil {
+		return "", errors.New(err.Error() + " : " + cTopicString)
```
richard-ramos, line 224: Probably makes sense to use fmt.Errorf here instead of using the concatenation

waku/v2/protocol/filter/client.go
```diff
@@ -212,18 +218,50 @@ func (wf *WakuFilterLightNode) request(ctx context.Context, params *FilterSubscr
...
+	return pTopic.String(), nil
+}
+
+// This function converts a contentFilter into a map of pubSubTopics and corresponding contentTopics
+func contentFilterToPubSubTopicMap(contentFilter ContentFilter) (map[string][]string, error) {
+	pubSubTopicMap := make(map[string][]string)
+
+	if contentFilter.Topic != "" {
+		pubSubTopicMap[contentFilter.Topic] = make([]string, 0)
+		pubSubTopicMap[contentFilter.Topic] = contentFilter.ContentTopics
```
richard-ramos, line 237: Is line 236 necessary? it's being overwritten in 237 anyways

waku/v2/protocol/filter/client.go
```diff
@@ -212,18 +218,50 @@ func (wf *WakuFilterLightNode) request(ctx context.Context, params *FilterSubscr
...
+	} else {
+		//Parse the content-Topics to figure out shards.
+		for _, cTopicString := range contentFilter.ContentTopics {
+			pTopicStr, err := getPubSubTopicFromContentTopic(cTopicString)
+			if err != nil {
+				return nil, err
+			}
+			_, ok := pubSubTopicMap[pTopicStr]
+			if !ok {
+				pubSubTopicMap[pTopicStr] = make([]string, 1)
```
richard-ramos, line 247: Why a length of 1 here? Wouldnt you end up with an empty string at index 0?

waku/v2/protocol/filter/client.go
```diff
@@ -248,11 +286,24 @@ func (wf *WakuFilterLightNode) Subscribe(ctx context.Context, contentFilter Cont
...
 		return nil, err
 	}
-	return wf.subscriptions.NewSubscription(params.selectedPeer, contentFilter.Topic, contentFilter.ContentTopics), nil
+	subscriptions := make([]*SubscriptionDetails, 0)
+	for pubSubTopic, cTopics := range pubSubTopicMap {
+		var cFilter ContentFilter
+		cFilter.Topic = pubSubTopic
+		cFilter.ContentTopics = cTopics
+		//TO OPTIMIZE: Should we parallelize these, if so till how many batches?
+		err := wf.request(ctx, params, pb.FilterSubscribeRequest_SUBSCRIBE, cFilter)
```
richard-ramos, line 299: What happens if one of the requests in this loop fail and previous iterations were succesful?

richard-ramos (commented):
Looks good, however, where pubsub topic is marked as Optional in the different READMEs and comments, I think we need to add add some note indicating that pubsub topics are considered mandatory when used with named and static sharding, and not when using autosharding (in the case of the READMEs), and expand more on what does it mean for the pubsub topic to be optional?

Makes me wonder if we should convert `ContentFilter` into an interface and implementations both for named/static sharding and autosharding, with the former requiring the PubsubTopic? i.e.:

```go
// Names are just an example

contentFilterA := NewAutoshardingContentFilter("theContentTopic)
subs, err := node.FilterLightClient().Subscribe(ctx, contentFilterA)

contentFilterB := NewContentFilter("thePubsubTopic", []string{"theContentTopics"...}
subs, err := node.FilterLightClient().Subscribe(ctx, contentFilterB)
```

The idea would be to allow the pubsub topic to be optional, but make it harder for the library user to use the pubsub topic incorrectly.


I left also some additional comments and questions

chaitanyaprem:
> Looks good, however, where pubsub topic is marked as Optional in the different READMEs and comments, I think we need to add add some note indicating that pubsub topics are considered mandatory when used with named and static sharding, and not when using autosharding (in the case of the READMEs), and expand more on what does it mean for the pubsub topic to be optional?
> 
> Makes me wonder if we should convert `ContentFilter` into an interface and implementations both for named/static sharding and autosharding, with the former requiring the PubsubTopic? i.e.:
> 
> ```go
> // Names are just an example
> 
> contentFilterA := NewAutoshardingContentFilter("theContentTopic)
> subs, err := node.FilterLightClient().Subscribe(ctx, contentFilterA)
> 
> contentFilterB := NewContentFilter("thePubsubTopic", []string{"theContentTopics"...}
> subs, err := node.FilterLightClient().Subscribe(ctx, contentFilterB)
> ```
> 
> The idea would be to allow the pubsub topic to be optional, but make it harder for the library user to use the pubsub topic incorrectly.
> 
> I left also some additional comments and questions

Will update Readme's. 

I do like the idea of making contentFilter in a way so that people cannot misuse.  Will do that.

waku/v2/protocol/filter/client.go
chaitanyaprem: Yeah, took lazy approach. Fixing it by making pubSubTopic a string var.

waku/v2/protocol/filter/client.go
chaitanyaprem: That is alreadu included in the errorString returned from getPubSubTopicFromContentTopic. Hence not needed to add here.

waku/v2/protocol/filter/client.go
chaitanyaprem: Good point, now that i am relooking at it...maybe we can collect the errors and return them to user indicating which contentTopics failed?

richard-ramos (approved):

richard-ramos:
> WDYT?

Sounds good :rocket:

chaitanyaprem:
> > WDYT?
> 
> Sounds good 🚀

Will take care of this in a separate PR, as this PR has been open for too long and has too many merges.

PR #724, opened 2023-09-11
feat: add common protocol design

harsh-98:
# Description
All protocols (lightpush, store, filter, filter_legacy etc ) share same start and stop logic. 

# Changes

# Tests

### DIFF
--- a/waku/v2/protocol/common_service.go
+++ b/waku/v2/protocol/common_service.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 )
 
+// this is common layout for all the services that require mutex protection and a guarantee that all running goroutines will be finished before stop finishes execution. This guarantee comes from waitGroup all one has to use CommonService.WaitGroup() in the goroutines that should finish by the end of stop function.
 type CommonService struct {
 	sync.RWMutex
 	cancel  context.CancelFunc
@@ -56,6 +57,7 @@ func (sp *CommonService) Stop(fn func()) {
 	sp.started = false
 }
 
+// This is not a mutex protected function, it is up to the caller to use it in a mutex protected context
 func (sp *CommonService) ErrOnNotRunning() error {
 	if !sp.started {
 		return ErrNotStarted


waku/v2/protocol/app_design.go
```diff
@@ -0,0 +1,82 @@
...
+		return
+	}
+	sp.cancel()
+	fn()
+	sp.wg.Wait()
+	sp.started = false
+}
+
+func (sp AppDesign) ErrOnNotRunning() error {
+	if !sp.started {
```
richard-ramos, line 71:
```suggestion
        defer sp.RLockAndUnlock()()
	if !sp.started {
```

waku/v2/protocol/app_design.go
```diff
@@ -0,0 +1,82 @@
...
+		return ErrNotStarted
+	}
+	return nil
+}
+
+func (sp *AppDesign) Context() context.Context {
+	return sp.ctx
```
richard-ramos, line 78:
```suggestion
        defer sp.RLockAndUnlock()()
	return sp.ctx
```

waku/v2/protocol/app_design.go
```diff
@@ -0,0 +1,82 @@
+package protocol
+
+import (
+	"context"
+	"errors"
+	"sync"
+)
+
+type AppDesign struct {
```
richard-ramos, line 9: Maybe instead of `AppDesign`, it should be `Service` or `WakuProtocol` or similar? This is used to control the lifetime of the protocols, so probably should have a name closely associated to its purpose

waku/v2/protocol/app_design.go
```diff
@@ -0,0 +1,82 @@
...
+	mu      *sync.RWMutex
+	cancel  context.CancelFunc
+	ctx     context.Context
+	wg      *sync.WaitGroup
+	started bool
+}
+
+func NewAppDesign() AppDesign {
```
richard-ramos, line 17: This is more of a question, but shouldn't this function return a `*AppDesign`?

waku/v2/protocol/filter/client.go
```diff
@@ -236,11 +203,9 @@ func (wf *WakuFilterLightNode) request(ctx context.Context, params *FilterSubscr
 
 // Subscribe setups a subscription to receive messages that match a specific content filter
 func (wf *WakuFilterLightNode) Subscribe(ctx context.Context, contentFilter ContentFilter, opts ...FilterSubscribeOption) (*SubscriptionDetails, error) {
-	wf.RLock()
-	defer wf.RUnlock()
-
-	if !wf.started {
-		return nil, errNotStarted
+	defer wf.RLockAndUnlock()()
```
richard-ramos, line 206: This is probably subjective, but I find `wf.RLockAndUnlock()()` less readable than using `wf.RLock()` followed by `defer wf.RUnlock()`, which is a much more common pattern

richard-ramos (approved):
Very cool PR! :rocket:

waku/v2/protocol/app_design.go
chaitanyaprem: Service may cause confusion with Service-Protocol terminology in Waku which doesn't include relay. 
Maybe we can just call it Protocol or WakuProtocol.

waku/v2/protocol/app_design.go
```diff
@@ -0,0 +1,82 @@
...
+	sp.mu.RLock()
+	return func() {
+		sp.mu.RUnlock()
+	}
+}
+
+// mutex protected start function
+// creates internal context over provided context and runs fn safely
+// fn is excerpt to be executed to start the protocol
+func (sp *AppDesign) Start(ctx context.Context, fn func() error) error {
```
chaitanyaprem, line 40: Let us add some unit tests to validate simultaneous start/stop, invoking stop without calling start etc.
Calling start and stop in a loop etc.

waku/v2/protocol/filter/client.go
chaitanyaprem: The function is actually not unlocking the mutex rather returning another function which can be used to unlock.
I agree with @richard-ramos , Let us just use Rlock and RUnlock. Would make reading code easier.

waku/v2/protocol/filter/server.go
```diff
@@ -317,15 +311,8 @@ func (wf *WakuFilterFullNode) pushMessage(ctx context.Context, peerID peer.ID, e
...
-	}
-
-	wf.h.RemoveStreamHandler(FilterSubscribeID_v20beta1)
-
-	wf.cancel()
-
-	wf.msgSub.Unsubscribe()
-
-	wf.wg.Wait()
+	wf.AppDesign.Stop(func() {
```
chaitanyaprem, line 314: I am finding this pattern hard to go through in the code. 

Inside Stop function, we are calling Stop for AppDesign to which we are passing another set of instructions to stop the original component (e.g: WakuFilterFullNode).  It is somehow looking convoluted.

Wondering if there is a simpler way of doing this which makes code easy to read.
One simple way could be to have a `stop` function which contains all of these actual stop instructions.

chaitanyaprem:
We should also change waku_rln_relay to use this pattern. 
cc @richard-ramos

waku/v2/protocol/filter/server.go
chaitanyaprem: I understood the importance of having this CommonService. Let me include a suggestion to simplify how the usage makes the code look.

waku/v2/protocol/common_service.go
```diff
@@ -0,0 +1,71 @@
+package protocol
+
+import (
+	"context"
+	"errors"
+	"sync"
+)
+
+type CommonService struct {
```
chaitanyaprem, line 9: Let's add a few comments to define scope of this.

waku/v2/protocol/legacy_filter/waku_filter.go
```diff
@@ -327,19 +321,13 @@ func (wf *WakuFilter) Unsubscribe(ctx context.Context, contentFilter ContentFilt
...
-	}
-
-	wf.cancel()
-
-	wf.msgSub.Unsubscribe()
-
-	wf.h.RemoveStreamHandler(FilterID_v20beta1)
-	wf.filters.RemoveAll()
-	wf.subscribers.Clear()
+	wf.CommonService.Stop(func() {
```
chaitanyaprem, line 324: Let us use the stop function in all places to keep it uniform.

chaitanyaprem: Got it, that is also fine.

chaitanyaprem (approved):
LGTM

PR #726, opened 2023-09-11
feat(rln-relay): isReady

richard-ramos:
# Description
Adds a set of functions to verify the RLN state, as well as a `/health` endpoint to the REST server

## Issue
Part of #655


cmd/waku/server/rest/health.go
```diff
@@ -0,0 +1,48 @@
...
+
+	m.Get(routeHealth, h.getHealth)
+
+	return h
+}
+
+type HealthResponse string
+
+func (d *HealthService) getHealth(w http.ResponseWriter, r *http.Request) {
+	isReady, err := d.node.RLNRelay().IsReady(r.Context())
```
chaitanyaprem, line 33: Is just checking health of relay sufficient?
If services are enabled, we should also include those in this healthcheck endpoint.
If someone builds a TUI/GUI later for serviceNode then this endpoint would be useful in fetching and displaying health status of each service.

chaitanyaprem, line 33: Also, shouldn't relay health include ensuring min peering-degree is met?

cmd/waku/server/rest/health.go
```diff
@@ -0,0 +1,48 @@
...
+		if errors.Is(err, context.DeadlineExceeded) {
+			writeResponse(w, HealthResponse("Health check timed out"), http.StatusInternalServerError)
+		} else {
+			writeResponse(w, HealthResponse(err.Error()), http.StatusInternalServerError)
+		}
+		return
+	}
+
+	if isReady {
+		writeResponse(w, HealthResponse("Node is healthy"), http.StatusOK)
```
chaitanyaprem, line 44: Now i see this, i am doubting if this should be health or readiness check? Maybe better to have 2 different endpoints.

I like the pattern [k8s](https://medium.com/@AADota/kubernetes-liveness-and-readiness-probes-difference-1b659c369e17) follows for services . It has liveness and readiness checks which are separate.

cmd/waku/server/rest/health_api.yaml
```diff
@@ -0,0 +1,41 @@
+openapi: 3.0.3
```
chaitanyaprem, line 1: We should maintain these yml files for REST API's also in a common repo like `proto` files rather than each impl having a copy of them.
cc @NagyZoltanPeter  @jm-clius

waku/v2/protocol/rln/group_manager/dynamic/dynamic.go
```diff
@@ -223,6 +227,8 @@ func (gm *DynamicGroupManager) InsertMembers(toInsert *om.OrderedMap) error {
 		}
 		gm.metrics.RecordMembershipInsertionDuration(time.Since(start))
 
+		gm.metrics.RecordRegisteredMembership(startIndex + uint(len(idCommitments)))
+
 		_, err = gm.rootTracker.UpdateLatestRoot(pair.Key.(uint64))
 		if err != nil {
 			return err
```
chaitanyaprem, line 234: Should this become a critical error that we are not able to updateLatestRoot? 
Eventhough a new member is inserted, if the root is not updated then until another event occurs, the new member will not be valid(which would be an issue) in case no events are noticed for a long time.

Should we need to reload rln or retry in this case?

waku/v2/protocol/rln/group_manager/dynamic/dynamic.go
```diff
@@ -274,3 +280,25 @@ func (gm *DynamicGroupManager) Stop() error {
 
 	return nil
 }
+
+func (mf *DynamicGroupManager) IsReady(ctx context.Context) (bool, error) {
```
chaitanyaprem, line 284: mf -> gm

waku/v2/protocol/rln/group_manager/dynamic/metrics.go
```diff
@@ -60,10 +60,10 @@ func (m *metricsImpl) RecordMembershipCredentialsImportDuration(duration time.Du
 }
 
 // RecordRegisteredMembership records the number of registered memberships
-func (m *metricsImpl) RecordRegisteredMembership(num int) {
+func (m *metricsImpl) RecordRegisteredMembership(num uint) {
 	if num < 0 {
```
chaitanyaprem, line 64: This check can be removed now as param is changed to uint.

cmd/waku/server/rest/health.go
richard-ramos: For now this is aligned with the behavior from nwaku

waku/v2/protocol/rln/group_manager/dynamic/dynamic.go
richard-ramos: Tracking work for this here - https://github.com/waku-org/go-waku/issues/738
I think we should panic, because the merkle tree will be inconsistent

cmd/waku/server/rest/health.go
chaitanyaprem: Maybe we can open an enhancement for this task then. Would be better to have something like this.

cmd/waku/server/rest/health_api.yaml
richard-ramos: Tracking work for this in here: https://github.com/waku-org/go-waku/issues/739

cmd/waku/server/rest/health.go
richard-ramos: Tracking work for this in https://github.com/waku-org/go-waku/issues/740

richard-ramos: Tracking work for this in https://github.com/waku-org/go-waku/issues/740

PR #730, opened 2023-09-12
feat: make RLN available by default

richard-ramos:
# Description
This includes RLN support by default, instead of it being optional with the usage of `RLN` environment variable.
Since status-go is not using RLN, the following tag was defined: `gowaku_no_rln` which if defined, will end up not including the RLN dependency, and this way not increasing the size of the mobile and desktop applications by including the `go-zerokit-rln` package as part of the dependencies

## Issue

part of #655


chaitanyaprem (approved):
LGTM

Issue #727, opened 2023-09-12
feat: discovery & peer management for static shards


richard-ramos:
### Problem
Work is in progress in status-go to integrate go-waku's rendezvous, but it will run into a blocking issue: we noticed there is even tho you can advertise the pubsubtopic/shards you are subscribed to, and ask for peers from those shards,  you cannot later retrieve and connect to those peers as information about the pubsub topics supported by the nodes is not stored.

Something else, mentioned by @chaitanyaprem during today's go-waku PM call is that we need to ensure there's at least D peers connected per shard.

Note: This issue tracks the basic Peer-Management updates related to sharding (static/auto). More advanced features like peer-scoring etc would be handled as part of separate issue. This is to reduce scope and iterate.

Tasks

- [x] #734 
- [x] #751 
- [x] #764 
- [x] Add tests to validate specific functionality

Service Peer Management for static shards is being handled by #769

chaitanyaprem:
@richard-ramos , I started working on storing pubSubTopic(s) a peer is participating in as part of peerMetaData. But then i had noticed that go-libp2p-pubSub is having this information wrt peers which are connected. We already seem to be using it with rendezvous https://github.com/waku-org/go-waku/blob/4b1c188cf02ca65493f707946b8b6f131cc2f49d/cmd/waku/node.go#L365

Do you still see the need to store this info additionally as part of peerStore? One limitation i see with using pubSub's information is, i think only the actively connected peers are maintained and peers to whom connections have been dropped are not maintained.

richard-ramos:
Hm. I see. Good point. Perhaps the focus for this task should be then on having the peer connector keep track of the number of peers connected on the pubsub topics that the waku node is currently subscribed to so we can guarantee there's at least D peers connected for each pubsub topic  and once that number is achieved not connect to any additional peer? 

I do find the idea of adding the pubsub topic(s) of a discovered peer in here: https://github.com/waku-org/go-waku/blob/master/waku/v2/peermanager/peer_connector.go#L29-L33 so the peer connector can take an informed decision on whether is it worth it to connect to that peer or not.

--- 
@ilmotta, perhaps this code will help you with integrating rendezvous in status-go https://github.com/waku-org/go-waku/blob/4b1c188cf02ca65493f707946b8b6f131cc2f49d/cmd/waku/node.go#L355-L380

chaitanyaprem:
[...]

PR #734, opened 2023-09-14
feat: update WakuPeerStore to store pubSubTopics for a peer

chaitanyaprem:
# Description
Store pubSubTopics that a peer supports. This is done keeping static sharding in mind as per  #727 
For autosharding, this can be reused, whereas API updates would be requried to only accept contentTopics.

Will take up sharded peer-mgmt in a separate PR in order to avoid making this one too big :) 

# Changes

- [x] Update WakuPeerStoreImpl to provide additional methods to set, retrieve pubSubTopics for a peer
- [x] AddPeer shall take an optional list of pubSubTopics. If nothing is passed, ENR is used to derive topics for relaySharding. If ENR is not present/empty for the peer then defaultRelayTopic is set.
- [x] Static Peers are provided with all pubSubTopics passed in the config as well as derived from content-topics(in case of auto-sharding). This means configuration should be changed to make it more fine-grain so that a static-Node can be specified which topics it supports.
- [x] Peers discovered with discv5 , PeerExchange, DNSDiscovery shall have ENR based on which pubSubTopics are derived.
- [x] Provide API's to fetch  peers based on pubSubTopic

# Tests

All unit tests have passed.
All builds have passed.

Note: Did not add new tests for these, as existing tests for discovery protocols seem to cover all flows.


waku/v2/peerstore/waku_peer_store.go
```diff
@@ -139,3 +145,43 @@ func (ps *WakuPeerstoreImpl) Direction(p peer.ID) (network.Direction, error) {
...
+	return ps.peerStore.Put(p, peerPubSubTopics, existingTopics)
+}
+
+// SetPubSubTopics sets pubSubTopics for a peer, it also overrides existing ones that were set previously..
+func (ps *WakuPeerstoreImpl) SetPubSubTopics(p peer.ID, topics []string) error {
+	return ps.peerStore.Put(p, peerPubSubTopics, topics)
+}
+
+// PubSubTopics fetches list of pubSubTopics for a peer
+func (ps *WakuPeerstoreImpl) PubSubTopics(p peer.ID) ([]string, error) {
```
richard-ramos, line 165: Would it make sense to check the shards stored in the ENRs of peers retrieved via discv5/peer exchange, and return these as well? (this question applies to `PeersByPubSubTopic` too)

richard-ramos (approved):
LGTM, however we'll probably need to add code in future PRs to maintain the peerstore, as peers might join/leave the pubsub topic. Looking at go-libp2p-pubsub code, I see there are some PeerJoin / PeerLeave events we can probably subscribe to, and this way determine whether a peer still is part of a pubsub topic or not

chaitanyaprem:
> LGTM, however we'll probably need to add code in future PRs to maintain the peerstore, as peers might join/leave the pubsub topic. Looking at go-libp2p-pubsub code, I see there are some PeerJoin / PeerLeave events we can probably subscribe to, and this way determine whether a peer still is part of a pubsub topic or not

Right, that also needs to be handled. 
Let me add this as a task in the parent issue to keep track of this.

waku/v2/peerstore/waku_peer_store.go
chaitanyaprem: Those are already being translated to pubSubTopics and added when the peer is added to peerStore. 
https://github.com/waku-org/go-waku/blob/6f8933ef7971e83987f53f07b8ff89423ec964b2/waku/v2/peermanager/peer_manager.go#L219

https://github.com/waku-org/go-waku/blob/6f8933ef7971e83987f53f07b8ff89423ec964b2/waku/v2/discv5/discover.go#L492-L498

and 
https://github.com/waku-org/go-waku/blob/6f8933ef7971e83987f53f07b8ff89423ec964b2/waku/v2/protocol/peer_exchange/client.go#L111-L120

Unless i am missing something.

richard-ramos: oh i see. I missed this!

PR #735, opened 2023-09-14
feat(CommonService): add channel and use commonService in discv5

harsh-98:
# Description
Add commonService to discv5.

# Changes

- Add channel to CommonService. This channel is closed after all the running goRoutine in the service have ended. Ensuring no panics.

# Tests

- All test passed.

### DIFF
--- a/waku/v2/discv5/discover.go
+++ b/waku/v2/discv5/discover.go
@@ -16,7 +16,6 @@ import (
 	"github.com/waku-org/go-waku/logging"
 	"github.com/waku-org/go-waku/waku/v2/peermanager"
 	"github.com/waku-org/go-waku/waku/v2/peerstore"
-	"github.com/waku-org/go-waku/waku/v2/protocol"
 	wenr "github.com/waku-org/go-waku/waku/v2/protocol/enr"
 	"github.com/waku-org/go-waku/waku/v2/utils"
 	"go.uber.org/zap"
@@ -46,7 +45,7 @@ type DiscoveryV5 struct {
 
 	log *zap.Logger
 
-	*protocol.CommonService[peermanager.PeerData]
+	*peermanager.CommonDiscoveryService
 }
 
 type discV5Parameters struct {
@@ -128,12 +127,12 @@ func NewDiscoveryV5(priv *ecdsa.PrivateKey, localnode *enode.LocalNode, peerConn
 	}
 
 	return &DiscoveryV5{
-		params:        params,
-		peerConnector: peerConnector,
-		NAT:           NAT,
-		CommonService: protocol.NewCommonService[peermanager.PeerData](),
-		localnode:     localnode,
-		metrics:       newMetrics(reg),
+		params:                 params,
+		peerConnector:          peerConnector,
+		NAT:                    NAT,
+		CommonDiscoveryService: peermanager.NewCommonDiscoveryService(),
+		localnode:              localnode,
+		metrics:                newMetrics(reg),
 		config: discover.Config{
 			PrivateKey: priv,
 			Bootnodes:  params.bootnodes,
@@ -194,7 +193,7 @@ func (d *DiscoveryV5) SetHost(h host.Host) {
 
 // only works if the discovery v5 hasn't been started yet.
 func (d *DiscoveryV5) Start(ctx context.Context) error {
-	return d.CommonService.Start(ctx, d.start)
+	return d.CommonDiscoveryService.Start(ctx, d.start)
 }
 
 func (d *DiscoveryV5) start() error {
@@ -234,7 +233,7 @@ func (d *DiscoveryV5) Stop() {
 			d.log.Info("recovering from panic and quitting")
 		}
 	}()
-	d.CommonService.Stop(func() {
+	d.CommonDiscoveryService.Stop(func() {
 		if d.listener != nil {
 			d.listener.Close()
 			d.listener = nil
--- a/waku/v2/peerstore/waku_peer_store.go
+++ b/waku/v2/peerstore/waku_peer_store.go
@@ -20,6 +20,7 @@ const (
 	PeerExchange
 	DNSDiscovery
 	Rendezvous
+	PeerManager
 )
 
 const peerOrigin = "origin"
--- a/waku/v2/protocol/common_service.go
+++ b/waku/v2/protocol/common_service.go
@@ -7,17 +7,16 @@ import (
 )
 
 // this is common layout for all the services that require mutex protection and a guarantee that all running goroutines will be finished before stop finishes execution. This guarantee comes from waitGroup all one has to use CommonService.WaitGroup() in the goroutines that should finish by the end of stop function.
-type CommonService[T any] struct {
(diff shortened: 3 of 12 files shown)


waku/v2/protocol/common_service.go
```diff
@@ -4,19 +4,21 @@ import (
...
 
 // this is common layout for all the services that require mutex protection and a guarantee that all running goroutines will be finished before stop finishes execution. This guarantee comes from waitGroup all one has to use CommonService.WaitGroup() in the goroutines that should finish by the end of stop function.
-type CommonService struct {
+type CommonService[T any] struct {
 	sync.RWMutex
 	cancel  context.CancelFunc
 	ctx     context.Context
 	wg      sync.WaitGroup
-	started bool
+	started atomic.Bool
```
chaitanyaprem, line 16: Did not get the need for an atomic here since this var is already protected by a mutex.

waku/v2/protocol/common_service.go
```diff
@@ -45,29 +49,44 @@ var ErrAlreadyStarted = errors.New("already started")
...
 }
-func (sp *CommonService) WaitGroup() *sync.WaitGroup {
+func (sp *CommonService[T]) WaitGroup() *sync.WaitGroup {
 	return &sp.wg
 }
+func (sp *CommonService[T]) GetListeningChan() <-chan T {
+	return sp.channel
+}
+func (sp *CommonService[T]) PushToChan(data T) bool {
+	if !sp.started.Load() {
```
chaitanyaprem, line 83: This check would not be sufficient. 
Let's say stop was called and started is still not set to false, then this check will go through and stop will also continue. This may cause race condition. You will have to use the readMutex to protect this.

waku/v2/discv5/discover.go
```diff
@@ -34,23 +33,20 @@ type PeerConnector interface {
...
 
 	peerConnector PeerConnector
 	NAT           nat.Interface
 
 	log *zap.Logger
 
-	started atomic.Bool
-	cancel  context.CancelFunc
-	wg      *sync.WaitGroup
+	*protocol.CommonService[peermanager.PeerData]
```
chaitanyaprem, line 49: Rather than updating the commonService to include this channel and data which is specific to discovery protocols only. 
Wouldn't be better to use composition and reuse this PeerChannel across all discovery services?
Here it feels like we are trying to fit into existing CommonService and it is not looking clean.

waku/v2/discv5/discover.go
```diff
@@ -197,74 +192,24 @@ func (d *DiscoveryV5) SetHost(h host.Host) {
 	d.host = h
 }
 
-type peerChannel struct {
```
chaitanyaprem, line 200: As i had mentioned above, it would be a better approach to extract this outside of discv5 and reuse it across all discovery services and this way we keep the other protocols(relay, filter etc) untouched.

waku/v2/discv5/discover.go
chaitanyaprem: I understood the reasoning behind this, but maybe we can do it in a better way.
Passing an empty struct in start of a service which doesn't need channel is not looking neat.
Also, PeerChannel is a common pattern in all discovery services to communicate with peerConnector...so i was thinking maybe it is better to extract that out of discovery and reuse it. 
One approach could be ,we can use options pattern to indicate to CommonService whether peerChannel is required or not and use/enable it accordingly in CommonService. 
There would be no change for existing other services, only for services which require working with peerChannel can set this option with a value and CommonService can internally handle it. This way code will look cleaner.

chaitanyaprem: I thought of this approach as well, but why do we want to put generic data channel when we know it is peerChannel. Instead of this, i think it might be better to use options pattern and make channel optional, which are used only by services that need it.
How about something like below?

`type CommonService struct {
    ...
    chan PeerChannel
}

CommonService.WithPeerChannel(c Peer.Data){
  ...
}

func (c *CommonService)Start(){
     ... //other starting tasks 
     //channel specific tasks...
}

`

chaitanyaprem: Hmm, make sense. 
Did not like the style of CommonService and then CommonChanService. You could maybe embed generic channel itself with options pattern? 
If not possible, then proceed with 2 service types.

chaitanyaprem: Was just thinking about this and realized that instead of using generics, can't you use an interface type for the channel? 
Then you can use options pattern to achieve this with CommonService itself.

richard-ramos: I like the idea of having separate services (i.e. `CommonService` and `CommonChanService`) with the idea of not affecting existing protocols, since they don't need to use this channel at all. But looking at this PR, something on which I'm not entirely sold is why use generics, if the purpose of this channel is just to pass a `peermanager.PeerData`, which is used only for passing information to the peer connector.

I'd suggest to besides having two separate base objects: `CommonService` and `CommonChanService`, to also rename the latter to `CommonDiscoveryService`, and use peerstore.PeerData directly. There's probably no need for having generics in this as this component will probably be used only by discovery services.

richard-ramos:
>  @richard-ramos how did you fix failing codeClimate due to generic syntax.

I disabled golint ;) https://github.com/waku-org/go-waku/blob/master/.codeclimate.yml#L2-L3

waku/v2/discv5/discover.go
chaitanyaprem: > I'd suggest to besides having two separate base objects: CommonService and CommonChanService, to also rename the latter to CommonDiscoveryService, and use peerstore.PeerData directly. There's probably no need for having generics in this as this component will probably be used only by discovery services.

This is something i am also fine with. But i think @harsh-98 wanted to include PeerConnector service as well. Which is why we cannot use the term CommonDiscoveryService and rather probably can be done using options pattern WithPeerChannel to enable channel.

waku/v2/peerstore/waku_peer_store.go
```diff
@@ -20,6 +20,7 @@ const (
 	PeerExchange
 	DNSDiscovery
 	Rendezvous
+	PeerManager
```
chaitanyaprem, line 23: Is this required?

waku/v2/rendezvous/rendezvous.go
```diff
@@ -56,14 +56,19 @@ func (r *Rendezvous) SetHost(h host.Host) {
...
 }
 
 func (r *Rendezvous) start() error {
-	err := r.db.Start(r.Context())
-	if err != nil {
-		return err
+	if r.db != nil {
+		if err := r.db.Start(r.Context()); err != nil {
+			return err
+		}
```
chaitanyaprem, line 66: Maybe we should fail the start in case db or peerConnector is not set.

chaitanyaprem (approved):
LGTM

Do address minor comments.

waku/v2/rendezvous/rendezvous.go
chaitanyaprem: Ok, then it is fine.

PR #737, opened 2023-09-14
feat: use CommonService in peerConnector

harsh-98:
# Description
Refactor peerConnector to include CommonService

# Changes

# Tests

### DIFF
--- a/waku/v2/rendezvous/rendezvous.go
+++ b/waku/v2/rendezvous/rendezvous.go
@@ -110,6 +110,7 @@ func (r *Rendezvous) DiscoverWithNamespace(ctx context.Context, namespace string
 				PubSubTopics: []string{namespace},
 			}
 			if !r.PushToChan(peer) {
+				r.log.Error("could push to closed channel/context completed")
 				return
 			}
 		}


chaitanyaprem:
I don't think we need to use generics here.
Look at the discussion in #735 . Better to go ahead with a different approach. 

Generics make code hard to maintain and read. So if there is an alternative approach to generics, i would always prefer that. Looks like @richard-ramos also agrees with this.

chaitanyaprem (approved):
LGTM

But, i would suggest to do some testing by running service-node locally and trying to connect from multiple nodes to ensure nothing is broken wrt peer connectivity and management.

waku/v2/rendezvous/rendezvous.go
```diff
@@ -98,19 +103,14 @@ func (r *Rendezvous) DiscoverWithNamespace(ctx context.Context, namespace string
...
-		r.peerConnector.Subscribe(ctx, peerCh)
 		for _, p := range addrInfo {
 			peer := peermanager.PeerData{
 				Origin:       peerstore.Rendezvous,
 				AddrInfo:     p,
 				PubSubTopics: []string{namespace},
 			}
-			select {
-			case <-ctx.Done():
+			if !r.PushToChan(peer) {
```
richard-ramos, line 112: Might be a good idea to add a warning log here indicating that the push failed

richard-ramos (approved):

PR #742, opened 2023-09-14
feat: allow running rln with no credentials

richard-ramos:
# Description
It's not necessary to have credentials to be a relayer. 

# Changes

- [x] Remove default keystore password and path from config flags
- [x] Instantiate relay before rlnRelay
- [x] Keystore is optional for creating a dynamic group manager

## Issue

Part of #655


chaitanyaprem (approved):
LGTM

But i am wondering if it would be better to provide a flag to enable publishing using RLN and set default params based on that.

PR #747, opened 2023-09-16
fix: nix build with RLN

### LINKED ISSUE #731: bug: nix build fails when using RLN support
In https://github.com/waku-org/go-waku/pull/730 RLN support is included by default in go-waku. However, it seems that nix does not like it because `nix build`  fails with the following error:
```
error: builder for '/nix/store/d9rma0sayfm9pl8042810rqa7mggdh8c-go-waku.drv' failed with exit code 1;
       last 10 log lines:
       > unpacking source archive /nix/store/8pn5ks6ylg54ng9ax9ppfks7ifqpgshw-source
       > source root is source
       > patching sources
       > configuring
       > building
       > Building subPackage ./cmd/waku
       > # github.com/waku-org/go-waku/cmd/waku
       > /nix/store/r0x0agq0vwn0p6z99vkkvn8l8a8idzsb-go-1.19.8/share/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
       > /nix/store/ybw485608d7f1yv1v071j2052q64mvla-binutils-2.40/bin/ld: cannot find -lrln: No such file or directory
       > collect2: error: ld returned 1 exit status
       For full logs, run 'nix-store -l /nix/store/d9rma0sayfm9pl8042810rqa7mggdh8c-go-waku.drv'.
```
It seems that it has trouble finding `librln.a` which is included as part of `go-zerokit-rln-x86_64` and  `go-zerokit-rln-apple` dependencies. This error does not happen when using `make build`

In the meantime I had to add `tags = [ "gowaku_no_rln" ];` to `buildGo119Module` so go-waku compiles.

@jakubgs, any idea what could be happening here?

richard-ramos:
# Description
`buildGo119Module` uses `go mod vendor`. Looking at how this command works, I noticed that it only downloads the files that are inside packages referenced in the code. The linking for RLN is not referenced as a package but via cgo. To solve this, I added some dummy files in go-zerokit-rln [x86_64](https://github.com/waku-org/go-zerokit-rln-x86_64/commit/2a77c3734dd190404c072d6378140045aa7aafe5), arm and apple, and references to this file to force `go mod vendor` to download the static libraries

## Issue

closes #731


flake.nix
```diff
@@ -22,14 +22,14 @@
           name = "go-waku";
           src = self;
           inherit subPackages;
-          tags = [ "gowaku_no_rln" ];
+          tags = [ ];
```
jakubgs, line 25: Might as well drop it.

jakubgs (approved):
Good catch.

PR #750, opened 2023-09-18
feat: topic refactors for filter client

siphiuel:
Changes included:
  - Rename Topic to PubsubTopic to match spec
  - Use set for ContentFilter.ContentTopics
  - client.Subscribe() will return an already existing subscription if we happen to run into the same [peer,contentFilter] pair

### DIFF
--- a/waku/v2/protocol/filter/client.go
+++ b/waku/v2/protocol/filter/client.go
@@ -22,6 +22,7 @@ import (
 	"github.com/waku-org/go-waku/waku/v2/protocol/relay"
 	"github.com/waku-org/go-waku/waku/v2/timesource"
 	"go.uber.org/zap"
+	"golang.org/x/exp/maps"
 )
 
 // FilterPushID_v20beta1 is the current Waku Filter protocol identifier used to allow
@@ -44,8 +45,12 @@ type WakuFilterLightNode struct {
 }
 
 type ContentFilter struct {
-	Topic         string
-	ContentTopics []string
+	PubsubTopic   string
+	ContentTopics ContentTopicSet
+}
+
+func (cf ContentFilter) ContentTopicsList() []string {
+	return maps.Keys(cf.ContentTopics)
 }
 
 type WakuFilterPushResult struct {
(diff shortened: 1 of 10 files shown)


waku/v2/protocol/filter/client.go
```diff
@@ -263,11 +269,11 @@ func (wf *WakuFilterLightNode) FilterSubscription(peerID peer.ID, contentFilter
 		return nil, err
 	}
 
-	if !wf.subscriptions.Has(peerID, contentFilter.Topic, contentFilter.ContentTopics...) {
+	if !wf.subscriptions.Has(peerID, contentFilter.PubsubTopic, maps.Keys(contentFilter.ContentTopics)...) {
```
richard-ramos, line 272: Wondering if we could add a function like this:

```
func (cf ContentFilter) ContentTopics() []string {
   return map.Keys(cf.ContentTopics)
}
```
And call that function directly instead of using `map.Keys` here  and in line 181 and 353?

richard-ramos (approved):
Cool PR :rocket:

richard-ramos:
`jenkins/prs/nix-flake` is failing. This needs to be updated https://github.com/waku-org/go-waku/blob/master/flake.nix#L32

waku/v2/protocol/filter/client.go
```diff
@@ -44,8 +45,8 @@ type WakuFilterLightNode struct {
 }
 
 type ContentFilter struct {
-	Topic         string
-	ContentTopics []string
+	PubsubTopic   string
+	ContentTopics ContentTopicSet
```
chaitanyaprem, line 49: I don't think we should call this a Set as no sorted order is maintained. Wouldn't it be better to use string array itself?

chaitanyaprem (approved):
LGTM

Just a minor comment.

waku/v2/protocol/filter/client.go
siphiuel: @chaitanyaprem thank you for reviewing! With this naming, i wanted to avoid potential duplication issues. The `Set` name is used to signal that 
a) it should only contains unique elements
b) order is not important, as you pointed out.
This als seems to be a common pattern across go-waku code, we often use similar structures, e.g. also `PeerSet` in `subscribers_map.go`, etc.

chaitanyaprem: Got it, also heard your reasoning for keeping it in tandem with status-go data structures.

waku/v2/protocol/filter/client.go
siphiuel: Good idea, thanks, done.

siphiuel:
@richard-ramos fixed Nix build

PR #751, opened 2023-09-18
Feat : handle dynamic peer topic sub unsub

chaitanyaprem:
# Description
Handle dynamic Join and leave of a peer to a pubSubTopic as per #727 .

# Changes

- [x] Expose PeerJoin and Leave events via relay eventBus
- [x] PeerManager to listen for these events and update peer metadata accordingly in peerStore
- [x] Pending go.mod updates in noise example

# Tests
Ran all unit tests with and without race flag.
Validated events by running multiple service nodes locally and connecting with each other.


waku/v2/protocol/relay/waku_relay.go
```diff
@@ -495,3 +527,43 @@ func (w *WakuRelay) Params() pubsub.GossipSubParams {
...
+	w.WaitGroup().Add(1)
+	go w.topicEventPoll(topic.String(), handler)
+	return handler, nil
+}
+
+func (w *WakuRelay) topicEventPoll(topic string, handler *pubsub.TopicEventHandler) {
+	defer w.WaitGroup().Done()
+	for {
+		evt, err := handler.NextPeerEvent(w.Context())
+		if evt.Peer == "" { //Empty peerEvent is returned when context passed in done.
```
richard-ramos, line 545: go-libp2p has the Validate() method which verifies if the peer is nil or not.
```suggestion
		if evt.Peer.Validate() != nil { //Empty peerEvent is returned when context passed in done.
```
Also, this verification should probably go after the `if err != nil` on line 548, since we usually check for errors first before checking for the value returned by a function

richard-ramos (approved):
Cool PR.

PR #753, opened 2023-09-19
feat: force unreachability

richard-ramos:
# Description
Adds a hidden `--force-unreachable` flag that will force the node to use circuit relay. After some seconds the node will have negotiated a circuit relay address if a circuit relay node is added as static node:
```
./build/waku --unreachable --staticnode=/dns4/node-01.do-ams3.go-waku.test.statusim.net/tcp/30303/p2p/16Uiu2HAm9vnvCQgCDrynDK1h7GJoEZVGvnuzq84RyDQ3DEdXmcX7 --wss --wss-key=/path/to/key/file --wss-cert=/path/to/cert/file
```

While testing the functionality with @danisharora099 we noticed that when using an nwaku fleet node, and having wss enabled, the node gets dropped as a peer hence for now I recommend using a node from the gowaku fleet, however the gowaku fleet is returning IP addresses in the list of multiaddresses, so a search/replace must be done to replace the /ip4 address returned by a /dns4 value

This PR also fixes an issue with adding pubsub topics to the peer store that happens when adding a peer for the first time, and changes the TTL for addresses being added to the store since addresses with a ConnectedAddrTTL still had a chance of being removed by identify

### DIFF
--- a/cmd/waku/flags.go
+++ b/cmd/waku/flags.go
@@ -171,6 +171,14 @@ var (
 		Destination: &options.CircuitRelay,
 		EnvVars:     []string{"WAKUNODE2_CIRCUIT_RELAY"},
 	})
+	ForceUnreachable = altsrc.NewBoolFlag(&cli.BoolFlag{
+		Name:        "unreachable",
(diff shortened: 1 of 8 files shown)


chaitanyaprem (approved):
LGTM

PR #757, opened 2023-09-20
ci: use wakuorg Docker Hub organization

jakubgs:
We've been using `statusteam` Docker Hub org for a while now which was never intended for public use, and we should be using `wakuorg` instead.

I've also updated references to `statusteam` in documentation and scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -107,7 +107,7 @@ coverage:
 
 # build a docker image for the fleet
 docker-image: DOCKER_IMAGE_TAG ?= latest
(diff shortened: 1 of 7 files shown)


jakubgs:
I have also pushed prominent Docker images to the new repo:

image

And added deprecation warnings to the old repo:

image

chaitanyaprem (approved):
LGTM

Similar changes are required for nwaku as well. 
cc @vpavlin

richard-ramos (approved):

PR #760, opened 2023-09-21
Postgres tests

harsh-98:
# Description
- Change test that are using sqlite to use postgres. Related to https://github.com/waku-org/go-waku/issues/607

# Changes

- [x] Limit can't be negative for queries in postgres.
- [x] Change store test to use postgres.
- [x] Fixature for testing postgres/sqlite store test. 
- [x] Add postgres module tests.
- [x] Based on driverType, NewQueries from sqlite/postgres should be used in waku_node.  

# Tests

### DIFF
--- a/waku/persistence/postgres/mock.go
+++ b/waku/persistence/postgres/mock.go
@@ -7,8 +7,7 @@ import (
 	"os"
 )
 
-// var dbUrlTemplate = "postgres://postgres@localhost:%s/%s?sslmode=disable"
-var dbUrlTemplate = "postgres://harshjain@localhost:%s/%s?sslmode=disable"
+var dbUrlTemplate = "postgres://postgres@localhost:%s/%s?sslmode=disable"
 
 func ResetDefaultTestPostgresDB(dropDBUrl string) error {
 	db, err := sql.Open("pgx", dropDBUrl)


jakubgs:
Fixed.

waku/persistence/store.go
```diff
@@ -224,7 +224,7 @@ func (d *DBStore) cleanOlderRecords(ctx context.Context) error {
 	// Limit number of records to a max N
 	if d.maxMessages > 0 {
 		start := time.Now()
-		sqlStmt := `DELETE FROM message WHERE id IN (SELECT id FROM message ORDER BY receiverTimestamp DESC LIMIT -1 OFFSET $1)`
+		sqlStmt := `DELETE FROM message WHERE id IN (SELECT id FROM message ORDER BY receiverTimestamp DESC OFFSET $1)`
```
richard-ramos, line 227: This does not work in SQLite. The LIMIT keyword is required when specifying an OFFSET.   If postgresql does not allow negative limits, then we need to extract this query, similar to how it was done with CreateTable in here https://github.com/waku-org/go-waku/blob/16ec22596ef3e4b1a65af1ea1f2a0429847c9f7a/waku/persistence/postgres/postgres.go#L59-L66 which has db specific implementations

richard-ramos (changes requested):

waku/persistence/store.go
richard-ramos: This sounds like a bug!

ci/Jenkinsfile.tests
```diff
@@ -30,6 +30,8 @@ pipeline {
 
   environment {
     TARGET  = 'tests'
+    DB_CONT  = "status-go-test-db-${env.EXECUTOR_NUMBER.toInteger() + 1}"
```
jakubgs, line 33: Why is this calling the container `status-go-test-db-...`? Pretty sure this isn't `status-go` repo.

ci/Jenkinsfile.tests
```diff
@@ -58,6 +60,28 @@ pipeline {
       } } }
     }
 
+    stage('postgres tests') {
```
jakubgs, line 63: The whole Jenkinsfile is for tests, and the job is called `tests`. I don't think yet another `tests` in the stage names. Just call the stage `PostgreSQL`.

waku/persistence/postgres/mock.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"database/sql"
+	"fmt"
+	"log"
+	"os"
+)
+
+// var dbUrlTemplate = "postgres://postgres@localhost:%s/%s?sslmode=disable"
+var dbUrlTemplate = "postgres://harshjain@localhost:%s/%s?sslmode=disable"
```
jakubgs, line 11: Pretty sure you don't want to leave this like this. Also, shouldn't the host and the user be also templateable?

jakubgs (approved):
Mostly looks fine.

richard-ramos (approved):
Very cool. 
if you decide to Squash & Merge the PR, do remember to change the name to match the format (something like `feat: postgresql tests`), or just use rebase & merge since the individual commit messages are fine :)

PR #764, opened 2023-09-22
feat: Sharded peer management - Relay

chaitanyaprem:
# Description
Ensure healthy relay peer connections for pubSub topics or shards subscribed to by the node.
Refer to issue #727 for overall scope.

# Changes

- [x] Handle new topic subscription by identifying and connecting to peers if required
- [x] Handle topic unsubscription and tearing down connections to peers if they don't support any other shards.
- [x] Maintain healty relay peer connections per shard
- [x] Discovered peers are connected immediately until minThreshold peers are reached, post which they are added to peerStore only. Peer connection management is handled by peerManager.

# Tests

- Ran all unit tests
- Ran a serviceNode with single topic and observed topic being identified as unhealthy and moved to healthy by establishing more conns. Connected 10 other nodes to this node to validate connection pruning.


SionoiS (approved):
The logic is good.

In practice it might have problems if there's a lot of node churn.

Until we have something more sophisticated this works.

P.S. can't read Go very well.

waku/v2/peermanager/peer_manager.go
```diff
@@ -31,9 +37,12 @@ type PeerManager struct {
 	serviceSlots        *ServiceSlots
 	ctx                 context.Context
 	sub                 event.Subscription
+	topicMutex          sync.RWMutex
+	subRelayTopics      map[string]*NodeTopicDetails
 }
 
 const peerConnectivityLoopSecs = 15
+const relayOptimalPeersPerShard = 6
```
richard-ramos, line 45: I wonder if it's possible to obtain this value from wakurelay.
It should be possible to obtain `Relay().Params().D`, the problem is how to inject this value to the peer manager.

waku/v2/peermanager/peer_manager.go
```diff
@@ -182,37 +157,45 @@ func (pm *PeerManager) getRelayPeers() (inRelayPeers peer.IDSlice, outRelayPeers
...
+func (pm *PeerManager) connectToRelayPeers() {
 	//Check for out peer connections and connect to more peers.
-	inRelayPeers, outRelayPeers := pm.getRelayPeers()
-	pm.logger.Info("Number of Relay peers connected", zap.Int("inRelayPeers", inRelayPeers.Len()),
+	pm.ensureMinRelayConnsPerTopic()
+
+	inRelayPeers, outRelayPeers := pm.getRelayPeers(nil)
+	pm.logger.Info("Number of Relay peers connected",
+		zap.Int("inRelayPeers", inRelayPeers.Len()),
 		zap.Int("outRelayPeers", outRelayPeers.Len()))
```
richard-ramos, line 194: I think we can remove the `RelayPeers` suffix from here, as the log text already indicates these numbers refer to relay peers
```suggestion
	pm.logger.Info("number of relay peers connected",
		zap.Int("in", inRelayPeers.Len()),
		zap.Int("out", outRelayPeers.Len()))
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -245,6 +234,7 @@ func (pm *PeerManager) pruneInRelayConns(inRelayPeers peer.IDSlice) {
 	//Start disconnecting peers, based on what?
 	//For now, just disconnect most recently connected peers
 	//TODO: Need to have more intelligent way of doing this, maybe peer scores.
+	//TODO: Keep optimalPeersRequired for a pubSubTopic in mind while pruning connections to peers.
 	pm.logger.Info("Number of in peer connections exceed targer relay peers, hence pruning",
 		zap.Int("inRelayPeers", inRelayPeers.Len()), zap.Int("inRelayPeersTarget", pm.InRelayPeersTarget))
```
richard-ramos, line 239:
```suggestion
	pm.logger.Info("peer connections exceed target relay peers, hence pruning",
		zap.Int("cnt", inRelayPeers.Len()), zap.Int("target", pm.InRelayPeersTarget))
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -262,7 +252,13 @@ func (pm *PeerManager) pruneInRelayConns(inRelayPeers peer.IDSlice) {
 // AddDiscoveredPeer to add dynamically discovered peers.
 // Note that these peers will not be set in service-slots.
 // TODO: It maybe good to set in service-slots based on services supported in the ENR
-func (pm *PeerManager) AddDiscoveredPeer(p PeerData) {
+func (pm *PeerManager) AddDiscoveredPeer(p PeerData, connectNow bool) {
+	//Check if the peer is already present, if so skip adding
+	_, err := pm.host.Peerstore().(wps.WakuPeerstore).Origin(p.AddrInfo.ID)
+	if err == nil || err != peerstore.ErrNotFound {
```
richard-ramos, line 258:
```suggestion
	if err == nil || !errors.Is(err, peerstore.ErrNotFound) {
```

I'm confused with this condition. Why does an error other than `ErrNotFound` mean that the discovered peer already exist in the store?

waku/v2/peermanager/topic_event_handler.go
```diff
@@ -0,0 +1,167 @@
...
+	"go.uber.org/zap"
+)
+
+func (pm *PeerManager) SubscribeToRelayEvtBus(bus event.Bus) error {
+	var err error
+	pm.sub, err = bus.Subscribe([]interface{}{new(relay.EvtPeerTopic), new(relay.EvtRelaySubscribed), new(relay.EvtRelayUnsubscribed)})
+	if err != nil {
+		return err
+	}
+	return nil
```
richard-ramos, line 22: It's a nitpick but we can return the error directly
```suggestion
	return err
```

waku/v2/peermanager/topic_event_handler.go
```diff
@@ -0,0 +1,167 @@
...
+				pm.logger.Error("Could not retrieve pubsub topics for peer", zap.Error(err),
+					logging.HostID("peerID", peer))
+				continue
+			}
+			if len(peerTopics) == 1 && peerTopics[0] == pubsubTopic {
+				err := pm.host.Network().ClosePeer(peer)
+				if err != nil {
+					pm.logger.Warn("Failed to disconnect connection towards peer",
+						logging.HostID("peerID", peer))
+				}
```
richard-ramos, line 110: Should `continue` here?
```suggestion
				if err != nil {
					pm.logger.Warn("Failed to disconnect connection towards peer",
						logging.HostID("peerID", peer))
					continue
				}
```

waku/v2/peermanager/topic_event_handler.go
```diff
@@ -0,0 +1,167 @@
...
+					eventDetails := (relay.EvtRelaySubscribed)(e)
+					pm.handleNewRelayTopicSubscription(eventDetails.Topic, eventDetails.TopicInst)
+				}
+			case relay.EvtRelayUnsubscribed:
+				{
+					eventDetails := (relay.EvtRelayUnsubscribed)(e)
+					pm.handleNewRelayTopicUnSubscription(eventDetails.Topic)
+				}
+			default:
+				pm.logger.Error("Received an unsupported event type", zap.Any("eventType", e))
```
richard-ramos, line 160:
```suggestion
				pm.logger.Error("unsupported event type", zap.Any("eventType", e))
```

richard-ramos (commented):
nice PR. Looking fwd to see it in action.
I added some minor comments / nitpick (like reducing verbosity of some logs), but overall it looks great

richard-ramos (approved):

waku/v2/peermanager/peer_manager.go
chaitanyaprem: You are right, I thought i saw in libp2p-peerstore some weird logic of an error being returned other than errNotFound when key is present.
But on rechecking it, i am wrong.

waku/v2/peermanager/peer_manager.go
chaitanyaprem: Yeah, this should be from configuration.
Since we set value of D also from configuration in relay, maybe we can pass the same during PeerManager initialization.

PR #762, opened 2023-09-21
Allow multiple subs for same (peer, contentFilter) pairs

siphiuel:
Following changes are included:
  - allow creating a new Subscription for same (peer, contentFilter) pairs (this was mistakenly assumed to be wrong and thus removed in https://github.com/waku-org/go-waku/pull/750)
  - modify UnsubscribeWithSubscription so that it may close a single subscription only, and not all with a given contentFilter
  - merge FilterSubscribe and FilterUnsubscribe options/params, there was some overlap there
  - resource cleanup: close subscription channels when unsubscribing
  - minor refactors
  - test updates

PR #759, opened 2023-09-20
fix: panic when removing pubsub topic

richard-ramos:
# Description
Fixes this error:
```
panic: runtime error: slice bounds out of range [2:1]

goroutine 291 [running]:
github.com/waku-org/go-waku/waku/v2/peerstore.(*WakuPeerstoreImpl).RemovePubSubTopic(0xc000018b40, {0xc00088a360, 0x27}, {0x0?, 0x11?})
	/home/richard/waku-org/go-waku/waku/v2/peerstore/waku_peer_store.go:169 +0x1dc
github.com/waku-org/go-waku/waku/v2/peermanager.(*PeerManager).peerEventLoop(0xc0003f0c60, {0x26de4d0, 0xc00046e8c0})
	/home/richard/waku-org/go-waku/waku/v2/peermanager/peer_manager.go:109 +0x25c
created by github.com/waku-org/go-waku/waku/v2/peermanager.(*PeerManager).Start
```

PR #766, opened 2023-09-22
Fix deleting pubsub topics; only allow adding unique topics

siphiuel:
Fix issue introduced in https://github.com/waku-org/go-waku/pull/759

Context: https://github.com/waku-org/go-waku/pull/762#issuecomment-1731212779

**EDIT**: i see that it was independently fixed in https://github.com/waku-org/go-waku/pull/765

@chaitanyaprem can you please review this and decide if the uniqueness check is still worth it?

### DIFF
--- a/waku/v2/peerstore/waku_peer_store.go
+++ b/waku/v2/peerstore/waku_peer_store.go
@@ -155,6 +155,11 @@ func (ps *WakuPeerstoreImpl) AddPubSubTopic(p peer.ID, topic string) error {
 	if err != nil {
 		return err
 	}
+	for _, t := range existingTopics {
+		if t == topic {
+			return nil
+		}
+	}
 	existingTopics = append(existingTopics, topic)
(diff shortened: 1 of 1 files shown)


chaitanyaprem (approved):
LGTM

waku/v2/peerstore/waku_peer_store.go
```diff
@@ -155,6 +155,11 @@ func (ps *WakuPeerstoreImpl) AddPubSubTopic(p peer.ID, topic string) error {
 	if err != nil {
 		return err
 	}
+	for _, t := range existingTopics {
+		if t == topic {
+			return errors.New("Pubsub topic already exists")
```
chaitanyaprem, line 160: Now that i am taking a relook, i am wondering if it makes sense to silently return rather than an error. 
That way no harm is done

siphiuel: Ok, will change this.

PR #768, opened 2023-09-22
fix: add ws to enr

richard-ramos:
# Description
Adds ws addresses to the ENR instead of only wss

### DIFF
--- a/waku/v2/node/localnode.go
+++ b/waku/v2/node/localnode.go
@@ -240,17 +240,17 @@ func (w *WakuNode) getENRAddresses(addrs []ma.Multiaddr) (extAddr *net.TCPAddr, 
 		return nil, nil, err
 	}
 
-	multiaddr = append(multiaddr, wssAddrs...)
+	circuitAddrs, err := selectCircuitRelayListenAddresses(addrs)
+	if err != nil {
+		return nil, nil, err
+	}
 
-	// to use WSS, you should have a valid certificate with a domain name.
-	// that means you're reachable. So circuit relay addresses are ignored
-	if len(wssAddrs) == 0 {
-		circuitAddrs, err := selectCircuitRelayListenAddresses(addrs)
-		if err != nil {
-			return nil, nil, err
-		}
-
+	if len(circuitAddrs) != 0 {
+		// Node is unreachable, hence why we have circuit relay multiaddr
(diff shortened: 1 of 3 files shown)


waku/v2/node/localnode.go
```diff
@@ -200,14 +200,9 @@ func selectWSSListenAddresses(addresses []ma.Multiaddr) ([]ma.Multiaddr, error)
...
-		// Only WSS with a domain name are allowed
-		_, err = addr.ValueForProtocol(ma.P_DNS4)
-		if err != nil {
-			continue
-		}
-
-		_, err = addr.ValueForProtocol(ma.P_WSS)
-		if err != nil {
+		_, noWS := addr.ValueForProtocol(ma.P_WSS)
+		_, noWSS := addr.ValueForProtocol(ma.P_WS)
```
chaitanyaprem, line 204: Since we are adding ws now, maybe we should support circuit-relay addresses as well?
There seems to be an outer check like this meant for wss only.
https://github.com/waku-org/go-waku/blob/52353eecba0d2e65d1ef306038fd479f6e2bc65c/waku/v2/node/localnode.go#L247

chaitanyaprem (approved):
LGTM, one comment though.

waku/v2/node/localnode.go
richard-ramos: circuit-relay addresses are added here: https://github.com/waku-org/go-waku/blob/master/waku/v2/node/localnode.go#L223
Probably can be refactored to use a single loop to extract both ws and circuit relay addresses. Can be done in a separate PR if necessary, but since a node listens to only few multiaddresses, the performance improvement of using a single loop vs two might not be worth the effort!

chaitanyaprem: I saw this, but this is done under a check with below comment. But since the return is not just for WSS anymore, if we only have ws addresses, should we include circuit-relay addresses as well. 

```
// to use WSS, you should have a valid certificate with a domain name.
	// that means you're reachable. So circuit relay addresses are ignored 
```
`
 Probably can be refactored to use a single loop to extract both ws and circuit relay addresses. Can be done in a separate PR if necessary, but since a node listens to only few multiaddresses, the performance improvement of using a single loop vs two might not be worth the effort!`

Not worried about this optimization for now

Issue #579, opened 2023-06-23
Refactor : Peer Store 


chaitanyaprem:
### Issue
This issue will be used to track the refactor and various improvements pertaining to peer store. Peer and connection managment related enhancements are being tracked by #594. 

As the objectives of go-waku are different from nwaku and js-waku, changes required would be identified by 
going through the improvements done in nwaku and js-waku along with study on improvements specific to go-waku.

Following are the list of items/considerations for the refactor.

Some high level thoughts for consideration
1. Persisting peers - Implement persistent peer-store, and remove dependency from libp2p peer-store due to [this]( https://github.com/libp2p/go-libp2p/issues/2355).  Establish connections from persisted peers during initialization, only use discover if not enough peers are available.
2. Limit peerStorage capacity similar to - https://github.com/waku-org/nwaku/pull/1513
3. Review Peer related metrics and useful reporting and see if any enhancements are required.
4. Usage specific behaviours (what kind of config to be provided to user layers in order to tune the peer management) Considering service node and light client usage of go-waku - To be analyzed

chaitanyaprem:
As per discussion here, waku has a special case where each time a wakunode is restarted or in status desktop case a user logs out and logs in a new peerID is generated which can bloat up peerStore.

For now a workaround has been implemented which is not clean. 

Taking a look at Prysm, noticed that they don't use libp2p peerStore at all and have their own peerStore defined. 
We should also take this approach as Waku's dynamically changing peerID concept doesn't go well with the way libp2p peerstore is implemented.

chair28980:
Review issue's status in January '24.

chair28980:
Add to Doha agenda for review.

chaitanyaprem:
Refer https://github.com/waku-org/go-waku/issues/852 as well during discussion

chaitanyaprem:
This would no longer be required as nwaku would be replacing go-waku in status by end of 2024.
Hence closing this as there is no plan to implement it.

PR #770, opened 2023-09-25
chore: add config to limit peerstore capacity

chaitanyaprem:
# Description

go-libp2p doesn't provide an option to limit peer-store capacity which may cause it to grow unbounded.
This PR includes a config flag which is already there in nwaku to set peerStore capacity. If not set, set it to 5*maxConnections flag.

Ref https://github.com/waku-org/nwaku/pull/1525 

# Changes

- [ ] New config flag for peer-store-capacity
- [ ] Fail adding peers in case this capacity is breached.

# Tests

Ran all unit tests.
Ran service-node with and without setting peer-store-capacity flag to confirm values are set appropriately.

Did not add any new unit tests as changes seem trivial.

### DIFF
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -69,7 +69,7 @@ func NewPeerManager(maxConnections int, maxPeers int, logger *zap.Logger) *PeerM
 	inRelayPeersTarget, outRelayPeersTarget := inAndOutRelayPeers(maxRelayPeers)
 
 	if maxPeers == 0 || maxConnections > maxPeers {
-		maxPeers = 5 * maxConnections
+		maxPeers = maxConnsToPeerRatio * maxConnections
 	}
 
 	pm := &PeerManager{
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -24,6 +24,7 @@ import (
 	"go.uber.org/zap"
 )
 
+// NodeTopicDetails stores pubSubTopic related data like topicHandle for the node.
 type NodeTopicDetails struct {
 	topic *pubsub.Topic
 }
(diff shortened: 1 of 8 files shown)


richard-ramos (approved):

waku/v2/peermanager/peer_manager.go
```diff
@@ -61,23 +63,29 @@ func inAndOutRelayPeers(relayPeers int) (int, int) {
...
 
 // NewPeerManager creates a new peerManager instance.
-func NewPeerManager(maxConnections int, logger *zap.Logger) *PeerManager {
+func NewPeerManager(maxConnections int, maxPeers int, logger *zap.Logger) *PeerManager {
 
 	maxRelayPeers, _ := relayAndServicePeers(maxConnections)
 	inRelayPeersTarget, outRelayPeersTarget := inAndOutRelayPeers(maxRelayPeers)
 
+	if maxPeers == 0 || maxConnections > maxPeers {
+		maxPeers = 5 * maxConnections
```
richard-ramos, line 72: Maybe the `5` could be made into a constant?

richard-ramos:
Thinking more about this PR, something that i find concerning is that peerID are ephemeral (just logout and login in desktop and you get a new peerID).  Wouldn't this be problematic? I wonder you'd end up filling slots in the peerstore with peers that might no longer exist, and reach a limit in the peerstore

chaitanyaprem:
> Thinking more about this PR, something that i find concerning is that peerID are ephemeral (just logout and login in desktop and you get a new peerID).  Wouldn't this be problematic? I wonder you'd end up filling slots in the peerstore with peers that might no longer exist, and reach a limit in the peerstore

Good point. At some point during this I was thinking if there is a TTL for not connected peers in the peerstore. That should take care of it. If not, another approach we could take is remove peers from the store once they are disconnected, anyways now we have discv5 running constantly doing discovery which would help find peers as and when peers are less.

chaitanyaprem:
For now, I have set address expiry of discovered peers.
When we want to connect to those peers again and addresses are not available, removing them from peerStore.
This should take care of concern you have highlighted.


Looks like we need to handle this in Waku layer in a different way. I looked at Prysm and noticed that they don't use the libp2p peerStore at all and have their own peerStore. Probably we may have to take #579 up to handle such scenarios.

PR #774, opened 2023-09-26
feat: update lightpush API for autosharding

chaitanyaprem:
# Description
Update lightPush API to take pubSub topic optionally and derive from contentTopic using autosharding logic.

# Changes

- [x] Change lightpush API to make pubSubTopic optional
- [x] Autoshard specific light push tests

# Tests

Ran all unit tests

### DIFF
--- a/library/c/api_lightpush.go
+++ b/library/c/api_lightpush.go
@@ -6,7 +6,7 @@ package main
 import "C"
 import "github.com/waku-org/go-waku/library"
 
-// Publish a message using waku lightpush. Use NULL for topic to use the default pubsub topic..
+// Publish a message using waku lightpush. Use NULL for topic to derive the pubsub topic from the contentTopic.
 // peerID should contain the ID of a peer supporting the lightpush protocol. Use NULL to automatically select a node
 // If ms is greater than 0, the broadcast of the message must happen before the timeout
 // (in milliseconds) is reached, or an error will be returned
--- a/waku/v2/protocol/lightpush/waku_lightpush.go
+++ b/waku/v2/protocol/lightpush/waku_lightpush.go
@@ -215,15 +215,22 @@ func (wakuLP *WakuLightPush) Stop() {
 	wakuLP.h.RemoveStreamHandler(LightPushID_v20beta1)
 }
 
-// PublishToTopic is used to broadcast a WakuMessage to a pubsub topic via lightpush protocol
-func (wakuLP *WakuLightPush) PublishToTopic(ctx context.Context, message *wpb.WakuMessage, topic string, opts ...Option) ([]byte, error) {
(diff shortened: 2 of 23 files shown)


library/c/api_lightpush.go
```diff
@@ -6,7 +6,7 @@ package main
 import "C"
 import "github.com/waku-org/go-waku/library"
 
-// Publish a message using waku lightpush. Use NULL for topic to use the default pubsub topic..
+// Publish a message using waku lightpush. Use NULL for topic to derive the pubsub topic from the contentTopic.
```
richard-ramos, line 9: Hm. I have to take this into account for the rust bindings, since it's a breaking change (NULL = DefaultWakuTopic in current version)

waku/v2/protocol/lightpush/waku_lightpush.go
```diff
@@ -215,31 +215,38 @@ func (wakuLP *WakuLightPush) Stop() {
 	wakuLP.h.RemoveStreamHandler(LightPushID_v20beta1)
 }
 
-// PublishToTopic is used to broadcast a WakuMessage to a pubsub topic via lightpush protocol
-func (wakuLP *WakuLightPush) PublishToTopic(ctx context.Context, message *wpb.WakuMessage, topic string, opts ...Option) ([]byte, error) {
+// Optional PublishToTopic is used to broadcast a WakuMessage to a pubsub topic via lightpush protocol
+// If pubSubTopic is not provided, then contentTopic is use to derive the relevant pubSubTopic via autosharding.
+func (wakuLP *WakuLightPush) PublishToTopic(ctx context.Context, message *wpb.WakuMessage, pubsubTopic string, opts ...Option) ([]byte, error) {
```
richard-ramos, line 220: Since pubsubTopic is optional, what do you think of making it an `Option` like... `withPubsubTopic`?

richard-ramos (approved):

library/c/api_lightpush.go
chaitanyaprem: I am wondering if we can have some github action that would open an issue in rust-bindings as and when a change is made in c-bindings so that we won't miss such things.
For now, i will create an issue in rust-bindings for autosharding support that takes care of all such items

chaitanyaprem: Created an issue https://github.com/waku-org/waku-rust-bindings/issues/73

waku/v2/protocol/lightpush/waku_lightpush.go
chaitanyaprem: I was thinking about something like this yesterday when i was working on peer selection logic, but would like to do at some common-Place as this option could be used in many places in protocols and peerManager etc.

chaitanyaprem: Looks like that is not possible, we just have to have this option specified wherever required.

PR #776, opened 2023-09-26
fix(filter2): add requestID to pings and remove unneeded log

richard-ramos:
Fixes this errors that appears on the fleet while doing a ping:
```
    "(kind: MissingRequiredField, field: "request_id")"
    Failed to decode filter subscribe request
```

### DIFF
--- a/waku/v2/protocol/filter/client.go
+++ b/waku/v2/protocol/filter/client.go
@@ -362,7 +362,7 @@ func (wf *WakuFilterLightNode) Ping(ctx context.Context, peerID peer.ID) error {
 
 	return wf.request(
 		ctx,
-		&FilterSubscribeParameters{selectedPeer: peerID},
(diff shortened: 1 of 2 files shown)


siphiuel (approved):
Great :)

chaitanyaprem (approved):
LGTM

Surprising that this did not get caught at protobuf level... probably because it is a string.

richard-ramos:
it' because nwaku's is mixing protobuffer decoding with field validation: https://github.com/waku-org/nwaku/blob/master/waku/waku_filter_v2/rpc_codec.nim#L31-L53 . IMO since the protobuffer request is valid, nwaku should evaluate if a field is required or not just like it is done here: https://github.com/waku-org/nwaku/blob/3264a4f5877203ba085f1a4e8e02e4ffc6ce9249/waku/waku_filter_v2/protocol.nim#L46-L50  and not fail silently

PR #779, opened 2023-09-27
Extract contentFilter and subscriptions out of filter to reuse in relay

chaitanyaprem:
# Description
Since relay with autosharding requires a very similar functionality as filter, extracted some common utils outside in order to reuse it.

# Changes

- [x] Moved contentFilter outside of filter package
- [x] Moved Subscription outside filter package

# Tests

All unit tests have passed.

### DIFF
--- a/waku/v2/protocol/filter/subscribers_map.go
+++ b/waku/v2/protocol/filter/subscribers_map.go
@@ -8,23 +8,14 @@ import (
 
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/waku-org/go-waku/waku/v2/protocol"
 )
 
 var ErrNotFound = errors.New("not found")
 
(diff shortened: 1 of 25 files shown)


waku/v2/protocol/filter/subscribers_map.go
```diff
@@ -106,12 +95,12 @@ func (sub *SubscribersMap) Delete(peerID peer.ID, pubsubTopic string, contentTop
 
 	pubsubTopicMap, ok := sub.items[peerID]
 	if !ok {
-		return ErrNotFound
+		return subscription.ErrNotFound
```
richard-ramos, line 98: I wonder if we should instead create non-exported errors in this file, instead of reusing those from subscription package. Seems somewhat strange to have subscriber functions return subscription errors, but maybe it's too much bikeshedding :)

richard-ramos (approved):
:rocket:

PR #780, opened 2023-09-27
fix: use https for cloning submodules instead of ssh

richard-ramos:
This simplifies the checkout process for github actions and also shortens the path to some files
Required for https://github.com/waku-org/waku-rust-bindings/pull/70

### DIFF
--- a/waku/v2/protocol/rln/contracts/import.json
+++ b/waku/v2/protocol/rln/contracts/import.json
@@ -2,27 +2,27 @@
   "language": "Solidity",
   "sources": {
     "WakuRln.sol": {
-      "urls": ["./waku-rln-contract/contracts/WakuRln.sol"]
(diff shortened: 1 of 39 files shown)


chaitanyaprem (approved):
LGTM

waku/v2/protocol/rln/contracts/import.json
```diff
@@ -2,27 +2,27 @@
   "language": "Solidity",
   "sources": {
     "WakuRln.sol": {
-      "urls": ["./waku-rln-contract/contracts/WakuRln.sol"]
+      "urls": ["../../../../../libs/waku-rln-contract/contracts/WakuRln.sol"]
```
chaitanyaprem, line 5: Would be nice if we could use some env variable instead of all these ..'s

richard-ramos: Tracking this request in https://github.com/waku-org/go-waku/issues/789

Issue #680, opened 2023-08-24
feat: Service peer selection on specific shards


chair28980:
Currently a service peer (e.g. a store node) is either provided by the application or it expects the peer manager to suggest a suitable service peer (selectPeer()). This task focuses on the latter: peer manager should provide ability to find service peers for a specific shard, based on the content topic that the application is interested in (e.g. an API call to selectPeer(Protocol, ContentTopic)), or kick off an ad-hoc discovery process until a suitable service peer is found. Client APIs (for filter, lightpush, peer-exchange) should automatically make use of this mechanism if the application does not explicitly set the desired service peer.

Priority: critical for launch

Tasks

- [x] #783 
- [x] #787 
- [x] #834

chaitanyaprem:
**Weekly Update**

- _achieved_:  PeerSelection updated to be based on pubsubTopic or contentTopic 
- _next_: Update lightClient API to consider new peerSelection options

chaitanyaprem:
**Weekly Update**

- _achieved_:  Refactor and migrate peer selection to peer maanger and update lightclient API to use new options
- _next_: on-demand discovery if peers are not available for the shard

chaitanyaprem:
**Weekly update**
- _achieved_:  draft PR #834 opened for on-demand peer discovery 
- _next_: use on-demand peer discovery for service and relay peer selection

chaitanyaprem:
**Weekly update**
- _achieved_:  on-demand peer discovery for service peer selection and new relay shard subscriptions

PR #783, opened 2023-09-28
Feat: filter select peer for sharding

chaitanyaprem:
# Description
Current selectPeer functionality works without taking sharding into consideration. #680  
This enhances to take pubSubTopic or contentTopic into consideration during peer selection.

# Changes

- [x] Updated selectPeer function to take pubSub topic as argument
- [x] Added a new function to selectPeer based on contentTopic

# Tests

Ran all existing unit tests.
Added few new ones to validate functionality.

### DIFF
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -127,7 +127,7 @@ func (pm *PeerManager) connectivityLoop(ctx context.Context) {
 }
 
 // GroupPeersByDirection returns all the connected peers in peer store grouped by Inbound or outBound direction
-func (pm *PeerManager) GroupPeersByDirection(specificPeers []peer.ID) (inPeers peer.IDSlice, outPeers peer.IDSlice, err error) {
+func (pm *PeerManager) GroupPeersByDirection(specificPeers ...peer.ID) (inPeers peer.IDSlice, outPeers peer.IDSlice, err error) {
 	if len(specificPeers) == 0 {
 		specificPeers = pm.host.Network().Peers()
 	}
@@ -150,9 +150,9 @@ func (pm *PeerManager) GroupPeersByDirection(specificPeers []peer.ID) (inPeers p
 
 // getRelayPeers - Returns list of in and out peers supporting WakuRelayProtocol within specifiedPeers.
 // If specifiedPeers is empty, it checks within all peers in peerStore.
-func (pm *PeerManager) getRelayPeers(specificPeers []peer.ID) (inRelayPeers peer.IDSlice, outRelayPeers peer.IDSlice) {
+func (pm *PeerManager) getRelayPeers(specificPeers ...peer.ID) (inRelayPeers peer.IDSlice, outRelayPeers peer.IDSlice) {
 	//Group peers by their connected direction inbound or outbound.
-	inPeers, outPeers, err := pm.GroupPeersByDirection(specificPeers)
+	inPeers, outPeers, err := pm.GroupPeersByDirection(specificPeers...)
 	if err != nil {
 		return
 	}
@@ -206,7 +206,7 @@ func (pm *PeerManager) connectToRelayPeers() {
 	//Check for out peer connections and connect to more peers.
 	pm.ensureMinRelayConnsPerTopic()
 
-	inRelayPeers, outRelayPeers := pm.getRelayPeers(nil)
+	inRelayPeers, outRelayPeers := pm.getRelayPeers()
 	pm.logger.Info("number of relay peers connected",
 		zap.Int("in", inRelayPeers.Len()),
 		zap.Int("out", outRelayPeers.Len()))
@@ -421,12 +421,12 @@ func (pm *PeerManager) addPeerToServiceSlot(proto protocol.ID, peerID peer.ID) {
 // If a list of specific peers is passed, the peer will be chosen from that list assuming
 // it supports the chosen protocol and contentTopic, otherwise it will chose a peer from the service slot.
 // If a peer cannot be found in the service slot, a peer will be selected from node peerstore
-func (pm *PeerManager) SelectPeerByContentTopic(proto protocol.ID, contentTopic string, specificPeers []peer.ID) (peer.ID, error) {
+func (pm *PeerManager) SelectPeerByContentTopic(proto protocol.ID, contentTopic string, specificPeers ...peer.ID) (peer.ID, error) {
 	pubsubTopic, err := waku_proto.GetPubSubTopicFromContentTopic(contentTopic)
 	if err != nil {
 		return "", err
 	}
-	return pm.SelectPeer(proto, pubsubTopic, specificPeers)
+	return pm.SelectPeer(proto, pubsubTopic, specificPeers...)
 }
 
 // SelectPeer is used to return a random peer that supports a given protocol.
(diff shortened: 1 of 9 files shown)


chaitanyaprem:
@SionoiS ,  I have noticed that in nwaku while selecting peers from serviceSlots the filtering is not done as per shards. 
It has been done only for relay peers.
https://github.com/waku-org/nwaku/blob/8455b8dd1764d7e82d51e6020ac613fb17bfbe6e/waku/node/peer_manager/peer_manager.nim#L653-L656

JFYI in case it is not already being tracked.

SionoiS:
> @SionoiS , I have noticed that in nwaku while selecting peers from serviceSlots the filtering is not done as per shards. It has been done only for relay peers. https://github.com/waku-org/nwaku/blob/8455b8dd1764d7e82d51e6020ac613fb17bfbe6e/waku/node/peer_manager/peer_manager.nim#L653-L656
> 
> JFYI in case it is not already being tracked.

Ah! good catch I didn't notice that. Thanks!

waku/v2/peermanager/peer_manager.go
```diff
@@ -417,29 +417,66 @@ func (pm *PeerManager) addPeerToServiceSlot(proto protocol.ID, peerID peer.ID) {
...
+	return pm.SelectPeer(proto, pubsubTopic, specificPeers)
+}
+
 // SelectPeer is used to return a random peer that supports a given protocol.
 // If a list of specific peers is passed, the peer will be chosen from that list assuming
 // it supports the chosen protocol, otherwise it will chose a peer from the service slot.
 // If a peer cannot be found in the service slot, a peer will be selected from node peerstore
-func (pm *PeerManager) SelectPeer(proto protocol.ID, specificPeers []peer.ID) (peer.ID, error) {
+// if pubSubTopic is specified, peer is selected from list that support the pubSubTopic
+func (pm *PeerManager) SelectPeer(proto protocol.ID, pubSubTopic string, specificPeers []peer.ID) (peer.ID, error) {
```
richard-ramos, line 437: Similar to a suggestion that i did in the changes for lightpush, i think it's a good idea to make the pubsubTopic an Option, and/or i'm also thinking we should make specificPeers `specificPeers ...peer.ID` to avoid having to use a `nil` when we dont need specific peers

waku/v2/peerstore/waku_peer_store.go
```diff
@@ -208,9 +208,13 @@ func (ps *WakuPeerstoreImpl) PubSubTopics(p peer.ID) ([]string, error) {
 }
 
 // PeersByPubSubTopic Returns list of peers by pubSubTopic
-func (ps *WakuPeerstoreImpl) PeersByPubSubTopic(pubSubTopic string) peer.IDSlice {
+// If specifiPeers are listed, filtering is done from them otherwise from all peers in peerstore
+func (ps *WakuPeerstoreImpl) PeersByPubSubTopic(pubSubTopic string, specificPeers peer.IDSlice) peer.IDSlice {
```
richard-ramos, line 212: Let's make this function variadic, so we can skip having to set a `nil` if we dont want specific peers
```suggestion
func (ps *WakuPeerstoreImpl) PeersByPubSubTopic(pubSubTopic string, specificPeers ...peer.ID) peer.IDSlice {
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -417,29 +417,66 @@ func (pm *PeerManager) addPeerToServiceSlot(proto protocol.ID, peerID peer.ID) {
 	pm.serviceSlots.getPeers(proto).add(peerID)
 }
 
+// SelectPeerByContentTopic is used to return a random peer that supports a given protocol for given contentTopic.
+// If a list of specific peers is passed, the peer will be chosen from that list assuming
+// it supports the chosen protocol and contentTopic, otherwise it will chose a peer from the service slot.
+// If a peer cannot be found in the service slot, a peer will be selected from node peerstore
+func (pm *PeerManager) SelectPeerByContentTopic(proto protocol.ID, contentTopic string, specificPeers []peer.ID) (peer.ID, error) {
```
richard-ramos, line 424: Let's make this function variadic :)
```suggestion
func (pm *PeerManager) SelectPeerByContentTopic(proto protocol.ID, contentTopic string, specificPeers ...peer.ID) (peer.ID, error) {
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -417,29 +417,66 @@ func (pm *PeerManager) addPeerToServiceSlot(proto protocol.ID, peerID peer.ID) {
...
 		return "", err
 	}
-
+	if pubSubTopic != "" {
+		filteredPeers = pm.host.Peerstore().(wps.WakuPeerstore).PeersByPubSubTopic(pubSubTopic, filteredPeers)
+	}
 	return utils.SelectRandomPeer(filteredPeers, pm.logger)
 }
+
+func (pm *PeerManager) selectServicePeer(proto protocol.ID, pubSubTopic string, specificPeers []peer.ID) (peerIDPtr *peer.ID) {
```
richard-ramos, line 459:
```suggestion
func (pm *PeerManager) selectServicePeer(proto protocol.ID, pubSubTopic string, specificPeers ...peer.ID) (peerIDPtr *peer.ID) {
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -417,29 +417,66 @@ func (pm *PeerManager) addPeerToServiceSlot(proto protocol.ID, peerID peer.ID) {
...
+	peerIDPtr = nil
+
+	//Try to fetch from serviceSlot
+	if slot := pm.serviceSlots.getPeers(proto); slot != nil {
+		if pubSubTopic == "" {
+			if peerID, err := slot.getRandom(); err == nil {
+				peerIDPtr = &peerID
+			}
```
richard-ramos, line 467:
```suggestion
			if peerID, err := slot.getRandom(); err == nil {
				peerIDPtr = &peerID
			} else {
			        pm.logger.Error("could not retrieve random peer from slot", zap.Error(err))
			}
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -417,29 +417,66 @@ func (pm *PeerManager) addPeerToServiceSlot(proto protocol.ID, peerID peer.ID) {
...
+			var keys peer.IDSlice
+			keys = make([]peer.ID, 0, len(slot.m))
+			for i := range slot.m {
+				keys = append(keys, i)
+			}
+			selectedPeers := pm.host.Peerstore().(wps.WakuPeerstore).PeersByPubSubTopic(pubSubTopic, keys)
+			peerID, err := utils.SelectRandomPeer(selectedPeers, pm.logger)
+			if err == nil {
+				peerIDPtr = &peerID
+			}
```
richard-ramos, line 478:
```suggestion
			if err == nil {
				peerIDPtr = &peerID
			} else {
			        logger.Error("could not select random peer", zap.Error(err))
			}
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -417,29 +417,66 @@ func (pm *PeerManager) addPeerToServiceSlot(proto protocol.ID, peerID peer.ID) {
...
+
+	//Try to fetch from serviceSlot
+	if slot := pm.serviceSlots.getPeers(proto); slot != nil {
+		if pubSubTopic == "" {
+			if peerID, err := slot.getRandom(); err == nil {
+				peerIDPtr = &peerID
+			}
+		} else { //PubsubTopic based selection
+			var keys peer.IDSlice
+			keys = make([]peer.ID, 0, len(slot.m))
```
richard-ramos, line 470: I think you can do this
```suggestion
			keys := make([]peer.ID, 0, len(slot.m))
```

richard-ramos (approved):

waku/v2/peermanager/peer_manager.go
chaitanyaprem: I don't think we should log error here, since if this selection fails, we select randomly from peerStore.

waku/v2/peermanager/peer_manager.go
chaitanyaprem: I don't think we should log error here, since if this selection fails, we select randomly from peerStore.

PR #784, opened 2023-09-28
Chore: test cov improvement

chaitanyaprem:
# Description
By using -coverpkg flag, coverage is reported from tests that are run outside a package.
In most cases since tests run on waku/v2/node cover many areas under protocol etc, coverage percentage is not updated.
Similarly if tests are in a separate folder, then coverage is not updated for tests run.

Also added few additional tests to improve coverage in recently added code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,7 +80,7 @@ test-with-race:
 	${GOBIN} test -race -timeout 300s ./waku/... 
 
 test:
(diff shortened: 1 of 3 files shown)


richard-ramos (approved):

chaitanyaprem:
**Weekly Update**

- _achieved_: Improved test coverage for  wakunode2

PR #785, opened 2023-09-28
refactor(c-bindings): userdata

richard-ramos:
# Description
Adds an `void* user_data` parameter to all c-bindings API functions so it's possible to use them without having to use global variables

# New API:
```c
// Initialize a waku node. Receives a JSON string containing the configuration
// for the node. It can be NULL. Example configuration:
// ```
// {"host": "0.0.0.0", "port": 60000, "advertiseAddr": "1.2.3.4", "nodeKey": "0x123...567", "keepAliveInterval": 20, "relay": true}
// ```
// All keys are optional. If not specified a default value will be set:
// - host: IP address. Default 0.0.0.0
// - port: TCP port to listen. Default 60000. Use 0 for random
// - advertiseAddr: External IP
// - nodeKey: secp256k1 private key. Default random
// - keepAliveInterval: interval in seconds to ping all peers
// - relay: Enable WakuRelay. Default `true`
// - relayTopics: Array of pubsub topics that WakuRelay will automatically subscribe to when the node starts
// - gossipsubParams: an object containing custom gossipsub parameters. All attributes are optional, and if not specified, it will use default values.
//   - d: optimal degree for a GossipSub topic mesh. Default `6`
//   - dLow: lower bound on the number of peers we keep in a GossipSub topic mesh. Default `5`
//   - dHigh: upper bound on the number of peers we keep in a GossipSub topic mesh. Default `12`
//   - dScore: affects how peers are selected when pruning a mesh due to over subscription. Default `4`
//   - dOut: sets the quota for the number of outbound connections to maintain in a topic mesh. Default `2`
//   - historyLength: controls the size of the message cache used for gossip. Default `5`
//   - historyGossip: controls how many cached message ids we will advertise in IHAVE gossip messages. Default `3`
//   - dLazy: affects how many peers we will emit gossip to at each heartbeat. Default `6`
//   - gossipFactor: affects how many peers we will emit gossip to at each heartbeat. Default `0.25`
//   - gossipRetransmission: controls how many times we will allow a peer to request the same message id through IWANT gossip before we start ignoring them. Default `3`
//   - heartbeatInitialDelayMs: short delay in milliseconds before the heartbeat timer begins after the router is initialized. Default `100` milliseconds
//   - heartbeatIntervalSeconds: controls the time between heartbeats. Default `1` second
//   - slowHeartbeatWarning: duration threshold for heartbeat processing before emitting a warning. Default `0.1`
//   - fanoutTTLSeconds: controls how long we keep track of the fanout state. Default `60` seconds
//   - prunePeers: controls the number of peers to include in prune Peer eXchange. Default `16`
//   - pruneBackoffSeconds: controls the backoff time for pruned peers. Default `60` seconds
//   - unsubscribeBackoffSeconds: controls the backoff time to use when unsuscribing from a topic. Default `10` seconds
//   - connectors: number of active connection attempts for peers obtained through PX. Default `8`
//   - maxPendingConnections: maximum number of pending connections for peers attempted through px. Default `128`
//   - connectionTimeoutSeconds: timeout in seconds for connection attempts. Default `30` seconds
//   - directConnectTicks: the number of heartbeat ticks for attempting to reconnect direct peers that are not currently connected. Default `300`
//   - directConnectInitialDelaySeconds: initial delay before opening connections to direct peers. Default `1` second
//   - opportunisticGraftTicks: number of heartbeat ticks for attempting to improve the mesh with opportunistic grafting. Default `60`
//   - opportunisticGraftPeers: the number of peers to opportunistically graft. Default `2`
//   - graftFloodThresholdSeconds: If a GRAFT comes before GraftFloodThresholdSeconds has elapsed since the last PRUNE, then there is an extra score penalty applied to the peer through P7. Default `10` seconds
//   - maxIHaveLength: max number of messages to include in an IHAVE message, also controls the max number of IHAVE ids we will accept and request with IWANT from a peer within a heartbeat. Default `5000`
//   - maxIHaveMessages: max number of IHAVE messages to accept from a peer within a heartbeat. Default `10`
//   - iWantFollowupTimeSeconds: Time to wait for a message requested through IWANT following an IHAVE advertisement. Default `3` seconds
//   - seenMessagesTTLSeconds:  configures when a previously seen message ID can be forgotten about. Default `120` seconds
//
// - minPeersToPublish: The minimum number of peers required on a topic to allow broadcasting a message. Default `0`
// - legacyFilter: Enable LegacyFilter. Default `false`
// - discV5: Enable DiscoveryV5. Default `false`
// - discV5BootstrapNodes: Array of bootstrap nodes ENR
// - discV5UDPPort: UDP port for DiscoveryV5
// - logLevel: Set the log level. Default `INFO`. Allowed values "DEBUG", "INFO", "WARN", "ERROR", "DPANIC", "PANIC", "FATAL"
// - store: Enable Store. Default `false`
// - databaseURL: url connection string. Default: "sqlite3://store.db". Also accepts PostgreSQL connection strings
// - storeRetentionMaxMessages: max number of messages to store in the database. Default 10000
// - storeRetentionTimeSeconds: max number of seconds that a message will be persisted in the database. Default 2592000 (30d)
// - websockets: an optional object containing settings to setup the websocket configuration
//   - enabled: indicates if websockets support will be enabled. Default `false`
//   - host: listening address for websocket connections. Default `0.0.0.0`
//   - port: TCP listening port for websocket connection (0 for random, binding to 443 requires root access). Default: `60001“, if secure websockets support is enabled, the default is `6443“
//   - secure: enable secure websockets support. Default `false`
//   - certPath: secure websocket certificate path
//   - keyPath: secure websocket key path
//
// - dns4DomainName: the domain name resolving to the node's public IPv4 address.
//
extern int waku_new(char* configJSON, WakuCallBack cb, void* userData);

// Starts the waku node
//
extern int waku_start(WakuCallBack onErr, void* userData);

// Stops a waku node
//
extern int waku_stop(WakuCallBack onErr, void* userData);

// Determine is a node is started or not
//
extern int waku_is_started();

// Obtain the peer ID of the waku node
//
extern int waku_peerid(WakuCallBack cb, void* userData);

// Obtain the multiaddresses the wakunode is listening to
//
extern int waku_listen_addresses(WakuCallBack cb, void* userData);

// Add node multiaddress and protocol to the wakunode peerstore
//
extern int waku_add_peer(char* address, char* protocolID, WakuCallBack cb, void* userData);

// Connect to peer at multiaddress. if ms > 0, cancel the function execution if it takes longer than N milliseconds
//
extern int waku_connect(char* address, int ms, WakuCallBack cb, void* userData);

// Connect to known peer by peerID. if ms > 0, cancel the function execution if it takes longer than N milliseconds
//
extern int waku_connect_peerid(char* peerID, int ms, WakuCallBack cb, void* userData);

// Close connection to a known peer by peerID
//
extern int waku_disconnect(char* peerID, WakuCallBack cb, void* userData);

// Get number of connected peers
//
extern int waku_peer_cnt(WakuCallBack cb, void* userData);

// Create a content topic string according to RFC 23
//
extern int waku_content_topic(char* applicationName, unsigned int applicationVersion, char* contentTopicName, char* encoding, WakuCallBack cb, void* userData);

// Create a pubsub topic string according to RFC 23
//
extern int waku_pubsub_topic(char* name, char* encoding, WakuCallBack cb, void* userData);

// Get the default pubsub topic used in waku2: /waku/2/default-waku/proto
//
extern int waku_default_pubsub_topic(WakuCallBack cb, void* userData);

// Register callback to act as signal handler and receive application signals
// (in JSON) which are used to react to asynchronous events in waku. The function
// signature for the callback should be `void myCallback(char* signalJSON)`
//
extern void waku_set_event_callback(WakuCallBack cb);

// Retrieve the list of peers known by the waku node
//
extern int waku_peers(WakuCallBack cb, void* userData);

// Returns a list of objects containing the peerID, enr and multiaddresses for each node found
//
//	given a url to a DNS discoverable ENR tree
//
// The nameserver can optionally be specified to resolve the enrtree url. Otherwise NULL or
// empty to automatically use the default system dns.
// If ms is greater than 0, the subscription must happen before the timeout
// (in milliseconds) is reached, or an error will be returned
//
extern int waku_dns_discovery(char* url, char* nameserver, int ms, WakuCallBack cb, void* userData);

// Update the bootnode list used for discovering new peers via DiscoveryV5
// The bootnodes param should contain a JSON array containing the bootnode ENRs i.e. `["enr:...", "enr:..."]`
//
extern int waku_discv5_update_bootnodes(char* bootnodes, WakuCallBack cb, void* userData);

// Decode a waku message using a 32 bytes symmetric key. The key must be a hex encoded string with "0x" prefix
//
extern int waku_decode_symmetric(char* messageJSON, char* symmetricKey, WakuCallBack cb, void* userData);

// Decode a waku message using a secp256k1 private key. The key must be a hex encoded string with "0x" prefix
//
extern int waku_decode_asymmetric(char* messageJSON, char* privateKey, WakuCallBack cb, void* userData);

// Encrypt a message with a secp256k1 public key.
// publicKey must be a hex string prefixed with "0x" containing a valid secp256k1 public key.
// optionalSigningKey is an optional hex string prefixed with "0x" containing a valid secp256k1 private key for signing the message. Use NULL otherwise
// The message version will be set to 1
//
extern int waku_encode_asymmetric(char* messageJSON, char* publicKey, char* optionalSigningKey, WakuCallBack cb, void* userData);

// Encrypt a message with a 32 bytes symmetric key
// symmetricKey must be a hex string prefixed with "0x" containing a 32 bytes symmetric key
// optionalSigningKey is an optional hex string prefixed with "0x" containing a valid secp256k1 private key for signing the message. Use NULL otherwise
// The message version will be set to 1
//
extern int waku_encode_symmetric(char* messageJSON, char* symmetricKey, char* optionalSigningKey, WakuCallBack cb, void* userData);

// Creates a subscription to a filter full node matching a content filter.
// filterJSON must contain a JSON with this format:
//
//		{
//		  "pubsubTopic": "the pubsub topic" // optional if using autosharding, mandatory if using static or named sharding.
//	      "contentTopics": ["the content topic"] // mandatory, at least one required, with a max of 10
//		}
//
// peerID should contain the ID of a peer supporting the filter protocol. Use NULL to automatically select a node
// If ms is greater than 0, the subscription must happen before the timeout
// (in milliseconds) is reached, or an error will be returned
// It returns a json object containing the details of the subscriptions along with any errors in case of partial failures
//
extern int waku_filter_subscribe(char* filterJSON, char* peerID, int ms, WakuCallBack cb, void* userData);

// Used to know if a service node has an active subscription for this client
// peerID should contain the ID of a peer we are subscribed to, supporting the filter protocol
// If ms is greater than 0, the subscription must happen before the timeout
// (in milliseconds) is reached, or an error will be returned
//
extern int waku_filter_ping(char* peerID, int ms, WakuCallBack cb, void* userData);

// Sends a requests to a service node to stop pushing messages matching this filter to this client.
// It might be used to modify an existing subscription by providing a subset of the original filter
// criteria
//
//		{
//		  "pubsubTopic": "the pubsub topic" //  optional if using autosharding, mandatory if using static or named sharding.
//	      "contentTopics": ["the content topic"] // mandatory, at least one required, with a max of 10
//		}
//
// peerID should contain the ID of a peer this client is subscribed to.
// If ms is greater than 0, the subscription must happen before the timeout
// (in milliseconds) is reached, or an error will be returned
//
extern int waku_filter_unsubscribe(char* filterJSON, char* peerID, int ms, WakuCallBack cb, void* userData);

// Sends a requests to a service node (or all service nodes) to stop pushing messages
// peerID should contain the ID of a peer this client is subscribed to, or can be NULL to
// stop all active subscriptions
// If ms is greater than 0, the subscription must happen before the timeout
// (in milliseconds) is reached, or an error will be returned
//
extern int waku_filter_unsubscribe_all(char* peerID, int ms, WakuCallBack cb, void* userData);

// Creates a subscription to a light node matching a content filter and, optionally, a pubSub topic.
// filterJSON must contain a JSON with this format:
//
//	{
//		 "contentFilters": [ // mandatory
//	    {
//	      "contentTopic": "the content topic"
//	    }, ...
//	  ],
//	  "pubsubTopic": "the pubsub topic" // optional
//	}
//
// peerID should contain the ID of a peer supporting the filter protocol. Use NULL to automatically select a node
// If ms is greater than 0, the subscription must happen before the timeout
// (in milliseconds) is reached, or an error will be returned
//
extern int waku_legacy_filter_subscribe(char* filterJSON, char* peerID, int ms, WakuCallBack cb, void* userData);

// Removes subscriptions in a light node matching a content filter and, optionally, a pubSub topic.
// filterJSON must contain a JSON with this format:
//
//	{
//		 "contentFilters": [ // mandatory
//	    {
//	      "contentTopic": "the content topic"
//	    }, ...
//	  ],
//	  "pubsubTopic": "the pubsub topic" // optional
//	}
//
// If ms is greater than 0, the subscription must happen before the timeout
// (in milliseconds) is reached, or an error will be returned
//
extern int waku_legacy_filter_unsubscribe(char* filterJSON, int ms, WakuCallBack cb, void* userData);

// Publish a message using waku lightpush. Use NULL for topic to derive the pubsub topic from the contentTopic.
// peerID should contain the ID of a peer supporting the lightpush protocol. Use NULL to automatically select a node
// If ms is greater than 0, the broadcast of the message must happen before the timeout
// (in milliseconds) is reached, or an error will be returned
//
extern int waku_lightpush_publish(char* messageJSON, char* topic, char* peerID, int ms, WakuCallBack cb, void* userData);

// Determine if there are enough peers to publish a message on a topic. Use NULL
// to verify the number of peers in the default pubsub topic
//
extern int waku_relay_enough_peers(char* topic, WakuCallBack cb, void* userData);

// Publish a message using waku relay and returns the message ID. Use NULL for topic to use the default pubsub topic
// If ms is greater than 0, the broadcast of the message must happen before the timeout
// (in milliseconds) is reached, or an error will be returned.
//
extern int waku_relay_publish(char* messageJSON, char* topic, int ms, WakuCallBack cb, void* userData);

// Subscribe to WakuRelay to receive messages matching a content filter.
// filterJSON must contain a JSON with this format:
//
//		{
//		  "pubsubTopic": "the pubsub topic" // optional if using autosharding, mandatory if using static or named sharding.
//	      "contentTopics": ["the content topic"] // mandatory, at least one required, with a max of 10
//		}
//
// When a message is received, a "message" event is emitted containing the message and pubsub topic in which
// the message was received
//
extern int waku_relay_subscribe(char* contentfilterJSON, WakuCallBack cb, void* userData);

// Returns a json response with the list of pubsub topics the node
// is subscribed to in WakuRelay
//
extern int waku_relay_topics(WakuCallBack cb, void* userData);

// Closes the pubsub subscription to stop receiving messages matching a content filter
// filterJSON must contain a JSON with this format:
//
//		{
//		  "pubsubTopic": "the pubsub topic" // optional if using autosharding, mandatory if using static or named sharding.
//	      "contentTopics": ["the content topic"] // mandatory, at least one required, with a max of 10
//		}
//
extern int waku_relay_unsubscribe(char* filterJSON, WakuCallBack cb, void* userData);

// Query historic messages using waku store protocol.
// queryJSON must contain a valid json string with the following format:
//
//	{
//		"pubsubTopic": "...", // optional string
//	 "startTime": 1234, // optional, unix epoch time in nanoseconds
//	 "endTime": 1234, // optional, unix epoch time in nanoseconds
//	 "contentFilters": [ // optional
//			{
//		      contentTopic: "contentTopic1"
//			}, ...
//	 ],
//	 "pagingOptions": {// optional pagination information
//	     "pageSize": 40, // number
//			"cursor": { // optional
//				"digest": ...,
//				"receiverTime": ...,
//				"senderTime": ...,
//				"pubsubTopic" ...,
//	     }
//			"forward": true, // sort order
//	 }
//	}
//
// If a non empty cursor is returned, this function should be executed again, setting  the `cursor` attribute with the cursor returned in the response
// peerID should contain the ID of a peer supporting the store protocol. Use NULL to automatically select a node
// If ms is greater than 0, the broadcast of the message must happen before the timeout
// (in milliseconds) is reached, or an error will be returned
//
extern int waku_store_query(char* queryJSON, char* peerID, int ms, WakuCallBack cb, void* userData);

// Query historic messages stored in the localDB using waku store protocol.
// queryJSON must contain a valid json string with the following format:
//
//	{
//		"pubsubTopic": "...", // optional string
//	 "startTime": 1234, // optional, unix epoch time in nanoseconds
//	 "endTime": 1234, // optional, unix epoch time in nanoseconds
//	 "contentFilters": [ // optional
//			{
//		      contentTopic: "contentTopic1"
//			}, ...
//	 ],
//	 "pagingOptions": {// optional pagination information
//	     "pageSize": 40, // number
//			"cursor": { // optional
//				"digest": ...,
//				"receiverTime": ...,
//				"senderTime": ...,
//				"pubsubTopic" ...,
//	     }
//			"forward": true, // sort order
//	 }
//	}
//
// If a non empty cursor is returned, this function should be executed again, setting  the `cursor` attribute with the cursor returned in the response
// Requires the `store` option to be passed when setting up the initial configuration
//
extern int waku_store_local_query(char* queryJSON, WakuCallBack cb, void* userData);


```

# Tests

### DIFF
--- a/library/c/api.go
+++ b/library/c/api.go
@@ -6,11 +6,11 @@ package main
 #include <stddef.h>
 
 // The possible returned values for the functions that return int
-#define RET_OK                0
-#define RET_ERR               1
-#define RET_MISSING_CALLBACK  2
+static const int RET_OK = 0;
+static const int RET_ERR = 1;
+static const int RET_MISSING_CALLBACK = 2;
 
-typedef void (*WakuCallBack) (const char* msg, size_t len_0);
+typedef void (*WakuCallBack) (int ret_code, const char* msg, void * user_data);
 */
 import "C"
 import (
@@ -20,10 +20,6 @@ import (
 	"github.com/waku-org/go-waku/library"
 	"github.com/waku-org/go-waku/waku/v2/protocol"
 )
-
-const retOk = C.RET_OK
-const retErr = C.RET_ERR
-const retMissingCallback = C.RET_MISSING_CALLBACK
 
 func main() {}
 
@@ -92,25 +88,25 @@ func main() {}
 // - dns4DomainName: the domain name resolving to the node's public IPv4 address.
 //
 //export waku_new
-func waku_new(configJSON *C.char, onErrCb C.WakuCallBack) C.int {
+func waku_new(configJSON *C.char, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	err := library.NewNode(C.GoString(configJSON))
-	return execErrCB(onErrCb, err)
+	return onError(err, cb, userData)
 }
 
 // Starts the waku node
 //
 //export waku_start
-func waku_start(onErrCb C.WakuCallBack) C.int {
+func waku_start(onErr C.WakuCallBack, userData unsafe.Pointer) C.int {
 	err := library.Start()
-	return execErrCB(onErrCb, err)
+	return onError(err, onErr, userData)
 }
 
 // Stops a waku node
 //
 //export waku_stop
-func waku_stop(onErrCb C.WakuCallBack) C.int {
+func waku_stop(onErr C.WakuCallBack, userData unsafe.Pointer) C.int {
 	err := library.Stop()
-	return execErrCB(onErrCb, err)
+	return onError(err, onErr, userData)
 }
 
 // Determine is a node is started or not
@@ -126,96 +122,96 @@ func waku_is_started() C.int {
 
 type fn func() (string, error)
 
-func singleFnExec(f fn, onOkCb C.WakuCallBack, onErrCb C.WakuCallBack) C.int {
+func singleFnExec(f fn, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	result, err := f()
 	if err != nil {
-		return execErrCB(onErrCb, err)
+		return onError(err, cb, userData)
 	}
-	return execOkCB(onOkCb, result)
+	return onSuccesfulResponse(result, cb, userData)
 }
 
 // Obtain the peer ID of the waku node
 //
 //export waku_peerid
-func waku_peerid(onOkCb C.WakuCallBack, onErrCb C.WakuCallBack) C.int {
+func waku_peerid(cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	return singleFnExec(func() (string, error) {
 		return library.PeerID()
-	}, onOkCb, onErrCb)
+	}, cb, userData)
 }
 
 // Obtain the multiaddresses the wakunode is listening to
 //
 //export waku_listen_addresses
-func waku_listen_addresses(onOkCb C.WakuCallBack, onErrCb C.WakuCallBack) C.int {
+func waku_listen_addresses(cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	return singleFnExec(func() (string, error) {
 		return library.ListenAddresses()
-	}, onOkCb, onErrCb)
+	}, cb, userData)
 }
 
 // Add node multiaddress and protocol to the wakunode peerstore
 //
 //export waku_add_peer
-func waku_add_peer(address *C.char, protocolID *C.char, onOkCb C.WakuCallBack, onErrCb C.WakuCallBack) C.int {
+func waku_add_peer(address *C.char, protocolID *C.char, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	return singleFnExec(func() (string, error) {
 		return library.AddPeer(C.GoString(address), C.GoString(protocolID))
-	}, onOkCb, onErrCb)
+	}, cb, userData)
 }
 
 // Connect to peer at multiaddress. if ms > 0, cancel the function execution if it takes longer than N milliseconds
 //
 //export waku_connect
-func waku_connect(address *C.char, ms C.int, onErrCb C.WakuCallBack) C.int {
+func waku_connect(address *C.char, ms C.int, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	err := library.Connect(C.GoString(address), int(ms))
-	return execErrCB(onErrCb, err)
+	return onError(err, cb, userData)
 }
 
 // Connect to known peer by peerID. if ms > 0, cancel the function execution if it takes longer than N milliseconds
 //
 //export waku_connect_peerid
-func waku_connect_peerid(peerID *C.char, ms C.int, onErrCb C.WakuCallBack) C.int {
+func waku_connect_peerid(peerID *C.char, ms C.int, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	err := library.ConnectPeerID(C.GoString(peerID), int(ms))
-	return execErrCB(onErrCb, err)
+	return onError(err, cb, userData)
 }
 
 // Close connection to a known peer by peerID
 //
 //export waku_disconnect
-func waku_disconnect(peerID *C.char, onErrCb C.WakuCallBack) C.int {
+func waku_disconnect(peerID *C.char, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	err := library.Disconnect(C.GoString(peerID))
-	return execErrCB(onErrCb, err)
+	return onError(err, cb, userData)
 }
 
 // Get number of connected peers
 //
 //export waku_peer_cnt
-func waku_peer_cnt(onOkCb C.WakuCallBack, onErrCb C.WakuCallBack) C.int {
+func waku_peer_cnt(cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	return singleFnExec(func() (string, error) {
 		peerCnt, err := library.PeerCnt()
 		return fmt.Sprintf("%d", peerCnt), err
-	}, onOkCb, onErrCb)
+	}, cb, userData)
 }
 
 // Create a content topic string according to RFC 23
 //
 //export waku_content_topic
-func waku_content_topic(applicationName *C.char, applicationVersion C.uint, contentTopicName *C.char, encoding *C.char, onOkCb C.WakuCallBack) C.int {
+func waku_content_topic(applicationName *C.char, applicationVersion C.uint, contentTopicName *C.char, encoding *C.char, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	contentTopic, _ := protocol.NewContentTopic(C.GoString(applicationName), uint32(applicationVersion), C.GoString(contentTopicName), C.GoString(encoding))
-	return execOkCB(onOkCb, contentTopic.String())
+	return onSuccesfulResponse(contentTopic.String(), cb, userData)
 }
 
 // Create a pubsub topic string according to RFC 23
 //
 //export waku_pubsub_topic
-func waku_pubsub_topic(name *C.char, encoding *C.char, onOkCb C.WakuCallBack) C.int {
+func waku_pubsub_topic(name *C.char, encoding *C.char, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	topic := library.PubsubTopic(C.GoString(name), C.GoString(encoding))
-	return execOkCB(onOkCb, topic)
+	return onSuccesfulResponse(topic, cb, userData)
 }
 
 // Get the default pubsub topic used in waku2: /waku/2/default-waku/proto
 //
 //export waku_default_pubsub_topic
-func waku_default_pubsub_topic(onOkCb C.WakuCallBack) C.int {
-	return execOkCB(onOkCb, library.DefaultPubsubTopic())
+func waku_default_pubsub_topic(cb C.WakuCallBack, userData unsafe.Pointer) C.int {
+	return onSuccesfulResponse(library.DefaultPubsubTopic(), cb, userData)
 }
 
 // Register callback to act as signal handler and receive application signals
@@ -230,8 +226,8 @@ func waku_set_event_callback(cb C.WakuCallBack) {
 // Retrieve the list of peers known by the waku node
 //
 //export waku_peers
-func waku_peers(onOkCb C.WakuCallBack, onErrCb C.WakuCallBack) C.int {
+func waku_peers(cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	return singleFnExec(func() (string, error) {
 		return library.Peers()
-	}, onOkCb, onErrCb)
+	}, cb, userData)
 }
--- a/library/c/api_relay.go
+++ b/library/c/api_relay.go
@@ -4,58 +4,74 @@ package main
 #include <cgo_utils.h>
 */
 import "C"
-import "github.com/waku-org/go-waku/library"
+import (
+	"unsafe"
+
+	"github.com/waku-org/go-waku/library"
+)
 
 // Determine if there are enough peers to publish a message on a topic. Use NULL
 // to verify the number of peers in the default pubsub topic
 //
 //export waku_relay_enough_peers
-func waku_relay_enough_peers(topic *C.char, onOkCb C.WakuCallBack, onErrCb C.WakuCallBack) C.int {
+func waku_relay_enough_peers(topic *C.char, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	return singleFnExec(func() (string, error) {
 		result, err := library.RelayEnoughPeers(C.GoString(topic))
 		if result {
 			return "true", err
 		}
 		return "false", err
-	}, onOkCb, onErrCb)
+	}, cb, userData)
 }
 
-// Publish a message using waku relay and returns the message ID. Use NULL for topic to use the default pubsub topic
+// Publish a message using waku relay and returns the message ID. Use NULL for topic to derive the pubsub topic from the contentTopic.
 // If ms is greater than 0, the broadcast of the message must happen before the timeout
 // (in milliseconds) is reached, or an error will be returned.
 //
 //export waku_relay_publish
-func waku_relay_publish(messageJSON *C.char, topic *C.char, ms C.int, onOkCb C.WakuCallBack, onErrCb C.WakuCallBack) C.int {
+func waku_relay_publish(messageJSON *C.char, topic *C.char, ms C.int, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	return singleFnExec(func() (string, error) {
 		return library.RelayPublish(C.GoString(messageJSON), C.GoString(topic), int(ms))
-	}, onOkCb, onErrCb)
+	}, cb, userData)
 }
 
-// Subscribe to a WakuRelay topic. Set the topic to NULL to subscribe
-// to the default topic. Returns a json response. When a message is received,
-// a "message" event is emitted containing the message and pubsub topic in which
+// Subscribe to WakuRelay to receive messages matching a content filter.
+// filterJSON must contain a JSON with this format:
+//
+//		{
+//		  "pubsubTopic": "the pubsub topic" // optional if using autosharding, mandatory if using static or named sharding.
+//	      "contentTopics": ["the content topic"] // optional
+//		}
+//
+// When a message is received, a "message" event is emitted containing the message and pubsub topic in which
 // the message was received
 //
 //export waku_relay_subscribe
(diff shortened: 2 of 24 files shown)


Ivansete-status (commented):
It looks really nice!Thanks for this!
However, the approach may be a little bit different from what we need in `nwaku`. In `nwaku` we have this global variable (now in `master` branch) which holds the internal _Context_ struct:

https://github.com/waku-org/nwaku/blob/8897ae1a2f9ef4dac98f982a6250ed378bc3d234/library/waku_thread/waku_thread.nim#L29C11-L29C11

You don't have such a similar context variable, do you?

And, in an upcoming PR we will extract this as a `void*` when a `waku_init` or `waku_new` is called. Then, we will need to pass down this "context" variable from outside the library to the rest of the library function invocations.

On the other hand, I was considering the `userData` parameter as a user-defined context, which in the case of _Rust_, is the address of the closure. I had suggested the creation of the next function, which links the library _Context_ to the user-defined context, although I'd like to revisit this approach: https://github.com/waku-org/nwaku/blob/653a712ba1f6ccab016ddc1ebc9319b572db275f/library/libwaku.nim#L121

richard-ramos:
@Ivansete-status,
something to keep in mind is that if the context variable is added, it  will be possible to have more than one running node. Is that okay?

For having an extra `waku_set_user_data` function, I'm not sure I like this approach. 
The `user_data` attribute is associated to the function being called, and `waku_user_data` unnecessarily decouples the `user_data` from the function being executed, as well as potentially introducing bugs in case someone forgets to call `waku_set_user_data` before actually executing the intended function.

A problem I also identified while doing this PR is that in the end I might end up adding an additional parameter to the callback: the return code, as I ended up noticing that from inside the callback it is actually not possible to know whether the function was executed correctly or not. So in the end WakuCallback might look like this:
```
typedef void (*WakuCallBack) (int ret_code, const char* msg, void * user_data);
```

Ivansete-status:
> @Ivansete-status, something to keep in mind is that if the context variable is added, it will be possible to have more than one running node. Is that okay?

Yes correct!

----

> For having an extra `waku_set_user_data` function, I'm not sure I like this approach. The `user_data` attribute is associated to the function being called, and `waku_user_data` unnecessarily decouples the `user_data` from the function being executed, as well as potentially introducing bugs in case someone forgets to call `waku_set_user_data` before actually executing the intended function.

I completely agree with that! I will avoid using the `waku_set_user_data` approach in `nwaku` then.

----

> A problem I also identified while doing this PR is that in the end I might end up adding an additional parameter to the callback: the return code, as I ended up noticing that from inside the callback it is actually not possible to know whether the function was executed correctly or not. So in the end WakuCallback might look like this:
> 
> ```
> typedef void (*WakuCallBack) (int ret_code, const char* msg, void * user_data);
> ```

Sorry but I don't fully understand this reasoning. Let's comment that tomorrow :)

richard-ramos:
@harsh-98 @chaitanyaprem sorry for the large PR :( 
this should be ready to review now.

@Ivansete-status I'll extract the context from `waku_init` in a separate PR

library/relay.go
```diff
@@ -55,43 +50,38 @@ func RelayPublish(messageJSON string, topic string, ms int) (string, error) {
...
-	if ok {
-		return nil
+func relaySubscribe(filterJSON string) error {
+	cf, err := toContentFilter(filterJSON)
+	if err != nil {
+		return err
 	}
 
-	subscription, err := wakuState.node.Relay().Subscribe(context.Background(), protocol.NewContentFilter(topicToSubscribe))
+	subscription, err := wakuState.node.Relay().Subscribe(context.Background(), cf)
```
chaitanyaprem, line 62: small nit: rename to subscriptions

library/encoding.go
```diff
@@ -137,7 +137,12 @@ func DecodeSymmetric(messageJSON string, symmetricKey string) (string, error) {
 	}
 
 	if msg.Version == 0 {
-		return marshalJSON(msg.Payload)
+		return marshalJSON(struct {
```
chaitanyaprem, line 140: Better ti define this struct as type and use it as i see 2 copies of same definition.

library/c/api_relay.go
```diff
@@ -4,58 +4,74 @@ package main
...
-// Publish a message using waku relay and returns the message ID. Use NULL for topic to use the default pubsub topic
+// Publish a message using waku relay and returns the message ID. Use NULL for topic to derive the pubsub topic from the contentTopic.
 // If ms is greater than 0, the broadcast of the message must happen before the timeout
 // (in milliseconds) is reached, or an error will be returned.
 //
 //export waku_relay_publish
-func waku_relay_publish(messageJSON *C.char, topic *C.char, ms C.int, onOkCb C.WakuCallBack, onErrCb C.WakuCallBack) C.int {
+func waku_relay_publish(messageJSON *C.char, topic *C.char, ms C.int, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	return singleFnExec(func() (string, error) {
 		return library.RelayPublish(C.GoString(messageJSON), C.GoString(topic), int(ms))
```
chaitanyaprem, line 34: would the call to C.GoString be fine and not cause issue if topic is NULL?

library/c/api_relay.go
```diff
@@ -4,58 +4,74 @@ package main
...
 
-// Subscribe to a WakuRelay topic. Set the topic to NULL to subscribe
-// to the default topic. Returns a json response. When a message is received,
-// a "message" event is emitted containing the message and pubsub topic in which
+// Subscribe to WakuRelay to receive messages matching a content filter.
+// filterJSON must contain a JSON with this format:
+//
+//		{
+//		  "pubsubTopic": "the pubsub topic" // optional if using autosharding, mandatory if using static or named sharding.
+//	      "contentTopics": ["the content topic"] // mandatory, at least one required, with a max of 10
```
chaitanyaprem, line 43: In relay subscribe, contentTopic is not mandatory as user can subscribe to just pubsubTopic as well without specifying a filter.

library/c/api_relay.go
```diff
@@ -4,58 +4,74 @@ package main
...
+	}, cb, userData)
 }
 
-// Closes the pubsub subscription to a pubsub topic
+// Closes the pubsub subscription to stop receiving messages matching a content filter
+// filterJSON must contain a JSON with this format:
+//
+//		{
+//		  "pubsubTopic": "the pubsub topic" // optional if using autosharding, mandatory if using static or named sharding.
+//	      "contentTopics": ["the content topic"] // mandatory, at least one required, with a max of 10
```
chaitanyaprem, line 70: Same comment as above.

library/c/api_store.go
```diff
@@ -4,30 +4,33 @@ package main
...
+	"unsafe"
+
+	"github.com/waku-org/go-waku/library"
+)
 
 // Query historic messages using waku store protocol.
 // queryJSON must contain a valid json string with the following format:
 //
 //	{
 //		"pubsubTopic": "...", // optional string
```
chaitanyaprem, line 17: For store query either a pubsubTopic or a contentTopic should be present right. 
Maybe we should mention that here.

library/c/api.go
```diff
@@ -92,25 +88,25 @@ func main() {}
 // - dns4DomainName: the domain name resolving to the node's public IPv4 address.
 //
 //export waku_new
-func waku_new(configJSON *C.char, onErrCb C.WakuCallBack) C.int {
+func waku_new(configJSON *C.char, cb C.WakuCallBack, userData unsafe.Pointer) C.int {
 	err := library.NewNode(C.GoString(configJSON))
-	return execErrCB(onErrCb, err)
+	return execErrCB(err, cb, userData)
```
chaitanyaprem, line 93: This name is confusing, the function actually executes callback and internally handles error, maybe we should rename it to something else.

chaitanyaprem (approved):
LGTM

Left some minor comments.

library/c/api_relay.go
richard-ramos: It doesn't cause a problem. `C.GoString(nil) == ""`

library/c/api_store.go
richard-ramos: It is possible for both fields to be empty to return all results with no filtering

chaitanyaprem: Probably. 
But we may want to keep the store protocol similar in the sense that wildcard store query like this should still be supported.
It would be interesting because how will the store client know if the storeNode is supporting all contentTopics it is interested in.
cc @SionoiS , store protocol may need an update to mandate contentTopic and/or pubsubTopic.

PR #787, opened 2023-09-29
Update peer selection options for light protocols

chaitanyaprem:
# Description
Update req/resp protocols to use new peerSelection logic based on pubSub/contentTopic as per #680 

@richard-ramos , it is becoming challenging to maintain both peerManager and option of using waku protocols without peer-manager (i.e using utils peer code). As you can see in this PR, there already is a lot of checks and duplicate code generated.

Now with peer-selection based on pubSubTopic, which cannot be done by default libp2p peerStore alone(since we store peer specific pubSub topics in wakuPeerStore), I am wondering if it is time to retire this option of peerutils without peerManager.
Maybe once we have WakuPeerStore implementation which is completely independent of libp2p peerStore as per #579  we can bring this option back. WDYT?

# Changes

- [x] Update peerSelection in lightpush to use peers filtered based on pubSubTopic.
- [ ] Update Filter to use peerSelection options based on pubSubTopic
- [ ] Update Store protocol to also use peerSelection based on pubSubTopic

# Tests

### DIFF
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -483,3 +483,17 @@ func (pm *PeerManager) selectServicePeer(proto protocol.ID, pubSubTopic string, 
 	}
 	return
 }
+
+func (pm *PeerManager) HandlePeerSelection(selectionType utils.PeerSelection, proto protocol.ID,
+	pubSubTopic string, specificPeers ...peer.ID) (peer.ID, error) {
+
+	switch selectionType {
+	case utils.Automatic:
+		return pm.SelectPeer(proto, pubSubTopic, specificPeers...)
+	case utils.LowestRTT:
+		//TODO: Move this to peer-manager
+		return utils.SelectPeerWithLowestRTT(context.Background(), pm.host, proto, specificPeers, pm.logger)
+	default:
+		return "", errors.New("unknown peer selection type specified")
+	}
+}
--- a/waku/v2/protocol/lightpush/waku_lightpush.go
+++ b/waku/v2/protocol/lightpush/waku_lightpush.go
@@ -17,6 +17,7 @@ import (
 	"github.com/waku-org/go-waku/waku/v2/protocol/lightpush/pb"
 	wpb "github.com/waku-org/go-waku/waku/v2/protocol/pb"
 	"github.com/waku-org/go-waku/waku/v2/protocol/relay"
+	"github.com/waku-org/go-waku/waku/v2/utils"
 	"go.uber.org/zap"
 )
 
@@ -150,11 +151,6 @@ func (wakuLP *WakuLightPush) request(ctx context.Context, req *pb.PushRequest, p
 		return nil, errors.New("lightpush params are mandatory")
 	}
 
-	if params.selectedPeer == "" {
-		wakuLP.metrics.RecordError(peerNotFoundFailure)
-		return nil, ErrNoPeersAvailable
-	}
-
 	if len(params.requestID) == 0 {
 		return nil, ErrInvalidID
 	}
@@ -210,16 +206,12 @@ func (wakuLP *WakuLightPush) Stop() {
 	wakuLP.h.RemoveStreamHandler(LightPushID_v20beta1)
 }
 
-// Optional PublishToTopic is used to broadcast a WakuMessage to a pubsub topic via lightpush protocol
-// If pubSubTopic is not provided, then contentTopic is use to derive the relevant pubSubTopic via autosharding.
-func (wakuLP *WakuLightPush) PublishToTopic(ctx context.Context, message *wpb.WakuMessage, opts ...Option) ([]byte, error) {
-	if message == nil {
-		return nil, errors.New("message can't be null")
-	}
+func (wakuLP *WakuLightPush) handleOpts(ctx context.Context, message *wpb.WakuMessage, opts ...Option) (*lightPushParameters, error) {
 	params := new(lightPushParameters)
 	params.host = wakuLP.h
 	params.log = wakuLP.log
 	params.pm = wakuLP.pm
+	var err error
 
 	optList := append(DefaultOptions(wakuLP.h), opts...)
 	for _, opt := range optList {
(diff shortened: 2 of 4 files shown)


chaitanyaprem:
@richard-ramos , it is becoming challenging to maintain both peerManager and option of using waku protocols without peer-manager (i.e using utils peer code). As you can see in this PR, there already is a lot of checks and duplicate code generated.

Now with peer-selection based on pubSubTopic, which cannot be done by default libp2p peerStore alone(since we store peer specific pubSub topics in wakuPeerStore), I am wondering if it is time to retire this option of peerutils without peerManager.
Maybe once we have WakuPeerStore implementation which is completely independent of libp2p peerStore as per #579  we can bring this option back. WDYT?

richard-ramos:
As long as the scenarios covered by the utils function can be met with the peer manager, it should be okay to retire the functions:
- choose a random peer (optionally selecting  the peer from a specific list)
- chose an specific peer (optionally selecting  the peer from a specific list)
- fastest peer (optionally selecting the peer from a specific list)

chaitanyaprem:
> As long as the scenarios covered by the utils function can be met with the peer manager, it should be okay to retire the functions:
> 
>     * choose a random peer (optionally selecting  the peer from a specific list)
> 
>     * chose an specific peer
> 
>     * fastest peer (optionally selecting the peer from a specific list)

The scenarios would be covered in peerManager. 
My only concern was that during one of the initial discussions we had, you indicated that we wanted to keep the Waku protocols go-implementation in such a way that they could also be used independently using just a libp2p host and not necessarily by initializing a WakuNode.
If we retire these functions, then these options cannot be used for someone wanting to use Waku protocols without initializing a WakuNode.

I guess it i still fine, going ahead with deleting some more code :)

waku/v2/peermanager/peer_manager.go
```diff
@@ -45,6 +47,19 @@ type PeerManager struct {
 	subRelayTopics      map[string]*NodeTopicDetails
 }
 
+// PeerSelection provides various options based on which Peer is selected from a list of peers.
+type PeerSelection int
+
+const (
+	Unknown PeerSelection = iota
+	Automatic
+	LowestRTT
```
richard-ramos, line 56: What do you think if we make `Automatic` the default and remove `Unknown`? that way it is not necessary to specify the peer selection criteria, and there would not be any need of having a `errors.New("unknown peer selection type specified")` error

waku/v2/peermanager/peer_manager.go
```diff
@@ -483,3 +498,140 @@ func (pm *PeerManager) selectServicePeer(proto protocol.ID, pubSubTopic string,
...
+				ctx, cancel := context.WithTimeout(criteria.Ctx, 3*time.Second)
+				defer cancel()
+				result := <-ping.Ping(ctx, pm.host, p)
+				if result.Error == nil {
+					pingCh <- pingResult{
+						p:   p,
+						rtt: result.RTT,
+					}
+				} else {
+					fmt.Println("Error in Ping", result)
```
richard-ramos, line 573:
```suggestion
					pm.logger.Error("could not ping", logging.HostID("peer", p), zap.Error(result.Error))
```

Do consider changing it to `Debug` tho, since there is going to be a lot of error logs depending on the number of peers, and failing a ping should be pretty common

waku/v2/peermanager/peer_manager.go
```diff
@@ -483,3 +498,140 @@ func (pm *PeerManager) selectServicePeer(proto protocol.ID, pubSubTopic string,
...
+		wg.Wait()
+		close(waitCh)
+		close(pingCh)
+	}()
+
+	select {
+	case <-waitCh:
+		var min *pingResult
+		for p := range pingCh {
+			fmt.Println("ping result", p)
```
richard-ramos, line 586:
```suggestion
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -483,3 +498,140 @@ func (pm *PeerManager) selectServicePeer(proto protocol.ID, pubSubTopic string,
...
+			if min == nil {
+				min = &p
+			} else {
+				if p.rtt < min.rtt {
+					min = &p
+				}
+			}
+		}
+		if min == nil {
+			pm.logger.Info("Could not find min")
```
richard-ramos, line 596: I'd remove this log. It's not really informative, and we're already returning an error, so we might end up double logging the same event

richard-ramos (commented):
Looking good.
Left some few comments/questions

waku/v2/peermanager/peer_manager.go
chaitanyaprem: Good idea, will do that.

waku/v2/protocol/store/waku_store_protocol_test.go
```diff
@@ -60,7 +60,9 @@ func TestWakuStoreProtocolQuery(t *testing.T) {
 		ContentTopics: []string{topic1},
 	}
 
-	response, err := s2.Query(ctx, q, DefaultOptions()...)
+	var hrOptions []HistoryRequestOption
+	hrOptions = append(hrOptions, WithPeer(host1.ID()))
+	response, err := s2.Query(ctx, q, hrOptions...)
```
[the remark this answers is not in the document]
chaitanyaprem: Yes, because now peermanager is required for peerSelection it is difficult to write unit tests at each protocol level for these. Unit tests at peerManager level validate automatic PeerSelection, just that at each protocol level we are not able to have tests.
Maybe this can be addressed by having such tests at WakuNode level.
Will create a separate issue to add tests for this to analyze and add.

waku/v2/peermanager/peer_manager_test.go
```diff
@@ -20,16 +21,22 @@ import (
 
 func getAddr(h host.Host) multiaddr.Multiaddr {
 	id, _ := multiaddr.NewMultiaddr(fmt.Sprintf("/p2p/%s", h.ID().Pretty()))
-	return h.Network().ListenAddresses()[0].Encapsulate(id)
+	var selectedAddr multiaddr.Multiaddr
+	//For now skipping circuit relay addresses as libp2p seems to be returning empty p2p-circuit addresses.
```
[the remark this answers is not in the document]
chaitanyaprem: https://github.com/waku-org/go-waku/issues/802

waku/v2/peermanager/peer_manager_test.go
```diff
@@ -20,16 +21,22 @@ import (
...
+	for _, addr := range h.Network().ListenAddresses() {
+		if strings.Contains(addr.String(), "p2p-circuit") {
+			continue
+		}
```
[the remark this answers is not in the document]
chaitanyaprem: This change won't work, because the multiaddr returned by libp2p which is not valid is just `/p2p-circuit`. Since this matches with value for protocol, the above check doesn't work. So, keeping the change as is.

waku/v2/node/wakunode2_test.go
```diff
@@ -303,9 +304,10 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
 	require.NoError(t, err)
 	defer wakuNode3.Stop()
 
-	err = wakuNode3.DialPeerWithMultiAddress(ctx, wakuNode2.ListenAddresses()[0])
+	//err = wakuNode3.DialPeerWithMultiAddress(ctx, wakuNode2.ListenAddresses()[0])
```
richard-ramos, line 307:
```suggestion
```

richard-ramos (approved):

PR #792, opened 2023-10-03
Test: container image workflow

romanzac:
# Description
Github workflow to build docker images for testing

# Changes
- Workflow file
- Docker file

# Tests
It has been tested with quay.io docker registry and go-waku repo specific credentials. Final test result visible at https://github.com/waku-org/go-waku/actions/runs/6392816182


richard-ramos (approved):
LGTM. 
Just a question: are the dockerfile from docker/ folder and the one in the root folder seem to be the same? is this okay? or maybe it's missing a commit?

romanzac:
> LGTM. Just a question: are the dockerfile from docker/ folder and the one in the root folder seem to be the same? is this okay? or maybe it's missing a commit?

Dockerfiles in /docker might be for testing, unless we agree otherwise. They are few differences in labels and bumped Go version. Initially, I didn't want to interfere with Dockerfile created by devs.  Later we may also think about reference build images for different platforms and have them published on DockerHub.

PR #794, opened 2023-10-06
fix: return errors in FilterSubscribeOption

richard-ramos:
# Description
Make the options for filter subscription return an err


romanzac:
I've retested, and got en error as expected. LGTM.

Thanks.

romanzac (approved):
I've retested, and got en error as expected. 

Thanks.

Issue #746, opened 2023-09-15
feat: filterv2 REST API


richard-ramos:
Implement filterV2 REST API, following this openapi yml file: https://github.com/waku-org/nwaku/pull/1890/files#diff-d0d5d66c0f96582065344acd54ce7f1b65fb973ed201435ac0689020c3464dce 
The REST api is currently located at https://github.com/waku-org/go-waku/tree/master/cmd/waku/server/rest

The focus of the task should be limited to only the REST server as the RPC can be considered deprecated and will be removed eventually so it's not worth it to add FilterV2 functionality to it.

richard-ramos:
I think this is where we need to deviate from the defined openapi.yaml file.

We need to be able to optionally define the pubsub topic. If it is not defined, we need to derive the pubsub topic from the content topics (similar to this https://github.com/waku-org/go-waku/blob/5dfbd98c74d21664caea1d6722c28d120f52c04b/waku/v2/protocol/filter/client.go#L144C3-L156), but if specified, we need to return the messages that match both the content topic and pubsub topic.

richard-ramos:
Perhaps a separate endpoint? i.e.:  `/filter/v2/messages_by_pubsubTopic/{pubsubTopic}/{contentTopic}`:

chaitanyaprem:
@richard-ramos is this something that is required on rust-bindings?
Else we can descope this from go-waku.

richard-ramos:
Not required. REST server is not used in bindings.

PR #798, opened 2023-10-09
feat: add filter v2 rpc

harsh-98:
ping, subscribe/unsubscribe and unsubscribeAll.

# Description
Filter rest endpoints for #746 . Getting current messages is not implemented.

# Changes

- [x] Add rest codebase for filterv2 endpoints ping, subscribe/unsubscribe and unsubscribeAll.
- [x] Fix writeResponse. http.ResponseWriter doesn't support writing multiple statusCode from 2xx-5xx. w.Write implicitly write http.StatusOK. Other code can't be written.
- [x] Fix https://github.com/waku-org/go-waku/pull/836
- [x] Change ping function on filter client pass PingOptions for requestID. 

# Tests

- [x] Write test for pingFailure, subscribe and then check if ping works, subscribe/unsubscribe and multiple subscription then unsubscribeAll.

# TODO
- Separate PR for supporting getting messages from filter client.

### DIFF
--- a/waku/v2/peermanager/peer_connector.go
+++ b/waku/v2/peermanager/peer_connector.go
@@ -35,10 +35,15 @@ type PeerConnectionStrategy struct {
 	paused      atomic.Bool
 	dialTimeout time.Duration
 	*CommonDiscoveryService
-	subscriptions []<-chan PeerData
+	subscriptions []subscription
 
 	backoff backoff.BackoffFactory
 	logger  *zap.Logger
+}
+
+type subscription struct {
+	ctx context.Context
+	ch  <-chan PeerData
 }
 
 // backoff describes the strategy used to decide how long to backoff after previously attempting to connect to a peer
@@ -85,7 +90,7 @@ func (c *PeerConnectionStrategy) Subscribe(ctx context.Context, ch <-chan PeerDa
 	// if not running yet, store the subscription and return
 	if err := c.ErrOnNotRunning(); err != nil {
 		c.mux.Lock()
-		c.subscriptions = append(c.subscriptions, ch)
+		c.subscriptions = append(c.subscriptions, subscription{ctx, ch})
 		c.mux.Unlock()
 		return
 	}
@@ -93,15 +98,17 @@ func (c *PeerConnectionStrategy) Subscribe(ctx context.Context, ch <-chan PeerDa
 	c.WaitGroup().Add(1)
 	go func() {
 		defer c.WaitGroup().Done()
-		c.consumeSubscription(ch)
+		c.consumeSubscription(subscription{ctx, ch})
 	}()
 }
 
-func (c *PeerConnectionStrategy) consumeSubscription(ch <-chan PeerData) {
+func (c *PeerConnectionStrategy) consumeSubscription(s subscription) {
 	for {
 		// for returning from the loop when peerConnector is paused.
 		select {
 		case <-c.Context().Done():
+			return
+		case <-s.ctx.Done():
 			return
 		default:
 		}
@@ -110,7 +117,9 @@ func (c *PeerConnectionStrategy) consumeSubscription(ch <-chan PeerData) {
 			select {
 			case <-c.Context().Done():
 				return
-			case p, ok := <-ch:
+			case <-s.ctx.Done():
+				return
+			case p, ok := <-s.ch:
 				if !ok {
 					return
 				}
@@ -166,7 +175,7 @@ func (c *PeerConnectionStrategy) isPaused() bool {
 func (c *PeerConnectionStrategy) consumeSubscriptions() {
 	for _, subs := range c.subscriptions {
 		c.WaitGroup().Add(1)
-		go func(s <-chan PeerData) {
+		go func(s subscription) {
 			defer c.WaitGroup().Done()
 			c.consumeSubscription(s)
 		}(subs)
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -339,7 +339,7 @@ func (pm *PeerManager) AddDiscoveredPeer(p PeerData, connectNow bool) {
 		}
 	}
 	if connectNow {
-		pm.peerConnector.PushToChan(p)
+		go pm.peerConnector.PushToChan(p)
 	}
 }
 
--- a/waku/v2/protocol/filter/client.go
+++ b/waku/v2/protocol/filter/client.go
@@ -134,6 +134,11 @@ func (wf *WakuFilterLightNode) onRequest(ctx context.Context) func(network.Strea
 
 		stream.Close()
 
+		if err = messagePush.Validate(); err != nil {
+			logger.Warn("received invalid messagepush")
+			return
+		}
+
 		pubSubTopic := ""
 		//For now returning failure, this will get addressed with autosharding changes for filter.
 		if messagePush.PubsubTopic == nil {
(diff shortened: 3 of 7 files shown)


richard-ramos:
Peer selection is done automatically. Look here how it is done in nwaku:
https://github.com/waku-org/nwaku/blob/880d018eaf9fbc8c94bb0da394140993f8d94e8c/waku/node/rest/filter/handlers.nim#L162-L169

Here's the relevant logic for how peers are selected https://github.com/waku-org/nwaku/blob/d5c3ade5e21eafe17ca1c3b6e43e0563df923beb/waku/node/peer_manager/peer_manager.nim#L700-L707 do note that service slots have priority. If there are no peers in the service slots, then a random peer with filter protocol support is selected

richard-ramos:
The same applies to Ping. The request seems to be missing the peerId.

cmd/waku/server/rest/filter.go
```diff
@@ -0,0 +1,265 @@
...
+}
+func (r filterRequestId) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf(`"%s"`, r.String())), nil
+}
+
+const filterv2Ping = "/filter/v2/subscriptions/{requestId}"
+const filterv2Subscribe = "/filter/v2/subscriptions"
+const filterv2SubscribeAll = "/filter/v2/subscriptions/all"
+
+// RelayService represents the REST service for WakuRelay
```
chaitanyaprem, line 43: FilterService

cmd/waku/server/rest/filter.go
```diff
@@ -0,0 +1,265 @@
...
+
+func (r filterRequestId) String() string {
+	return hex.EncodeToString(r)
+}
+func (r filterRequestId) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf(`"%s"`, r.String())), nil
+}
+
+const filterv2Ping = "/filter/v2/subscriptions/{requestId}"
+const filterv2Subscribe = "/filter/v2/subscriptions"
```
chaitanyaprem, line 40: Small nit: Rename to filterv2Subscription

cmd/waku/server/rest/filter.go
```diff
@@ -0,0 +1,265 @@
...
+		s.log.Error("bad request id", zap.Error(err))
+		writeResponse(w, &filterSubscriptionResponse{
+			RequestId:  requestId,
+			StatusDesc: "bad request id",
+		}, http.StatusBadRequest)
+		return
+	}
+
+	// selecting random peer that supports filter protocol
+	peerId := s.getRandomFilterPeer(req.Context(), requestId, w)
```
chaitanyaprem, line 80: I don't think this is a good idea, because if we are randomly selecting a peer from peerStore then there is a no gaurantee that we have a subscription with this peer which will inturn result in ping failure. 
Instead, we should take contentFilter as an argument for ping and use https://github.com/waku-org/go-waku/blob/2701a38b2aef3b1ece7b2aef7c8df13a88e069c5/waku/v2/protocol/filter/client.go#L332 to get the subscription and check if it is alive.

But since this looks like a change is required for REST endpoint, could be taken up later.

cmd/waku/server/rest/filter.go
```diff
@@ -0,0 +1,265 @@
...
+	PubsubTopic    string          `json:"pubsubTopic"`
+}
+
+type filterSubscriptionResponse struct {
+	RequestId  filterRequestId `json:"requestId"`
+	StatusDesc string          `json:"statusDesc"`
+}
+
+func toProtocolContentFilter(req filterSubscriptionRequest) protocol.ContentFilter {
+	contentTopics := map[string]struct{}{}
```
chaitanyaprem, line 117: Instead of this, better create the contentFilter using https://github.com/waku-org/go-waku/blob/2701a38b2aef3b1ece7b2aef7c8df13a88e069c5/waku/v2/protocol/content_filter.go#L31 by passing pubSubTopic and contentTopics.

cmd/waku/server/rest/filter.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+}
+
+// 400 on invalid request
+// 404 on failed subscription
+// 200 on single returned successful subscription
+// NOTE: subscribe on filter client randomly selects a peer if missing for given pubSubTopic
+func (s *FilterService) subscribe(w http.ResponseWriter, req *http.Request) {
+	message := filterSubscriptionRequest{}
+	if !s.readBody(w, req, &message) {
```
chaitanyaprem, line 133: Shouldn't we respond with 400 here?

cmd/waku/server/rest/filter.go
```diff
@@ -0,0 +1,265 @@
...
+		return
+	}
+
+	//
+	_, err := s.node.FilterLightnode().Subscribe(req.Context(), toProtocolContentFilter(message),
+		filter.WithRequestID(message.RequestId))
+	if err != nil {
```
chaitanyaprem, line 140: Do note that there is a possibility of a partial failure as internally the API can generate multiple requests. 
That needs to be addressed here indicating which ones have failed and which have succeeded.

cmd/waku/server/rest/filter.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	// unsubscribe on filter
+	_, err := s.node.FilterLightnode().Unsubscribe(
+		req.Context(),
+		toProtocolContentFilter(message),
+		filter.WithRequestID(message.RequestId),
+		filter.WithPeer(peerId),
+	)
+	if err != nil {
```
chaitanyaprem, line 179: Same comment as above, see how partial failures are to be reported back.

cmd/waku/server/rest/filter.go
```diff
@@ -0,0 +1,265 @@
...
+// 500 on failed subscription
+// 200 on all successful unsubscribe
+// unsubscribe all subscriptions for a given peer
+func (s *FilterService) unsubscribeAll(w http.ResponseWriter, req *http.Request) {
+	message := filterUnsubscribeAllRequest{}
+	if !s.readBody(w, req, &message) {
+		return
+	}
+
+	peerId := s.getRandomFilterPeer(req.Context(), message.RequestId, w)
```
chaitanyaprem, line 222: UnsubscribeAll doesn't need a peerID to be passed, as it unsubscribes from all existing filter subscriptions towards all peers. 

See https://github.com/waku-org/go-waku/blob/2701a38b2aef3b1ece7b2aef7c8df13a88e069c5/waku/v2/protocol/filter/client.go#L571

cmd/waku/server/rest/filter_api.yaml
```diff
@@ -0,0 +1,231 @@
...
+
+  /filter/v2/subscriptions:
+    post: # post_waku_v2_filter_v2_subscription
+      summary: Subscribe a peer to an array of content topics under a pubsubTopic
+      description: |
+          Subscribe a peer to an array of content topics under a pubsubTopic.
+
+          It is allowed to refresh or add new content topic to an existing subscription.
+
+          Fields pubsubTopic and contentFilters must be filled.
```
chaitanyaprem, line 65: With autosharding, pubsubTopic is not mandatory. 
Only contentTopics can be filled and using autosharding algorithm, pubsubTopic is derived. 
cc @NagyZoltanPeter , we need to change documentation for all APIs accordingly.

cmd/waku/server/rest/filter_api.yaml
```diff
@@ -0,0 +1,231 @@
...
+        '5XX':
+          description: Unexpected error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/FilterSubscriptionResponse'
+
+  /filter/v2/subscriptions/all:
+    delete: # delete_waku_v2_filter_v2_subscription
+      summary: Unsubscribe a peer from all content topics
```
chaitanyaprem, line 141: This is not unsubscribing a peer, rather unsubscribe from all filter peers for everything.
cc @NagyZoltanPeter

cmd/waku/server/rest/filter.go
chaitanyaprem: Ok, got it.

cmd/waku/server/rest/filter.go
chaitanyaprem: One way i could think of is to return 200 OK for such case and then include in the responseBody the successful and failed parts separately.

cmd/waku/server/rest/filter_api.yaml
chaitanyaprem: Feel free to move it there referring to this comment and mark this one as resolved.

chaitanyaprem (approved):
LGTM

PR #800, opened 2023-10-12
fix(store): query time comparison and max rows per page

richard-ramos:
# Description
Fixes an issue reported by @fbarbu15 where a store was not including a message whose timestamp matched the starttime of the query, and also sets the max page size to 20 to match nwaku

### DIFF
--- a/waku/persistence/db_key.go
+++ b/waku/persistence/db_key.go
@@ -35,7 +35,10 @@ func (k *DBKey) Bytes() []byte {
 
 // NewDBKey creates a new DBKey with the given values.
 func NewDBKey(senderTimestamp uint64, receiverTimestamp uint64, pubsubTopic string, digest []byte) *DBKey {
-	pubSubHash := hash.SHA256([]byte(pubsubTopic))
+	pubSubHash := make([]byte, PubsubTopicLength)
+	if pubsubTopic != "" {
+		pubSubHash = hash.SHA256([]byte(pubsubTopic))
+	}
 
 	var k DBKey
 	k.raw = make([]byte, DBKeyLength)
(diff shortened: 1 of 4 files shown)


waku/persistence/db_key.go
```diff
@@ -35,7 +35,10 @@ func (k *DBKey) Bytes() []byte {
 
 // NewDBKey creates a new DBKey with the given values.
 func NewDBKey(senderTimestamp uint64, receiverTimestamp uint64, pubsubTopic string, digest []byte) *DBKey {
-	pubSubHash := hash.SHA256([]byte(pubsubTopic))
+	pubSubHash := make([]byte, PubsubTopicLength)
+	if pubsubTopic != "" {
```
chaitanyaprem, line 39: Since we have the pubSubTopic in the StoreQuery, shouldn't we include that in forming the DB key and throw an error if not included?
Do you see a scenario of a valid store query without pubSubTopic?

richard-ramos: This would happen when you're doing a query and you don't specify the pubsub topic. Perhaps it should be clearer and use a `*string` instead of `string`? that way the intention of not using a pubsub topic is clearer. wdyt?

chaitanyaprem: Ok.
Now i have gone through the store code in a little more detail to realize that we are storing all messages in a single DB. I was under assumption that some sort of pubSubTopic based table is being maintained which is what lead me to think that it would be mandatory.

I got confused by looking at the NewDBKey since it is being generated by including pubSubTopic. It is just a way to get uniqueID for the message that is being stored. The actual query is just filtering messages from the DB based on actual fields stored for the message which would mean, it doesn't matter if pubSubTopic is provided or not. 

Maybe when store protocol is enhanced/upgraded we should consider partitioning the DB based on pubSubTopic and futher optimizing it based on contentTopics or something like that.

You can ignore my comment as it is fine with the way store is implemented currently.

fbarbu15 (approved):
Thanks

Issue #720, opened 2023-09-08
chore: Migrate to use proto files from common proto repo for all Waku protocols


chaitanyaprem:
### Background
As of now, Waku protocol proto files are duplicated in go-waku repo from https://github.com/waku-org/waku-proto and then the code is generated from them to use. This causes trouble because we need to keep track of reference proto file and its changes and manually diff and apply changes when a new version is created. 
This process would be errorprone.
Instead, once proto files are migrated to waku-org repo, generate all protobuf code from those proto files by making that repo a sub-module of go-waku repo. 

### Details

Note: Original proto files may have different package convention than what is used in go-waku, hence protogen command takes a flag which can specify the package to be used for the generated code.

### Acceptance criteria
Need to make sure all unit tests pass. 
Ensure interop tests with nwaku also pass.

PR #805, opened 2023-10-15
feat: metadata protocol

### LINKED ISSUE #799: feat: add new metadata protocol
Spec: https://github.com/vacp2p/rfc/pull/619

NWaku PR: https://github.com/waku-org/nwaku/pull/2062

Feature required for all implementations of TheWakuNetwork.

richard-ramos:
# Description
WakuMetadataProtocol implementation and tests

## Issue

closes #799


waku/v2/peermanager/topic_event_handler.go
```diff
@@ -118,6 +120,37 @@ func (pm *PeerManager) handlerPeerTopicEvent(peerEvt relay.EvtPeerTopic) {
 	wps := pm.host.Peerstore().(*wps.WakuPeerstoreImpl)
 	peerID := peerEvt.PeerID
 	if peerEvt.State == relay.PEER_JOINED {
+		// Metadata verification is done only if a clusterID is specified
```
alrevuelta, line 123: Not sure how this is used but perhaps its more efficient in the `ConnectionGater`?

Example https://github.com/prysmaticlabs/prysm/blob/develop/beacon-chain/p2p/connection_gater.go#L43C19-L61

richard-ramos: I'll try. Not sure if i can dial a node within the connection gater.

alrevuelta: mm interesting. actually that could be and endless recursive loop? since that dial will be handled by the gater of the other node, that will dial back and will be handled by the gater, that again will dial and will be handled by the gater?

unsure if its the case.

richard-ramos: I confirm that it gets stuck into a recursive loop.  :-1:

waku/v2/node/wakunode2.go
```diff
@@ -253,8 +255,12 @@ func New(opts ...WakuNodeOption) (*WakuNode, error) {
 		w.log.Error("creating localnode", zap.Error(err))
 	}
 
+	metadata := metadata.NewWakuMetadata(w.opts.clusterID, w.localNode, w.log)
```
chaitanyaprem, line 258: Isn't metadata protocol optional based on clusterID?
If so, shouldn't we initialize protocol only if clusterID is specified.

waku/v2/peermanager/topic_event_handler.go
```diff
@@ -118,6 +120,37 @@ func (pm *PeerManager) handlerPeerTopicEvent(peerEvt relay.EvtPeerTopic) {
...
+		if pm.metadata != nil && pm.metadata.ClusterID() != 0 {
```
chaitanyaprem, line 124: Don't think this is the right place for this check.
Because for lightNodes, this topicEvent won't be applicable. This event is generated only for relay nodes where the node advertises that it is interested in pubsubTopic as part of gossipsub Hello packet.

We can refer to how go-libp2p-pubsub is doing something similar for helloPacket as soon as peer is connected. https://github.com/libp2p/go-libp2p-pubsub/blob/d13e24ddc9f288557732c9c02b143c09b86c6778/pubsub.go#L677-L708

They seem to be using connectionNotifier to communicate Hello packet by opening a stream. We could do the same and then wait for response and decide whether to continue or not?

richard-ramos: Good point. I'll move this to a connection notifier.

waku/v2/node/wakunode2.go
richard-ramos: I'll add this logic on the Start() of WakuMetadata. That way wakunode2 does not have to contain logic related to the value of the clusterId

waku/v2/peermanager/topic_event_handler.go
richard-ramos: Done! code looks nicer now as the peer manager no longer has anything to do with the metadata protocol

waku/v2/protocol/metadata/pb/generate.go
```diff
@@ -0,0 +1,3 @@
+package pb
+
+//go:generate protoc -I. --go_opt=paths=source_relative --go_opt=Mwaku_metadata.proto=github.com/waku-org/go-waku/waku/v2/protocol/metadata/pb --go_out=. ./waku_metadata.proto
```
chaitanyaprem, line 3: Would be good to start using proto file from https://github.com/waku-org/waku-proto repo. 
The idea is to not have copies of files and all implementations refer to files from this repo.

waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -0,0 +1,243 @@
...
+
+	u32ClusterID := uint32(wakuM.clusterID)
+
+	return &u32ClusterID, shards, nil
+}
+
+func (wakuM *WakuMetadata) Request(ctx context.Context, peerID peer.ID) (*protocol.RelayShards, error) {
+	logger := wakuM.log.With(logging.HostID("peer", peerID))
+
+	connOpt, err := wakuM.h.NewStream(ctx, peerID, MetadataID_v1)
```
chaitanyaprem, line 95: Suggestion: rename connOpt to stream to make code more readable.

waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -0,0 +1,243 @@
...
+	if err != nil {
+		logger.Error("creating stream to peer", zap.Error(err))
+		return nil, err
+	}
+
+	defer connOpt.Close()
+	defer func() {
+		err := connOpt.Reset()
```
chaitanyaprem, line 103: Did not understand why we have to call both close and reset on the stream?
I was thinking calling Close is enough.

waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -0,0 +1,243 @@
...
+	if wakuM.clusterID == 0 {
+		wakuM.log.Warn("no clusterID is specified. Protocol will not be initialized")
+	}
+
+	ctx, cancel := context.WithCancel(ctx)
+
+	wakuM.ctx = ctx
+	wakuM.cancel = cancel
+
+	wakuM.h.Network().Notify(wakuM)
```
chaitanyaprem, line 67: Does libp2p accept more than 1 connectionNotifier or does it override? 
If it overrides, then the connectedness and connectionGater notifyDisconnect may not work.

richard-ramos: it accepts multiple connection notifiers!

waku/v2/protocol/metadata/pb/generate.go
richard-ramos: Indeed. Let's do that as part of https://github.com/waku-org/go-waku/issues/720  
If you wish I can work on it (currently it is assigned to you)

chaitanyaprem: Alright.Feel free to go ahead and take it up.

waku/v2/protocol/metadata/waku_metadata.go
richard-ramos: Implemented in https://github.com/waku-org/go-waku/pull/805/commits/ad3a17e6d233c4c8a07a2266e38db72453e3adad

waku/v2/protocol/metadata/waku_metadata.go
richard-ramos: hmm.. i'm trying to find out. I went inside the go-libp2p protocols (ping, identify...) and they use both. The diff according to the docs is that Close() only closes the writing side, but Reset() closes both writing and reading

chaitanyaprem: Ok, in that case just doing reset should be enough. Doesn't seem right to use both close and reset then.

richard-ramos: Found this interesting comment: https://github.com/libp2p/specs/issues/166#issuecomment-490017074 
Based on it, I'll use `Close()`

chaitanyaprem (approved):
LGTM

PR #806, opened 2023-10-16
fix: addr update

richard-ramos:
# Description
Simplifies the code for verifying if there was an update in the listening addresses, and also does the verification regardless of the order of the addresses returned by ListenAddresses()

The reason for this change is that I noticed a case when I received a listenAdd update, containing the same addresses as before, but the order was different


chaitanyaprem (approved):
LGTM

Did you face some challenge with Math.Rand and hence moved to crypto rand? 
Would be nice to know the reasons for the switch.

richard-ramos:
> Did you face some challenge with Math.Rand and hence moved to crypto rand?
Would be nice to know the reasons for the switch.

I changed it due to a deprecation warning from the IDE:

https://tip.golang.org/doc/go1.20
> The [math/rand](https://tip.golang.org/pkg/math/rand/) package now automatically seeds the global random number generator (used by top-level functions like Float64 and Int) with a random value, and the top-level [Seed](https://tip.golang.org/pkg/math/rand/#Seed) function has been deprecated. Programs that need a reproducible sequence of random numbers should prefer to allocate their own random source, using rand.New(rand.NewSource(seed)).
Programs that need the earlier consistent global seeding behavior can set GODEBUG=randautoseed=0 in their environment.
The top-level [Read](https://tip.golang.org/pkg/math/rand/#Read) function has been deprecated. In almost all cases, [crypto/rand.Read](https://tip.golang.org/pkg/crypto/rand/#Read) is more appropriate.

Issue #674, opened 2023-08-23
feat: Autosharding API for (relay) subscriptions


chair28980:
Currently the only way to manipulate a relay node's pubsub topic subscriptions is through an explicit subscribe call with the pubsub topic as argument or via static configuration when setting up the node. This task tracks work to add a subscribe method to relevant Relay APIs allowing applications to provide desired content topics as (relay) subscribe arguments, while the node translates those to relay subscriptions for specific shards. This task should involve a mechanism where an application is only notified of new messages on subscribed content topics, instead of all messages on the subscribed shards.

Priority: Critical for launch

- [x] #779 
- [x] #807
- [x] #822

chaitanyaprem:
**Weekly Update**

- _achieved_:  update relay to support new contentTopic based subscriptions, implement new autosharding relay RPC APIs
- _next_: implement new relay REST autoshard API

chaitanyaprem:
Unit tests for REST API is being tracked as part of #861

PR #807, opened 2023-10-16
feat :  autoshard relay api

chaitanyaprem:
# Description

Update Relay API to support autosharding per #674.
Following nwaku https://github.com/waku-org/nwaku/pull/1983/
Some refactoring is done in order to achieve this.
High level idea is to use ContentFilter for subscription and unsubscription. 

# Changes

- [x] Broadcaster is mandatory for relay to work
- [x] Modify relay Subscribe API to take in a contentFilter in order to subscribe to specific pubSubTopic and/or list of contentTopics
- [x] If subscription is for a specific content-Topic, filter messages delivered to subscription as per filter
- [x] Update Unsubscribe API
- [x] Unregister for subscriptions to be handled in broadcaster
- [x] Fix and update examples
- [x] Need to assess impact on ArraySubscriptions and refactor it
- [x] Implement new RPC API's to support contentTopics(Autosharding) based APIs
- [x] Optimize Relay code to not have msgChannel and rather increase buffersize for underlying pubsub subscription itself
- [x] Support more than 1 subscription for a pubSubTopic (uniqueness can be based on subID)
- [x] Test new autoshard relay RPC 
- [x] Add unit tests in relay for new subscription

Note: Added few unit tests, will track a separate PR for unit tests as this PR is becoming too big.

# Tests

- Tested all new relay RPC's by running 2 nodes locally
- Ran all existing unit tests
- Ran service node locally


waku/v2/protocol/relay/waku_relay.go
```diff
@@ -544,7 +612,7 @@ func (w *WakuRelay) topicEventPoll(topic string, handler *pubsub.TopicEventHandl
 	for {
 		evt, err := handler.NextPeerEvent(w.Context())
 		if err != nil {
-			if err == context.Canceled {
+			if err == context.Canceled || err == context.DeadlineExceeded {
```
richard-ramos, line 615:
```suggestion
			if errors.Is(err, context.Canceled )|| errors.Is(err, context.DeadlineExceeded) {
```

waku/v2/protocol/relay/subscription.go
```diff
@@ -1,25 +1,57 @@
...
+	ID            int
+	Unsubscribe   func() //for internal use only. For relay Subscription use relay protocol's unsubscribe
+	Ch            chan *protocol.Envelope
+	contentFilter protocol.ContentFilter
+	subType       SubsciptionType
 }
 
-// NoopSubscription creates a noop subscription that will not receive any envelope
-func NoopSubscription() Subscription {
+type SubsciptionType int
```
richard-ramos, line 17:
```suggestion
type SubscriptionType int
```

waku/v2/protocol/relay/subscription.go
```diff
@@ -1,25 +1,57 @@
...
 }
 
-// ArraySubscription creates a subscription for a list of envelopes
-func ArraySubscription(msgs []*protocol.Envelope) Subscription {
+// SimulateSubscription creates a subscription for a list of envelopes
+// This is used only for internal testing purposes.
+func SimulateSubscription(msgs []*protocol.Envelope) Subscription {
 	ch := make(chan *protocol.Envelope, len(msgs))
 	for _, msg := range msgs {
 		ch <- msg
```
richard-ramos, line 57: Perhaps it's a good idea to move this to a *_test.go file

waku/v2/protocol/relay/broadcast.go
```diff
@@ -62,36 +65,39 @@ func (s *chStore) broadcast(ctx context.Context, m *protocol.Envelope) {
...
-	// send to all registered subscribers
-	for _, ch := range s.topicToChans[""] {
+
+	// send to all wildcard subscribers
+	for _, sub := range s.topicsToSubs[""] {
 		select {
 		case <-ctx.Done():
 			return
-		case ch <- m:
+		case sub.Ch <- m:
```
richard-ramos, line 78: Shouldnt it be `sub.Submit` similar to line 69?

richard-ramos (commented):
PR is looking great so far!

waku/v2/protocol/legacy_filter/waku_filter_test.go
```diff
@@ -46,7 +50,8 @@ func makeWakuFilter(t *testing.T) (*WakuFilter, host.Host) {
 	require.NoError(t, b.Start(context.Background()))
 	filter := NewWakuFilter(b, false, timesource.NewDefaultClock(), prometheus.DefaultRegisterer, utils.Logger())
 	filter.SetHost(host)
-	err = filter.Start(context.Background(), relay.NoopSubscription())
+	sub := relay.NewSubscription(protocol.NewContentFilter(testTopic, testContentTopic))
+	err = filter.Start(context.Background(), *sub)
```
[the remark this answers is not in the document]
chaitanyaprem: Makes sense, it is anyways better to store pointer to subscription rather than a copy.
Will change this.

cmd/waku/server/rpc/relay.go
```diff
@@ -120,28 +125,98 @@ func (r *RelayService) PostV1Message(req *http.Request, args *RelayMessageArgs,
...
+// PostV1AutoSubscription is invoked when the json rpc request uses the post_waku_v2_relay_v1_auto_subscription
+// Note that this method takes contentTopics as an argument instead of pubsubtopics and uses autosharding to derive pubsubTopics.
+func (r *RelayService) PostV1AutoSubscription(req *http.Request, args *TopicsArgs, reply *SuccessReply) error {
+
+	var err error
+	_, err = r.node.Relay().Subscribe(r.node.Relay().Context(), protocol.NewContentFilter("", args.Topics...))
+	if err != nil {
+		r.log.Error("subscribing to topics", zap.Strings("topics", args.Topics), zap.Error(err))
+		return err
+	}
```
richard-ramos, line 137:
```suggestion
	_, err := r.node.Relay().Subscribe(r.node.Relay().Context(), protocol.NewContentFilter("", args.Topics...))
	if err != nil {
		r.log.Error("subscribing to topics", zap.Strings("topics", args.Topics), zap.Error(err))
		return err
	}
```

waku/v2/protocol/relay/waku_relay_test.go
```diff
@@ -160,3 +167,142 @@ func TestMsgID(t *testing.T) {
...
+		ContentTopic: testcTopic1,
+		Timestamp:    0,
+	}
+
+	_, err = relay.PublishToTopic(context.Background(), msg1, subs[0].contentFilter.PubsubTopic)
+	require.NoError(t, err)
+
+	/* 	TODO: Debug why message is not received in this case.
+	wg = waitForMsg(t, subs1[0].Ch, testcTopic1)
+	   	wg.Wait() */
```
richard-ramos, line 281: Should this be solved before merging the PR?

waku/v2/protocol/relay/broadcast.go
```diff
@@ -62,36 +65,40 @@ func (s *chStore) broadcast(ctx context.Context, m *protocol.Envelope) {
...
-	for _, ch := range s.topicToChans[""] {
+
+	// send to all wildcard subscribers
+	for _, sub := range s.topicsToSubs[""] {
 		select {
 		case <-ctx.Done():
 			return
-		case ch <- m:
+		default:
+			sub.Submit(m)
```
richard-ramos, line 79: This `Submit` function can blocking, since ctx.Done() would not really cancel a submission. One way to solve this could be by changing the function so it looks like this:

```go
// Submit allows a message to be submitted for a subscription
func (s *Subscription) Submit(ctx context.Context, msg *protocol.Envelope) {
	//Filter and notify
	// - if contentFilter doesn't have a contentTopic
	// - if contentFilter has contentTopics and it matches with message
	if len(s.contentFilter.ContentTopicsList()) == 0 || (len(s.contentFilter.ContentTopicsList()) > 0 &&
		slices.Contains[string](s.contentFilter.ContentTopicsList(), msg.Message().ContentTopic)) {
                select {
                    case <-ctx.Done():
			return
                    case s.Ch <- msg:
                        //
                }
	}
}
```

Notice the `select` block i added. That way if the context is canceled, the attempt to push the `msg` to `s.Ch` will not stop the execution of `Submit`

waku/v2/protocol/relay/waku_relay.go
```diff
@@ -231,6 +232,9 @@ func (w *WakuRelay) Start(ctx context.Context) error {
 }
 
 func (w *WakuRelay) start() error {
+	if w.bcaster == nil {
+		return fmt.Errorf("broadcaster not specified for relay")
```
richard-ramos, line 236:
```suggestion
		return errors.New("broadcaster not specified for relay")
```

waku/v2/protocol/relay/waku_relay.go
```diff
@@ -385,9 +388,29 @@ func (w *WakuRelay) PublishToTopic(ctx context.Context, message *pb.WakuMessage,
...
+		return nil, err
+	}
+	contentFilter := waku_proto.NewContentFilter(pubSubTopic, contentTopic)
+	cSubs := w.contentSubs[pubSubTopic]
+	for _, sub := range cSubs {
+		if sub.contentFilter.Equals(contentFilter) {
+			return sub, nil
+		}
+	}
+	return nil, fmt.Errorf("no subscription found for content topic")
```
richard-ramos, line 413:
```suggestion
	return nil, errors.New("no subscription found for content topic")
```

waku/v2/protocol/relay/waku_relay.go
```diff
@@ -409,112 +432,145 @@ func (w *WakuRelay) EnoughPeersToPublishToTopic(topic string) bool {
...
 	if err != nil {
 		return nil, err
 	}
 
-	// Create client subscription
-	subscription := NoopSubscription()
-	if w.bcaster != nil {
-		subscription = w.bcaster.Register(topic, 1024)
+	for pubSubTopic, cTopics := range pubSubTopicMap {
+		w.log.Info("subscribing to ", zap.String("pubsubTopic", pubSubTopic), zap.Strings("content-topics", cTopics))
```
richard-ramos, line 445:
```suggestion
		w.log.Info("subscribing to", zap.String("pubsubTopic", pubSubTopic), zap.Strings("contenTopics", cTopics))
```

waku/v2/protocol/relay/waku_relay.go
```diff
@@ -409,112 +432,145 @@ func (w *WakuRelay) EnoughPeersToPublishToTopic(topic string) bool {
...
-	w.log.Info("unsubscribing from topic", zap.String("topic", sub.Topic()))
 
-	w.relaySubs[topic].Cancel()
-	delete(w.relaySubs, topic)
+	for pubSubTopic, cTopics := range pubSubTopicMap {
+		cfTemp := waku_proto.NewContentFilter(pubSubTopic, cTopics...)
+		pubsubUnsubscribe := false
+		sub, ok := w.relaySubs[pubSubTopic]
+		if !ok {
+			return fmt.Errorf("not subscribed to topic")
```
richard-ramos, line 500:
```suggestion
			return errors.New("not subscribed to topic")
```

waku/v2/protocol/relay/waku_relay.go
```diff
@@ -409,112 +432,145 @@ func (w *WakuRelay) EnoughPeersToPublishToTopic(topic string) bool {
...
+				}
+			}
+			if len(cSubs) == 0 {
+				pubsubUnsubscribe = true
+			}
+		} else {
+			//Should not land here ideally
+			w.log.Error("pubsub subscriptions exists, but contentSubscription doesn't for contentFilter",
+				zap.String("pubsubTopic", pubSubTopic), zap.Strings("contentTopics", cTopics))
+			return fmt.Errorf("unexpected error in unsubscribe")
```
richard-ramos, line 518:
```suggestion
			return errors.New("unexpected error in unsubscribe")
```

cmd/waku/node.go
```diff
@@ -382,15 +382,13 @@ func Execute(options NodeOptions) error {
...
-			sub, err := wakuNode.Relay().SubscribeToTopic(ctx, nodeTopic)
+			_, err := wakuNode.Relay().Subscribe(ctx, wprotocol.NewContentFilter(nodeTopic, cTopics...))
 			if err != nil {
 				return err
 			}
 
-			sub.Unsubscribe()
-
 			if len(options.Rendezvous.Nodes) != 0 {
 				// Register the node in rendezvous point
```
richard-ramos, line 393: We're creating a subscription here and not reading it. Wouldn't that mean that once the msg channel buffer is filled, we'd end up with a "subscriber too slow" error? I wonder if we should have an option in `Subscribe` to indicate that we're not interested in the messages themselves and that translates into a 
```
for _ := range mySub.Ch {
   // Do nothing
}
```

chaitanyaprem: Ah, you are right.
That is a good idea, will add that.

waku/v2/protocol/relay/waku_relay_test.go
chaitanyaprem: Thanks for catching this, it needs to be solved before merge.. seem to have missed it :)

chaitanyaprem: This is not an issue, i did not realize i was trying to send same message twice which causes relay to ignore the duplicate and hence it was not returning the message back to the subscription.
Once i modified it to send a different message, it started working. lol

richard-ramos (approved):
:fire:

PR #808, opened 2023-10-16
fix: setFallbackIP with reported libp2p addr

richard-ramos:
# Description
While testing `shards.test` fleet, i noticed that when discv5 endpoint predictor fails, the node's IP address reported in the ENR was 127.0.0.1 regardless of the value being set. With this change, the reported libp2p address will be set as fallback address to be used in case endpoint predictor fails to obtain the IP address


chaitanyaprem (approved):
LGTM

Issue #813, opened 2023-10-17
feat: lightpush REST API


richard-ramos:
OpenAPI file available here https://github.com/waku-org/nwaku/blob/3af0e13d7ed4475654e4b13b3f725b91df6322fe/waku/node/rest/lightpush/openapi.yaml

cc: @harsh-98 if you'd like to work on it?

PR #818, opened 2023-10-19
Rest light push

harsh-98:
# Description
Add Lightpush publish message REST endpoint.

# Changes

- [x] Rest endpoint
- [x] Tests 

# Tests

- [x] Ran new test and prev test

## Issue

#813

### DIFF
--- a/cmd/waku/server/rest/debug.go
+++ b/cmd/waku/server/rest/debug.go
@@ -21,7 +21,7 @@ type InfoReply struct {
 }
 
 const routeDebugInfoV1 = "/debug/v1/info"
-const routeDebugVersionV1 = "/debug/v1/info"
+const routeDebugVersionV1 = "/debug/v1/version"
 
 func NewDebugService(node *node.WakuNode, m *chi.Mux) *DebugService {
 	d := &DebugService{
--- a/cmd/waku/server/rest/relay.go
+++ b/cmd/waku/server/rest/relay.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 
@@ -19,6 +20,9 @@ import (
 
 const routeRelayV1Subscriptions = "/relay/v1/subscriptions"
 const routeRelayV1Messages = "/relay/v1/messages/{topic}"
+
+const routeRelayV1AutoSubscriptions = "/relay/v1/auto/subscriptions"
+const routeRelayV1AutoMessages = "/relay/v1/auto/messages"
 
 // RelayService represents the REST service for WakuRelay
 type RelayService struct {
@@ -49,6 +53,14 @@ func NewRelayService(node *node.WakuNode, m *chi.Mux, cacheCapacity int, log *za
 	m.Delete(routeRelayV1Subscriptions, s.deleteV1Subscriptions)
 	m.Get(routeRelayV1Messages, s.getV1Messages)
 	m.Post(routeRelayV1Messages, s.postV1Message)
+
+	m.Post(routeRelayV1AutoSubscriptions, s.postV1AutoSubscriptions)
+	m.Delete(routeRelayV1AutoSubscriptions, s.deleteV1AutoSubscriptions)
+
+	m.Route(routeRelayV1AutoMessages, func(r chi.Router) {
+		r.Get("/{contentTopic}", s.getV1AutoMessages)
+		r.Post("/", s.postV1AutoMessage)
+	})
 
 	return s
 }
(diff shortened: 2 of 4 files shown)


cmd/waku/server/rest/lightpush_api.yaml
```diff
@@ -0,0 +1,84 @@
...
+          type: number
+        timestamp:
+          type: number
+      required:
+        - payload
+        - contentTopic
+
+    PushRequest:
+      type: object
+      properties:
```
chaitanyaprem, line 78: We need to have a peerID also as an option while using lightpush. If peerID is not provided then automatic selection will happen.

cmd/waku/server/rest/lightpush_api.yaml
```diff
@@ -0,0 +1,84 @@
...
+
+components:
+  schemas:
+    PubsubTopic:
+      type: string
+
+    ContentTopic:
+      type: string
+
+    WakuMessage:
```
chaitanyaprem, line 60: Shouldn't this be reused from a common schema where WakuMessage is defined and referenced? 
Otherwise the same definition would get duplicated across multiple yaml files which would be a nightmare to maintain.

cmd/waku/server/rest/lightpush_api.yaml
```diff
@@ -0,0 +1,84 @@
+openapi: 3.0.3
+info:
+  title: Waku V2 node REST API
+  version: 1.0.0
+  contact:
+    name: VAC Team
```
chaitanyaprem, line 6: Also, we should change this to Waku Team right.
cc @NagyZoltanPeter

cmd/waku/server/rest/lightpush_api.yaml
```diff
@@ -0,0 +1,84 @@
...
+    url: https://forum.vac.dev/
```
chaitanyaprem, line 7: Even URL should point to waku org?

cmd/waku/server/rest/lightpush_rest.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	//
+
+	if serv.node.Lightpush() == nil {
+		w.WriteHeader(http.StatusServiceUnavailable)
+		return
+	}
+
+	_, err := serv.node.Lightpush().Publish(req.Context(), msg.Message, lightpush.WithPubSubTopic(msg.PubSubTopic))
+	writeErrOrResponse(w, err, true)
```
richard-ramos, line 67: Not 100% sure, but I think a response is not expected here (comparing with nim implementation (https://github.com/waku-org/nwaku/blob/master/waku/waku_api/rest/lightpush/handlers.nim). If that's the case, it might be worth to add a function like this (:

```go
if err != nil {
    w.WriteHeader(http.StatusServiceUnavailable)
    w.Write(err.Error())
} else {
    w.WriteHeader(http.StatusOK)
}
````

File: cmd/waku/server/rest/waku_rest.go [hunk 6]
```diff
@@ -36,6 +36,7 @@ func NewWakuRest(node *node.WakuNode, address string, port int, enablePProf bool
 	_ = NewDebugService(node, mux)
 	_ = NewHealthService(node, mux)
 	_ = NewStoreService(node, mux)
+	_ = NewLightpushService(node, mux, log)
```
richard-ramos, line 39:
```suggestion
        if node.LightPush() != nil {
        	_ = NewLightpushService(node, mux, log)
        }
```

cmd/waku/server/rest/lightpush_rest.go
```diff
@@ -0,0 +1,68 @@
...
+	msg := &lightpushRequest{}
+	decoder := json.NewDecoder(req.Body)
+	if err := decoder.Decode(msg); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	defer req.Body.Close()
+
+	if err := msg.Check(); err != nil {
+		writeErrOrResponse(w, err, http.StatusBadRequest)
```
richard-ramos, line 56: Do note that `writeErrOrResponse` adds a `"Content-Type", "application/json; charset=UTF-8"` , while the yaml specifies `text/plain` for the responses. Might be a good moment to refactor the `writeErrOrResponse` functions to accept different headers/statuscodes (maybe via options pattern?)

richard-ramos (commented):
Looks good. 
Left some comments of things that should probably be addressed before merging!

cmd/waku/server/rest/lightpush_rest_test.go
```diff
@@ -0,0 +1,75 @@
...
+	err := node2.Host().Peerstore().AddProtocols(node1.Host().ID(), lightpush.LightPushID_v20beta1)
+	require.NoError(t, err)
+	err = node2.Host().Peerstore().(*wakupeerstore.WakuPeerstoreImpl).SetPubSubTopics(node1.Host().ID(), []string{pubSubTopic})
+	require.NoError(t, err)
+
+	return node1, node2
+}
+
+func TestLightpushMessagev1(t *testing.T) {
+	pubSubTopic := "/waku/2/default-waku/proto"
```
chaitanyaprem, line 47: It would be good to add a test where only contentTopic is specified based on autosharding.

chaitanyaprem (commented):
Looks good.

Left some comments.

cmd/waku/server/rest/lightpush_api.yaml
```diff
@@ -0,0 +1,84 @@
...
+      summary: Request a message relay from a LightPush service provider
+      description: Push a message to be relayed on a PubSub topic.
+      operationId: postMessagesToPubsubTopic
+      tags:
+        - lightpush
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/PushRequest'
```
richard-ramos, line 25: @NagyZoltanPeter shouldnt the `requestyBody` be required?
```suggestion
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/PushRequest'
```

Also, in the schema, what do you think of adding an optional `peerID` so the user can specify which peerID to use to publish the messages, instead of making it random? and if the user does not specify this field, then just use the current behavior.

NagyZoltanPeter: I think so, request body is mandatory.

About peerID, currently it is not defined for the interface.
At least in wakunode, when enabling lightpush client interface the operator must define the target peer.

Please check out this: https://github.com/waku-org/nwaku/issues/2136
This is actually about this to improve.

cmd/waku/server/rest/lightpush_api.yaml
NagyZoltanPeter: This PR is about to address this particular issue as well: https://github.com/waku-org/nwaku/pull/2124
Feel free to comment on it there!

cmd/waku/server/rest/lightpush_api.yaml
chaitanyaprem: Alright, let us have an issue indicating to use the common REST API spec files once they are placed in their own repo. That issue should address such items then.

chaitanyaprem:
> The failing test, is unrelated to this PR. node test is failing due to test timeout of 300 secs.

Maybe just rerun that check to ensure it passes. That will rule out it is nothing to do with this PR.

cmd/waku/server/rest/lightpush_api.yaml
chaitanyaprem: Alright, reviewed the other PR. 
I think i reviewed this one first and hence added this comment.

cmd/waku/server/rest/lightpush_api.yaml
chaitanyaprem: cc @NagyZoltanPeter

richard-ramos (approved):

cmd/waku/server/rest/lightpush_api.yaml
chaitanyaprem: It is fine, you can ignore in this PR.

PR #820, opened 2023-10-20
fix: msg digest matches msg hash

richard-ramos:
# Description
Previously the digest was missing the pubsub topic and meta. With this PR we just use the message hash as digest so it contains these attributes


chaitanyaprem (approved):
LGTM

PR #826, opened 2023-10-23
fix: do not disconnect peers in `TestWakuMetadataRequest`

richard-ramos:
# Description
`TestWakuMetadataRequest` is a test intended to check the request/responses for metadata protocol are correct. As such, since we're interested in having deterministic responses, and not disconnect the peers automatically, in this PR I remove the notification process specifically for this test. Other metadata tests are not affected and will still use the notifier mechanism to disconnect peers


chaitanyaprem (approved):
LGTM

PR #828, opened 2023-10-24
chore: use waku-org/waku-proto repository for protobuffer definitions

### LINKED ISSUE #720: chore: Migrate to use proto files from common proto repo for all Waku protocols

richard-ramos:
# Description
This PR removes the .proto files from the repository and adds [waku-proto
](https://github.com/waku-org/waku-proto) as a submodule to be used to generate the protobuffer bindings in go
# Changes
- [x] legacy filter
- [x] filter
- [x] peer_exchange
- [x] lightpush
- [x] metadata
- [x] store  (There are differences between the spec and the .proto file)
- [x] message (The .proto file is missing the RLN proof fields)

Depends on: https://github.com/waku-org/waku-proto/pull/19

## Issue

closes #720


waku/v2/protocol/pb/waku_message.proto
```diff
@@ -2,6 +2,16 @@ syntax = "proto3";
 
 package pb;
 
+message WakuMessage {
```
chaitanyaprem, line 5: Shouldn't this also come from https://github.com/waku-org/waku-proto/blob/main/waku/message/v1/message.proto to avoid duplication?

Also this seems to be different from the one in proto repo.

waku/v2/protocol/pb/waku_message.proto
```diff
@@ -2,6 +2,16 @@ syntax = "proto3";
...
+    bytes payload = 1;
+    string contentTopic = 2;
+    uint32 version = 3;
+    sint64 timestamp = 10;
+    bytes meta = 11;
+    bool ephemeral = 31;
+    bytes rate_limit_proof = 21;
+}
+
 message RateLimitProof {
```
chaitanyaprem, line 15: And this is no longer required then? 
Otherwise let us add it to common proto repo itself and source these definitions from there as well.

If not as part of this PR, but a subsequent one.

library/Create multiple nodes1
```diff
@@ -0,0 +1,6 @@
+Create multiple nodes
```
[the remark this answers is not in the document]
richard-ramos: lol that's a file i commited by mistake

waku/v2/protocol/filter/pb/filter.pb.go
```diff
@@ -1,8 +1,8 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.31.0
-// 	protoc        v4.23.4
-// source: waku_filter_v2.proto
+// 	protoc        v3.21.12
```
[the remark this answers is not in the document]
richard-ramos: I did not notice I had such an old version installed

cmd/waku/server/rpc/filter.go
```diff
@@ -125,7 +125,11 @@ func (f *FilterService) GetV1Messages(req *http.Request, args *ContentTopicArgs,
 	}
 
 	for i := range f.messages[args.ContentTopic] {
-		*reply = append(*reply, ProtoToRPC(f.messages[args.ContentTopic][i]))
+		rpcMsg, err := ProtoToRPC(f.messages[args.ContentTopic][i])
+		if err != nil {
```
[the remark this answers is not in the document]
richard-ramos: Hm. good point. 
I'll change the code to log the error and message hash, and not return it.
If this error happens is due to the message `rate_limit_proof` bytes being incorrect, and it's probably a good idea to assume that the message is invalid then.

waku/v2/protocol/pb/waku_message.proto
richard-ramos: Forgot to commit this

library/encoding.go
```diff
@@ -52,7 +53,7 @@ func EncodeSymmetric(messageJSON string, symmetricKey string, optionalSigningKey
 		}
 	}
 
-	msg.Version = 1
+	msg.Version = proto.Uint32(1)
```
chaitanyaprem, line 56: Would be good to define 1 and 0 as constants and use them to avoid magic numbers.

waku/v2/protocol/rln/pb/rln.pb.go
```diff
@@ -0,0 +1,202 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.31.0
+// 	protoc        v3.21.12
```
chaitanyaprem, line 4: This file seems to be still using older version of protoc, whereas other files are generated using protoc  v4.24.4.
Better to verify all files once for protoc version used to generate. Did not check each and every file, but randomly verified this one and found out version difference.

waku/v2/protocol/pb/message.pb.go
```diff
@@ -0,0 +1,210 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.31.0
+// 	protoc        v3.21.12
```
chaitanyaprem, line 4: This also needs to be regenerated with latest protoc.

waku/v2/protocol/noise/payload.go
```diff
@@ -5,12 +5,13 @@ import (
...
 	n "github.com/waku-org/go-noise"
 	"github.com/waku-org/go-waku/waku/v2/protocol/pb"
+	"google.golang.org/protobuf/proto"
 )
 
 // DecodePayloadV2 decodes a WakuMessage to a PayloadV2
 // Currently, this is just a wrapper over deserializePayloadV2 and encryption/decryption is done on top (no KeyInfo)
 func DecodePayloadV2(message *pb.WakuMessage) (*n.PayloadV2, error) {
-	if message.Version != 2 {
+	if message.GetVersion() != 2 {
```
chaitanyaprem, line 14: Better to use constant for this. 
I am wondering since message versions are defined in rfc for values 0,1 and 2 , if we should include them as enums in protobuf itself rather than using int.
cc @jm-clius , @alrevuelta

jm-clius: Although a good idea to have the client implementation be specific here, I'm not sure we want to overdefine this in the protobuf. Since this is mostly an application-defined field and only interesting to the application, I don't see much value being gained in changing from `int`.

cmd/waku/server/rpc/relay.go
```diff
@@ -157,10 +161,11 @@ func (r *RelayService) DeleteV1AutoSubscription(req *http.Request, args *TopicsA
 
 // PostV1AutoMessage is invoked when the json rpc request uses the post_waku_v2_relay_v1_auto_message
 func (r *RelayService) PostV1AutoMessage(req *http.Request, args *RelayAutoMessageArgs, reply *SuccessReply) error {
-	var err error
-	msg := args.Message.toProto()
-	if msg == nil {
-		err := fmt.Errorf("invalid message format received")
+	msg, err := args.Message.toProto()
+	if msg == nil || err != nil {
```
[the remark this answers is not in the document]
richard-ramos: This is incorrect. Will fix it

waku/v2/protocol/noise/payload.go
chaitanyaprem: WakuMessage.Version is a field defined in the WakuMessage.proto right. 
Since message versions are specified as 0,1 and 2 in rfc https://rfc.vac.dev/spec/14/#payload-encryption , i was wondering if it makes sense to include it as part of protobuf definition.
Am i missing something, when you say this is implementation specific and the field in application-defined?

jm-clius: Ah "implementation specific" in that it is populated by the application and only of interest to the application and not on the Waku level. Waku (at a routing layer) does not decrypt or encrypt the messages based on this field value, so does not need to parse this field. I'm not sure that the version field should even be defined in the Message RFC, but rather in various higher-layer specs. In a sense this is similar to the `meta` and `content_topic` fields (populated by the application for its own use), except that both of these are used on the Waku layer to determine the message hash. Version field is simply a helper field for the application.

chaitanyaprem (approved):
LGTM 
Can be merged if protoc version is fixed.

PR #836, opened 2023-10-25
fix(filter-subscribe): params.selectedPeer not set

harsh-98:
# Description
Bug found like working on #746. In subscribe if params.selectPeer is missing,  we select peer from peer manager. But the selected peer is not being passed to wf.request.

# Changes

# Tests

- Test are being written as part of FilterRest PR.

### DIFF
--- a/waku/v2/protocol/filter/client.go
+++ b/waku/v2/protocol/filter/client.go
@@ -134,6 +134,11 @@ func (wf *WakuFilterLightNode) onRequest(ctx context.Context) func(network.Strea
 
 		stream.Close()
 
+		if err = messagePush.Validate(); err != nil {
+			logger.Warn("received invalid messagepush")
+			return
+		}
+
 		pubSubTopic := ""
 		//For now returning failure, this will get addressed with autosharding changes for filter.
 		if messagePush.PubsubTopic == nil {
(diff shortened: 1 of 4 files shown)


waku/v2/protocol/filter/client.go
```diff
@@ -306,7 +306,12 @@ func (wf *WakuFilterLightNode) Subscribe(ctx context.Context, contentFilter prot
 		cFilter.PubsubTopic = pubSubTopic
 		cFilter.ContentTopics = protocol.NewContentTopicSet(cTopics...)
 
-		err := wf.request(ctx, params, pb.FilterSubscribeRequest_SUBSCRIBE, cFilter)
+		err := wf.request(
+			ctx,
+			&FilterSubscribeParameters{selectedPeer: selectedPeer, requestID: params.requestID},
```
chaitanyaprem, line 311: Good catch, but rather than creating a new one manually like this, it would be better to copy the existing params and just set the selectedPeer in the new copied version and pass that. 
That way in future if new params are added, we don't need to manually pass them here, rather the copying would take care.

chaitanyaprem: Ok, it's upto you if you want to take it as separate PR or include in this.
But, hope you do get the maintainability problem i see with this.

chaitanyaprem (approved):
LGTM

PR #837, opened 2023-10-25
fix: deadlock in peer connector / manager interaction

richard-ramos:
# Description
Fixes the error seen on this build: 
https://ci.status.im/blue/organizations/jenkins/go-waku%2Fprs%2Frace/detail/PR-828/4/pipeline/

# Changes

- [ ] When creating a subscription to a discovery service, use the context that is provided to jump out of loop
- [ ] When `AddDiscoveredPeer` is invoked, use a goroutine to call `PushToChan`. This is because when calling the PeerConnector `Stop()` function, a write lock is acquired, and `PushToChan` tries to acquire a read lock immediately, and then `Stop()` gets stuck


chaitanyaprem (approved):
Good find.
LGTM

PR #838, opened 2023-10-26
chore: print a message periodically indicating that VACUUM is still being executed

richard-ramos:
# Description
This will print every 2 minutes the following log line 
```
2023-10-25T21:00:22.973-0400    INFO    gowaku.db-setup sqlite/sqlite.go:63     still vacuuming...      {"node": "16Uiu2HAmJikHbVZJhVtbCvXGwCb6sekZKkZcyP9uU7BcaaAeUWoN"}
```

It's dumb, but it's good UX to indicate that something is happening :)

### DIFF
--- a/waku/persistence/postgres/postgres.go
+++ b/waku/persistence/postgres/postgres.go
@@ -1,8 +1,10 @@
 package postgres
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
+	"time"
 
 	"github.com/golang-migrate/migrate/v4/database"
 	"github.com/golang-migrate/migrate/v4/database/pgx"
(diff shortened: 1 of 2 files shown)


waku/persistence/sqlite/sqlite.go
```diff
@@ -32,10 +34,35 @@ func addSqliteURLDefaults(dburl string) string {
 
 func executeVacuum(db *sql.DB, logger *zap.Logger) error {
 	logger.Info("starting sqlite database vacuuming")
-	_, err := db.Exec("VACUUM")
-	if err != nil {
-		return err
+
+	ctx, cancel := context.WithCancel(context.Background())
```
chaitanyaprem, line 38: This code is exact duplicate of the code for postgres. Wondering if we could somehow have this function at a higher layer and call it from both these places? Especially since i don't see any db specific dependencies other than the actual command to execute which can be passed as an argument.

richard-ramos: Good idea. I ended up implementing a function that can be used for any slow query to log a message

yakimant (approved):
LGTM, but I don't know Go much.

waku/persistence/postgres/postgres.go
```diff
@@ -15,10 +17,35 @@ import (
...
+		_, err := db.Exec("VACUUM FULL")
+		if err != nil {
+			errCh <- err
+		}
+	}()
+
+	t := time.NewTicker(2 * time.Minute)
+	defer t.Stop()
+
+loop:
```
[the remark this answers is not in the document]
richard-ramos: the code formatter automatically moved that label to the left. Dunno why tho but seems to be correct?

Issue #839, opened 2023-10-26
feat: Support getting peers by shard  via PeerExchange


harsh-98:
### Problem
Support getting peers by shard  via PeerExchange. 

### Suggested solution
Provide an option in the client API of peerExchange to specify clusterID and shard to filter peers that are retrieved from PeerExchange server.

chaitanyaprem:
Changing the scope of this ticket to just provide a client side API to filter retrived peers by shard. 

If there is a need in the future the PeerExchange protocol can be upgraded to include shard based peer filtering.

chaitanyaprem:
Implemented, hence closing.

PR #840, opened 2023-10-26
refactor: shard LRU for storing peers by shard 

harsh-98:
# Description
Add data structure for storing peers by cluster/index in peerExchange.  Related to https://github.com/waku-org/go-waku/issues/839
- Addition of node takes time complexity  of O(count of prev index+ new indexes).

# Tests

- [x] add  test for functionality of the sharded cache

## How to test

- [x] make sure all the tests are passing

### DIFF
--- a/waku/v2/protocol/peer_exchange/enr_cache.go
+++ b/waku/v2/protocol/peer_exchange/enr_cache.go
@@ -3,7 +3,6 @@ package peer_exchange
 import (
 	"bufio"
 	"bytes"
-	"math/rand"
 	"sync"
 
 	"github.com/ethereum/go-ethereum/p2p/enode"
@@ -16,7 +15,6 @@ import (
 type enrCache struct {
 	// using lru, saves us from periodically cleaning the cache to mauintain a certain size
 	data *shardLRU
-	rng  *rand.Rand
 	mu   sync.RWMutex
 	log  *zap.Logger
 }
@@ -26,8 +24,6 @@ func newEnrCache(size int, log *zap.Logger) *enrCache {
 	inner := newShardLRU(int(size))
 	return &enrCache{
 		data: inner,
-		rng:  rand.New(rand.NewSource(rand.Int63())),
-		log:  log.Named("enr-cache"),
 	}
 }
 
@@ -44,22 +40,12 @@ func (c *enrCache) updateCache(node *enode.Node) {
 
 // get `numPeers` records of enr
 func (c *enrCache) getENRs(neededPeers int, clusterIndex *ShardInfo) ([]*pb.PeerInfo, error) {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
+	c.mu.Lock()
+	defer c.mu.Unlock()
 	//
-	availablePeers := c.data.len(clusterIndex)
-	if availablePeers == 0 {
-		return nil, nil
-	}
-	if availablePeers < neededPeers {
-		neededPeers = availablePeers
-	}
-
-	perm := c.rng.Perm(availablePeers)[0:neededPeers]
-	nodes := c.data.getNodes(clusterIndex)
+	nodes := c.data.getRandomNodes(clusterIndex, neededPeers)
 	result := []*pb.PeerInfo{}
-	for _, ind := range perm {
-		node := nodes[ind]
+	for _, node := range nodes {
 		//
 		var b bytes.Buffer
 		writer := bufio.NewWriter(&b)
--- a/waku/v2/protocol/peer_exchange/enr_cache.go
+++ b/waku/v2/protocol/peer_exchange/enr_cache.go
@@ -3,7 +3,6 @@ package peer_exchange
 import (
 	"bufio"
 	"bytes"
-	"sync"
 
 	"github.com/ethereum/go-ethereum/p2p/enode"
 
@@ -15,7 +14,6 @@ import (
 type enrCache struct {
 	// using lru, saves us from periodically cleaning the cache to mauintain a certain size
 	data *shardLRU
-	mu   sync.RWMutex
 	log  *zap.Logger
 }
 
@@ -29,8 +27,6 @@ func newEnrCache(size int, log *zap.Logger) *enrCache {
 
 // updating cache
 func (c *enrCache) updateCache(node *enode.Node) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
 	currNode := c.data.Get(node.ID())
 	if currNode == nil || node.Seq() > currNode.Seq() {
 		c.data.Add(node)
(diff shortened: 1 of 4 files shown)


waku/v2/protocol/peer_exchange/enr_cache.go
```diff
@@ -7,43 +7,43 @@ import (
 	"sync"
 
 	"github.com/ethereum/go-ethereum/p2p/enode"
-	"github.com/hashicorp/golang-lru/simplelru"
+
 	"github.com/waku-org/go-waku/waku/v2/protocol/peer_exchange/pb"
 )
 
-// simpleLRU internal uses container/list, which is ring buffer(double linked list)
+// simpleLRU uint16ernal uses container/list, which is ring buffer(double linked list)
```
richard-ramos, line 14:
```suggestion
// simpleLRU internal uses container/list, which is ring buffer(double linked list)
```

waku/v2/protocol/peer_exchange/enr_cache.go
```diff
@@ -7,43 +7,43 @@ import (
...
 type enrCache struct {
-	// using lru, saves us from periodically cleaning the cache to maintain a certain size
-	data *simplelru.LRU
+	// using lru, saves us from periodically cleaning the cache to mauint16ain a certain size
```
richard-ramos, line 16:
```suggestion
	// using lru, saves us from periodically cleaning the cache to maintain a certain size
```

waku/v2/protocol/peer_exchange/shard_lru.go
```diff
@@ -0,0 +1,137 @@
...
+		return nodes
+	}
+	//
+	if entries := l.shardNodes[*clusterIndex]; entries != nil && entries.Len() != 0 {
+		nodes := make([]*enode.Node, 0, entries.Len())
+		for ent := entries.Back(); ent != nil; ent = ent.Prev() {
+			nodes = append(nodes, ent.Value.(nodeWithClusterIndex).node)
+		}
+		return nodes
+	}
```
chaitanyaprem, line 115: Shouldn't the nodes retrieved be moved to the front so that they are not Least used anymore?

waku/v2/protocol/peer_exchange/shard_lru.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/waku-org/go-waku/waku/v2/protocol"
+	wenr "github.com/waku-org/go-waku/waku/v2/protocol/enr"
+)
+
+type ClusterIndex struct {
+	cluster uint16
+	index   uint16
+}
+type shardLRU struct {
```
chaitanyaprem, line 15: Rather than implementing an LRU ourselves, doesn't it make sense to use an existing LRU implementation to store nodes per shard?

waku/v2/protocol/peer_exchange/shard_lru.go
```diff
@@ -0,0 +1,137 @@
...
+
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/waku-org/go-waku/waku/v2/protocol"
+	wenr "github.com/waku-org/go-waku/waku/v2/protocol/enr"
+)
+
+type ClusterIndex struct {
+	cluster uint16
+	index   uint16
+}
```
richard-ramos, line 14:
```suggestion
type shard struct {
	clusterID uint16
	shardID   uint16
}
```

waku/v2/protocol/peer_exchange/shard_lru.go
```diff
@@ -0,0 +1,137 @@
...
+		idToNode:   map[enode.ID][]*list.Element{},
+		shardNodes: map[ClusterIndex]*list.List{},
+		size:       size,
+	}
+}
+
+type nodeWithClusterIndex struct {
+	key  ClusterIndex
+	node *enode.Node
+}
```
richard-ramos, line 32:
```suggestion
type nodeWithShard struct {
	key  Shard
	node *enode.Node
}
```

waku/v2/protocol/peer_exchange/shard_lru.go
chaitanyaprem: That i understand, since this is an LRU which means we need to keep track of which nodes are retrieved from the cache so that when it overflows only the least used gets removed. 
Without that, we cannot call it an LRU. Because we will just be overriding the oldest inserted entry.

waku/v2/protocol/peer_exchange/shard_lru.go
chaitanyaprem: Got it. 
My point is that if we can avoid spending time on implementing DS that are available and tested, as such code would be difficult to maintain.

waku/v2/protocol/peer_exchange/shard_lru.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+	} else if entries := l.shardNodes[*clusterIndex]; entries != nil && entries.Len() != 0 {
+		elements = make([]*list.Element, 0, entries.Len())
+		for ent := entries.Back(); ent != nil; ent = ent.Prev() {
+			elements = append(elements, ent)
+		}
+	}
+	utils.Logger().Info(fmt.Sprintf("%d", len(elements)))
+	indexes := l.rng.Perm(len(elements))[0:neededPeers]
+	for _, ind := range indexes {
```
chaitanyaprem, line 136: Optimization and can be taken up later: I am wondering if rather than deleting and adding nodes everytime there is a query to keep an indication of recently used nodes, would it be better to just update last fetched timestamp for node that has been returned. 

Only when cache is full and new nodes are to be added, then this timestamp can be used to determine which is least used and remove those nodes. 

Reason being, for a peer that is acting as a peer-exchange node, the number of queries to get nodes may be more than nodes getting added dynamically. But, this can always be relooked at a later stage based on node-churn and behavior in the network.

chaitanyaprem (approved):
LGTM

Few comments which can be addressed later.

waku/v2/protocol/peer_exchange/shard_lru.go
```diff
@@ -0,0 +1,167 @@
...
+		}
+	}
+	if len(l.idToNode[nodeID]) == 0 {
+		delete(l.idToNode, nodeID)
+	}
+}
+
+func nodeToRelayShard(node *enode.Node) *protocol.RelayShards {
+	shard, err := wenr.RelaySharding(node.Record())
+	if shard == nil || err != nil { // if no shard info, then add to node to Cluster 0, Index 0
```
richard-ramos, line 67: I think this condition is incorrect. We should probably return the error. Looking quickly at the implementation of `RelaySharding` if there's an error, it means that the `enode` contained incorrect information, so I'd probably ignore this peer

richard-ramos (approved):
Approving in the meantime but I think the `err` condition extracting the shards from the enr should be fixed.

waku/v2/protocol/peer_exchange/shard_lru.go
chaitanyaprem: I think this was added to handle defaultpububTopic?

richard-ramos: In that case shard would be nil. I'd do something like this, let me know what you think and I'll commit it:

```
func nodeToRelayShard(node *enode.Node) (*protocol.RelayShards, error) {
	shard, err := wenr.RelaySharding(node.Record())
        if err != nil {
           return nil, err
       }

	if shard == nil { // if no shard info, then add to node to Cluster 0, Index 0
		shard = &protocol.RelayShards{
			ClusterID: 0,
			ShardIDs:  []uint16{0},
		}
	}

	return shard, nil
}
```

PR #841, opened 2023-10-26
refactor: validate protobuffer for store

richard-ramos:
# Description
Validates the content of the protobuffers used in store protocol

### DIFF
--- a/waku/v2/protocol/store/waku_store_protocol.go
+++ b/waku/v2/protocol/store/waku_store_protocol.go
@@ -2,6 +2,7 @@ package store
 
 import (
 	"context"
+	"encoding/hex"
 	"errors"
 	"math"
 	"sync"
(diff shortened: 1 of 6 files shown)


waku/v2/protocol/store/waku_store_protocol.go
```diff
@@ -323,20 +349,31 @@ func (store *WakuStore) Resume(ctx context.Context, pubsubTopic string, peerList
...
+			r.err = errors.New("invalid cursor")
+		}
+
+		if r.err != nil {
+			store.log.Warn("could not resume message history", zap.Error(r.err), logging.HostID("peer", r.peerID))
+			continue
+		}
+
+		for _, msg := range r.response.Messages {
+			if err = store.storeMessage(protocol.NewEnvelope(msg, store.timesource.Now().UnixNano(), pubsubTopic)); err == nil {
```
chaitanyaprem, line 370: Here duplicate messages are also counted as query is sent to multiple peers. 
Or does storeMessage return's error in case msg is already stored.

chaitanyaprem (approved):
LGTM
Few comments.

waku/v2/protocol/store/waku_store_protocol.go
richard-ramos: It returns an error

Issue #842, opened 2023-10-26
feat: Deprecate Named Sharding and Update Lightpush Client API


chair28980:
In the [context of discussions](https://discord.com/channels/1110799176264056863/1166349141102837882/1166670331151388784) around named sharding and peer management for the Waku network, it is proposed to make the following changes to the Waku Network:

1. **Deprecate Support for Named Sharding:**

- It is recommended to remove support for named sharding.
- Peer management should ensure that only peers supporting sharded topics are connected.

2. **Lightpush Protocol Updates:**

- The lightpush protocol itself should never be aware of the sharded format of pubsub topics. It is an extension of libp2p's gossipsub.
- On a protocol level, the lightpush service node needs to handle requests on unsupported pubsub topics by returning a clear error code.

3. **Lightpush Client API Changes:**

- To maintain adherence to the sharded format, the lightpush client API should be updated.
- New proposed format: lightPublish(WakuMessage, ClusterID, ShardNumber)
- For legacy requests on named topics, the API could use a trial and error process to send the request to random service nodes. This can be done using the format: lightPublish(WakuMessage, PubsubTopic)
- A request format that bypasses our peer management/selection: lightPublish(WakuMessage, PubsubTopic, ServiceNodeMultiaddr)
---
- Users of custom pubsub topics should select a cluster id.
- Use static sharding until the launch of The Waku Network.
- Default Pubsub topic support remains for now, to be phased out as a later step

### Acceptance Criteria:

- [x]  Remove support for named sharding (except for default pubsub topic)
- [ ]  Update lightpush protocol to handle unsupported pubsub topics with clear error codes.
- [ ]  Introduce the new lightpush client API as mentioned above.
- [ ]  Update documentation to reflect these changes and guide users accordingly.

chair28980:
@harsh-98 good question, I'm not sure, @fryorcraken / @jm-clius your thoughts on this?

fryorcraken:
> @chair28980 regarding `Peer management should ensure that only peers supporting sharded topics are connected.`. This also includes default pubsub topic right?

Cc @jm-clius @SionoiS on this

SionoiS:
It should IMO but it's going to break so much stuff... Necessary pain I guess.

If we don't remove the default pubsub topic then what's the point of dropping support for named shard?

fryorcraken:
IMO default pubsub topic should be the only supported named shard moving forward.
[...]

PR #844, opened 2023-10-27
feat: remove named topic

harsh-98:
# Description
Named pubsubTopic is being deprecated. Only default and shared pubsubTopic will be supported.  In reference to https://github.com/waku-org/go-waku/issues/842

# Changes

- [x] update go-waku examples.
- [x] Remove PubsubTopic() from mobile, c and node libraries. 
- [x] Only support Default pubsub and sharded pubsubTopic.

# Tests

- [x] update pubSubTopic test.

### DIFF
--- a/waku/persistence/postgres/postgres.go
+++ b/waku/persistence/postgres/postgres.go
@@ -1,8 +1,10 @@
 package postgres
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
+	"time"
 
 	"github.com/golang-migrate/migrate/v4/database"
 	"github.com/golang-migrate/migrate/v4/database/pgx"
@@ -15,10 +17,35 @@ import (
 
 func executeVacuum(db *sql.DB, logger *zap.Logger) error {
 	logger.Info("starting PostgreSQL database vacuuming")
-	_, err := db.Exec("VACUUM FULL")
-	if err != nil {
-		return err
+
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	errCh := make(chan error)
+
+	go func() {
+		defer cancel()
+		_, err := db.Exec("VACUUM FULL")
+		if err != nil {
+			errCh <- err
+		}
+	}()
+
+	t := time.NewTicker(2 * time.Minute)
+	defer t.Stop()
+
+loop:
+	for {
+		select {
+		case <-ctx.Done():
+			break loop
+		case err := <-errCh:
+			return err
+		case <-t.C:
+			logger.Info("still vacuuming...")
+		}
 	}
+
 	logger.Info("finished PostgreSQL database vacuuming")
 	return nil
 }
--- a/waku/persistence/sqlite/sqlite.go
+++ b/waku/persistence/sqlite/sqlite.go
@@ -1,9 +1,11 @@
 package sqlite
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/golang-migrate/migrate/v4/database"
 	"github.com/golang-migrate/migrate/v4/database/sqlite3"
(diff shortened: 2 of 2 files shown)


waku/v2/protocol/pubsub_topic.go
```diff
@@ -168,31 +102,29 @@ func (s *StaticShardingPubsubTopic) Parse(topic string) error {
...
 	return nil
 }
 
-// ToShardedPubsubTopic takes a pubSub topic string and creates a NamespacedPubsubTopic object.
-func ToShardedPubsubTopic(topic string) (NamespacedPubsubTopic, error) {
+func ToShardPubsubTopic(topic WakuPubSubTopic) (StaticShardingPubsubTopic, error) {
+	if topic.String() != defaultPubsubTopic {
+		return topic.(StaticShardingPubsubTopic), nil
+	}
+	return StaticShardingPubsubTopic{}, ErrNotShardPubsubTopic
```
richard-ramos, line 113:
```suggestion
        result, ok := topic.(StaticShardingPubsubTopic)
        if !ok {
             return StaticShardingPubsubTopic{}, ErrNotShardPubsubTopic
        }
        return result, nil
```

library/mobile/api.go
```diff
@@ -78,14 +78,9 @@ func ContentTopic(applicationName string, applicationVersion int, contentTopicNa
...
 
-// PubsubTopic creates a pubsub topic string according to RFC 23
-func PubsubTopic(name string, encoding string) string {
-	return protocol.NewNamedShardingPubsubTopic(name + "/" + encoding).String()
-}
-
 // DefaultPubsubTopic returns the default pubsub topic used in waku2: /waku/2/default-waku/proto
 func DefaultPubsubTopic() string {
-	return protocol.DefaultPubsubTopic().String()
+	return protocol.DefaultPubsubTopic{}.String()
```
richard-ramos, line 83:
```suggestion
	return library.DefaultPubsubTopic()
```

richard-ramos (commented):
:rocket:

richard-ramos:
I'm wondering if we should we also replace the topic in by a `WakuPubsubTopic` in 
- https://github.com/waku-org/go-waku/blob/8e549f7049037f4ea7778695aa78b2926e1bc2f8/waku/v2/protocol/relay/waku_relay.go#L227
- https://github.com/waku-org/go-waku/blob/8e549f7049037f4ea7778695aa78b2926e1bc2f8/waku/v2/protocol/lightpush/waku_lightpush_option.go#L43
- as well as wherever a pubsub topic is used in filter and store protocol?

cc: @chaitanyaprem

waku/v2/protocol/pubsub_topic.go
```diff
@@ -7,104 +7,43 @@ import (
 	"strings"
 )
 
-// Waku2PubsubTopicPrefix is the expected prefix to be used for pubsub topics
-const Waku2PubsubTopicPrefix = "/waku/2"
```
chaitanyaprem, line 11: I think it is better to leave this prefix as is and append either default pubsubTopic suffix or sharded topic suffix.
That way, in future if there is a change to this prefix it would be easier to maintian.

library/c/api.go
```diff
@@ -203,14 +203,6 @@ func waku_content_topic(applicationName *C.char, applicationVersion C.uint, cont
 	return execOkCB(onOkCb, contentTopic.String())
 }
 
-// Create a pubsub topic string according to RFC 23
-//
-//export waku_pubsub_topic
-func waku_pubsub_topic(name *C.char, encoding *C.char, onOkCb C.WakuCallBack) C.int {
```
chaitanyaprem, line 209: I am wondering if we should have an API to create sharded pubsub topic based on cluster and shard.
WDYT @richard-ramos 
Atleast until all API's are migrated to only take either contentTopic or cluster/shard.

waku/v2/protocol/content_topic.go
```diff
@@ -7,7 +7,6 @@ import (
 	"strings"
 )
 
-// DefaultContentTopic is the default content topic used in Waku network if no content topic is specified.
 const DefaultContentTopic = "/waku/2/default-content/proto"
```
chaitanyaprem, line 10: I don't think this is used or required anymore other than in test code. 
We can move this to test code

chaitanyaprem:
> I'm wondering if we should we also replace the `topic string` by a `WakuPubsubTopic` in
> 
>     * https://github.com/waku-org/go-waku/blob/8e549f7049037f4ea7778695aa78b2926e1bc2f8/waku/v2/protocol/relay/waku_relay.go#L227
> 
>     * https://github.com/waku-org/go-waku/blob/8e549f7049037f4ea7778695aa78b2926e1bc2f8/waku/v2/protocol/lightpush/waku_lightpush_option.go#L43
> 
>     * as well as wherever a pubsub topic is used in filter and store protocol?
> 
> 
> cc: @chaitanyaprem
> 
> **EDIT**: Hmm... thinking more on this, I think we should change the protocols to `WakuPubsubTopic` wherever strings are used, but do it later, because if we do that now, we'd be automatically dropping support to named pubsub topics and this would mean that TheGraph wouldn't be able to use go-waku without changing their pubsub topic structure. We should keep support for string pubsub topics for now

I remember Hanno mentioning that we are going to still have legacy API's so that existing users like TheGraph are not affected. Instead of modifying existing API's, we should be having new API's with deprecation note mentioned for existing API's that take pubsubTopic string.

library/c/api.go
richard-ramos: Yeah, sounds good.
@harsh-98 this can be done as a separate PR, i think, to get this one merged asap :)

richard-ramos (approved):

PR #845, opened 2023-10-27
refactor: publish API for relay and lightpush

richard-ramos:
# Description
I noticed that the API for Lightpush and Relay were somewhat different and would really benefit from having a similar API.

# Changes

- [ ] Removed `PublishToTopic` from Lightpush and Relay to use only `Publish` with options
- [ ] Updated docs for both protocols
- [ ] Updated examples to use new API
- [ ] Added `examples/noise` to the `build-example` target

Probably most of this work will go away once named shard support is removed, but whatever :D at least we wont have two separate Publish functions anymore

### DIFF
--- a/docs/api/relay.md
+++ b/docs/api/relay.md
@@ -41,34 +41,19 @@ One of these options must be specified when instantiating a node supporting the 
 ## Receiving messages
 ```go
 ...
-sub, err := wakuNode.Relay().Subscribe(context.Background())
+contentFilter := protocol.NewContentFilter(relay.DefaultWakuTopic)
+sub, err := wakuNode.Relay().Subscribe(context.Background, contentFilter) ([]*Subscription, error)
 if err != nil {
     fmt.Println(err)
     return
 }
 
-for value := range sub.C {
+for value := range sub[0].C {
     fmt.Println("Received msg:", string(value.Message().Payload))
 }
 ...
 ```
-To receive messages sent via the relay protocol, you need to subscribe to a pubsub topic. This can be done via any of these functions:
-- `wakuNode.Relay().Subscribe(ctx)` - subscribes to the default waku pubsub topic `/waku/2/default-waku/proto`
-- `wakuNode.Relay().SubscribeToTopic(ctx, topic)` - subscribes to a custom pubsub topic
-
-These functions return a `Subscription` struct containing a channel on which messages will be received. To stop receiving messages in this channel `sub.Unsubscribe()` can be executed which will close the channel (without unsubscribing from the pubsub topic)
-
-> Pubsub topics should follow the [recommended usage](https://rfc.vac.dev/spec/23/) structure. For this purpose, the `NewPubsubTopic` helper function was created:
-```go
-import 	"github.com/waku-org/go-waku/waku/v2/protocol"
-
-topic := protocol.NewPubsubTopic("the_topic_name", "the_encoding")
-/*
-fmt.Println(topic.String())  // => `/waku/2/the_topic_name/the_encoding`
-*/
-```
-
-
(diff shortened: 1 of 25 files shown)


richard-ramos:
Opened also this issue due to noticing an incorrect behavior in the examples: https://github.com/waku-org/go-waku/issues/846

docs/api/relay.md
```diff
@@ -41,34 +41,19 @@ One of these options must be specified when instantiating a node supporting the
...
-import 	"github.com/waku-org/go-waku/waku/v2/protocol"
-
-topic := protocol.NewPubsubTopic("the_topic_name", "the_encoding")
-/*
-fmt.Println(topic.String())  // => `/waku/2/the_topic_name/the_encoding`
-*/
-```
-
-
+To receive messages sent via the relay protocol, you need to subscribe specifying a content filter with the function `Subscribe(ctx context.Context, contentFilter waku_proto.ContentFilter, opts ...RelaySubscribeOption) ([]*Subscription, error)`. This functions return a list of `Subscription` struct containing a channel on which messages will be received. To stop receiving messages in this channel `sub.Unsubscribe()` can be executed which will close the channel (without unsubscribing from the pubsub topic)
```
chaitanyaprem, line 56: We should ask users to use relay.ubsubscribe rather than sub.Unsubscribe. 
This will make sure subscription is stopped i.e messages not sent to subscription, and if no other subscriptions exist for underlying pubsub topic, the pubsub is also unsubscribed.

docs/api/relay.md
```diff
@@ -95,11 +80,13 @@ if err != nil {
 To send a message, it needs to be wrapped into a [`WakuMessage`](https://rfc.vac.dev/spec/14/) protobuffer. The payload of the message is not limited to strings. Any kind of data that can be serialized
 into a `[]byte` can be sent as long as it does not exceed the maximum length a message can have (~1MB)
 
-The following functions can be used to publish a message:
-- `wakuNode.Relay().Publish(ctx, msg)` - to send a message to the default waku pubsub topic
-- `wakuNode.Relay().PublishToTopic(ctx, msg, topic)` - to send a message to a custom pubsub topic
+`wakuNode.Relay().Publish(ctx, msg, opts...)` is used to publish a message. This function will return a message id on success, or an error if the message could not be published.
 
-Both of these functions will return a message id on success, or an error if the message could not be published.
+If no options are specified, go-waku will automatically choose the peer used to broadcast the message via Lightpush and publish the message to a pubsub topic derived from the content topic of the message. This behaviour can be controlled via options:
```
chaitanyaprem, line 85: This is relay API right, need to update to relay.
Similarly the options below also have to be updated with relay.

examples/filter2/main.go
```diff
@@ -98,6 +99,7 @@ func main() {
 
 	// Send FilterRequest from light node to full node
 	cf := protocol.ContentFilter{
+		PubsubTopic:   relay.DefaultWakuTopic,
```
chaitanyaprem, line 102: I think we should keep examples to use autosharding API, hence i remember changing it to not use default pubsubTopic.

waku/v2/protocol/lightpush/waku_lightpush_option.go
```diff
@@ -56,6 +51,20 @@ func WithFastestPeerSelection(fromThesePeers ...peer.ID) Option {
...
 }
 
+// WithPubSubTopic is used to specify the pubsub topic on which a WakuMessage will be broadcasted
+func WithPubSubTopic(pubsubTopic string) Option {
+	return func(params *lightPushParameters) {
+		params.pubsubTopic = pubsubTopic
+	}
+}
+
+// WithPubSubTopic is used to indicate that the message should be broadcasted in the default pubsub topic
```
chaitanyaprem, line 61: WithPubSubTopic -> WithDefaultPubsubTopic

examples/filter2/main.go
chaitanyaprem: Ignore it, just saw your comment regarding separate issue to track this.

chaitanyaprem (approved):
Looks Good

Few minor comments to be addressed.

PR #853, opened 2023-10-31
fix(subscription-map): uniform operations and encapsulation

harsh-98:
# Description
Refactor before adding getMessage to filter rest module. Related to https://github.com/waku-org/go-waku/issues/746. 

# Changes

- [x] Don't expose items from subscriptionMap.
- [x] In filter unsubscribe/unsubscribeAll, return list of err instead of chan. Reason: length of channel not known.
- [x] maintain a map of pub/contentTopic to number of subscription. This way we can check if there is atleast one subscription for given pair of pubsub/contentTopic. add handle updation of this map on deleteion/creation of subscription. `Note`: new contentTopics can be added/removed from subscription.
 

# Tests

- [x] All tests passing.

### DIFF
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -454,8 +454,12 @@ func (pm *PeerManager) SelectRandomPeer(criteria PeerSelectionCriteria) (peer.ID
 	//  - which topics they track
 	//  - latency?
 
-	if peerID := pm.selectServicePeer(criteria.Proto, criteria.PubsubTopic, criteria.SpecificPeers...); peerID != nil {
-		return *peerID, nil
+	peerID, err := pm.selectServicePeer(criteria.Proto, criteria.PubsubTopic, criteria.SpecificPeers...)
+	if err == nil {
+		return peerID, nil
+	} else if !errors.Is(err, ErrNoPeersAvailable) {
+		pm.logger.Debug("could not retrieve random peer from slot", zap.String("protocol", string(criteria.Proto)), zap.String("pubsubTopic", criteria.PubsubTopic), zap.Error(err))
+		return "", err
 	}
 
 	// if not found in serviceSlots or proto == WakuRelayIDv200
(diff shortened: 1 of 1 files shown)


waku/v2/protocol/filter/client.go
```diff
@@ -46,11 +47,25 @@ type WakuFilterLightNode struct {
...
-type WakuFilterPushResult struct {
+type WakuFilterPushError struct {
 	Err    error
 	PeerID peer.ID
 }
 
+type WakuFilterPushResult struct {
+	errs []WakuFilterPushError
+	mu   sync.Mutex
+}
```
richard-ramos, line 58: You could convert the WakuFilterPushResult to a mutex:
```suggestion
type WakuFilterPushResult struct {
        sync.RWMutex
	errs []WakuFilterPushError   
}
```

waku/v2/protocol/filter/client.go
```diff
@@ -46,11 +47,25 @@ type WakuFilterLightNode struct {
...
+
+func (arr *WakuFilterPushResult) Add(err WakuFilterPushError) {
+	arr.mu.Lock()
+	defer arr.mu.Unlock()
+	arr.errs = append(arr.errs, err)
+}
+func (arr *WakuFilterPushResult) Errors() []WakuFilterPushError {
+	return arr.errs
```
richard-ramos, line 66:
```suggestion
        arr.RLock()
        defer arr.RUnlock()
	return arr.errs
```

waku/v2/protocol/subscription/subscriptions_map.go
```diff
@@ -234,21 +193,24 @@ func iterateSubscriptionSet(logger *zap.Logger, subscriptions SubscriptionSet, e
...
 
-	result := resultType{
-		PeerID:      s.PeerID.Pretty(),
-		PubsubTopic: s.ContentFilter.PubsubTopic,
-	}
+	var output []*SubscriptionDetails
 
-	for c := range s.ContentFilter.ContentTopics {
-		result.ContentTopics = append(result.ContentTopics, c)
+	var peersId []peer.ID
```
richard-ramos, line 202:
```suggestion
	var peerIDs []peer.ID
```

waku/v2/protocol/subscription/subscription_details.go
```diff
@@ -0,0 +1,122 @@
...
+	s.Lock()
+	defer s.Unlock()
+
+	for _, ct := range contentTopics {
+		if _, ok := s.ContentFilter.ContentTopics[ct]; !ok {
+			s.ContentFilter.ContentTopics[ct] = struct{}{}
+			// Increase the number of subscriptions for this (pubsubTopic, contentTopic) pair
+			s.mapRef.Lock()
+			s.mapRef.increaseSubFor(s.ContentFilter.PubsubTopic, ct)
+			s.mapRef.Unlock()
```
richard-ramos, line 48: I wonder if we shouldn't instead add a public `IncreaseSubFor` in `SubscriptionMap` that has this definition:
```go
func (m *SubscriptionsMap) IncreaseSubFor(pubsubTopic, contentTopic string) {
    m.Lock()
    defer m.Unlock()
    m.increaseSubFor(pubsubTopic, ct)
} 
```
That way, `SubscriptionDetails` does not need to know that `SubscriptionMap` has a mutex, but would simply call the new function. The same applies for `decreaseSubFor`

richard-ramos (approved):

PR #854, opened 2023-10-31
feat: bridge relay topics

richard-ramos:
# Description
On today's message reliability call, creating a bridge between pubsub topics was suggested as a way to solve backward compatibility due to the breaking change introduced in https://github.com/status-im/status-go/pull/4161.

This PR adds a hidden `--bridge-topic=fromTopic:toTopic` flag that can be used to bridge messages from a pubsub topic to another. This is meant to be used with `status.prod` to get messages going from between the default shard and the default pubsub topic (since mixing named and static sharding introduces a lot of problems).

go-waku nodes must be configured with the following flags: 
```
--bridge-topic=/waku/2/default-waku/proto:/waku/2/rs/16/32 --bridge-topic=/waku/2/rs/16/32:/waku/2/default-waku/proto
```
(Maybe we can reuse `gowaku.prod` fleet)

### DIFF
--- a/cmd/waku/relay.go
+++ b/cmd/waku/relay.go
@@ -1,0 +1,156 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"sync"
+	"time"
+
+	"github.com/waku-org/go-waku/logging"
+	"github.com/waku-org/go-waku/waku/v2/node"
+	wprotocol "github.com/waku-org/go-waku/waku/v2/protocol"
+	"github.com/waku-org/go-waku/waku/v2/protocol/relay"
+	"github.com/waku-org/go-waku/waku/v2/rendezvous"
+	"github.com/waku-org/go-waku/waku/v2/utils"
+	"go.uber.org/zap"
+	"golang.org/x/exp/maps"
+)
+
+var fwdMetaTag = []byte{102, 119, 100} //"fwd"
+
+func handleRelayTopics(ctx context.Context, wg *sync.WaitGroup, wakuNode *node.WakuNode, pubSubTopicMap map[string][]string) error {
+	for nodeTopic, cTopics := range pubSubTopicMap {
+		nodeTopic := nodeTopic
+		_, err := wakuNode.Relay().Subscribe(ctx, wprotocol.NewContentFilter(nodeTopic, cTopics...), relay.WithoutConsumer())
+		if err != nil {
+			return err
+		}
+
+		if len(options.Rendezvous.Nodes) != 0 {
+			// Register the node in rendezvous point
+			iter := rendezvous.NewRendezvousPointIterator(options.Rendezvous.Nodes)
+
+			wg.Add(1)
+			go func(nodeTopic string) {
+				t := time.NewTicker(rendezvous.RegisterDefaultTTL)
+				defer t.Stop()
+				defer wg.Done()
+
+				for {
+					select {
+					case <-ctx.Done():
+						return
+					case <-t.C:
+						// Register in rendezvous points periodically
+						wakuNode.Rendezvous().RegisterWithNamespace(ctx, nodeTopic, iter.RendezvousPoints())
+					}
+				}
+			}(nodeTopic)
+
+			wg.Add(1)
+			go func(nodeTopic string) {
+				defer wg.Done()
+				desiredOutDegree := wakuNode.Relay().Params().D
+				t := time.NewTicker(7 * time.Second)
+				defer t.Stop()
+				for {
+					select {
+					case <-ctx.Done():
+						return
+					case <-t.C:
+						peerCnt := len(wakuNode.Relay().PubSub().ListPeers(nodeTopic))
+						peersToFind := desiredOutDegree - peerCnt
+						if peersToFind <= 0 {
+							continue
+						}
+
+						rp := <-iter.Next(ctx)
+						if rp == nil {
+							continue
+						}
+						ctx, cancel := context.WithTimeout(ctx, 7*time.Second)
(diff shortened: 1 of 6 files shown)


cmd/waku/relay.go
```diff
@@ -0,0 +1,137 @@
...
+
+		topics := maps.Keys(toTopics)
+		for _, subscription := range subscriptions {
+			wg.Add(1)
+			go func(subscription *relay.Subscription, topics []string) {
+				defer wg.Done()
+				for env := range subscription.Ch {
+					for _, topic := range topics {
+						// HACK: We append a magic number here, just so the pubsub message ID will change
+						env.Message().Meta = append(env.Message().Meta, 255)
```
richard-ramos, line 116: @jm-clius @alrevuelta 
Notice that here i'm appending a magic number just to force the pubsub message to have a different ID.
In https://github.com/waku-org/nwaku/blob/52759faa0341e5463afe6fb8cf51027216cc5d3b/waku/waku_relay/message_id.nim#L31-L33 We are only taking into account the payload of the pubsub message, but not the topic, meaning that you cant' have the same message payload publicated in different pubsub topics.  If I don't do this change, then I cannot fwd the message to a different pubsub topic.

I wonder if we should instead modify the message id function to also take into account the pubsub topic of the message? I'm okay with not modifying it tho, as this bridge functionality is something that will be used only during the transition period from default pubsub topic to static shard in status, and will probably remove this code after two status releases.

cmd/waku/relay.go
```diff
@@ -0,0 +1,147 @@
...
+		if !wakuNode.Relay().IsSubscribed(topic) {
+			_, err := wakuNode.Relay().Subscribe(ctx, wprotocol.NewContentFilter(topic), relay.WithoutConsumer())
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	for fromTopic, toTopics := range bridgedTopics {
+		subscriptions, err := wakuNode.Relay().Subscribe(ctx, wprotocol.NewContentFilter(fromTopic))
```
chaitanyaprem, line 106: Rather than publishing all messages on defaultTopic to the new shard. 
Wouldn't it be better to filter using contentTopics using by status app? 
DefaultTopic may be used by other users as well, and we would be flooding the shard 32 with unnecessary messages if we don't filter by contentTopics.

cmd/waku/relay.go
```diff
@@ -0,0 +1,147 @@
...
+						// HACK: message has been already fwded
+						metaLen := len(env.Message().Meta)
+						fwdTagLen := len(fwdMetaTag)
+						if metaLen >= fwdTagLen && bytes.Equal(env.Message().Meta[metaLen-fwdTagLen:], fwdMetaTag) {
+							continue
+						}
+
+						// HACK: We append magic numbers here, just so the pubsub message ID will change
+						env.Message().Meta = append(env.Message().Meta, fwdMetaTag...)
+						_, err := wakuNode.Relay().Publish(ctx, env.Message(), relay.WithPubSubTopic(topic))
```
chaitanyaprem, line 127: If we run this bridging logic in all fleet nodes, I am wondering if duplicates of same messages would get forwarded by each node onto the other topics. 
Not sure about this, but does messageID generation include the originating peerID which has published message into the pubsub? If so, then this would be a problem which needs to be solved.

cmd/waku/relay.go
```diff
@@ -0,0 +1,147 @@
...
+								zap.String("contentTopic", env.Message().ContentTopic), zap.Error(err))
+						}
+					}
+				}
+			}(subscription, topics)
+		}
+	}
+
+	// Protected topics
+	for _, protectedTopic := range options.Relay.ProtectedTopics {
```
chaitanyaprem, line 140: Better to do this before bridging go-routines are invoked above.

chaitanyaprem (commented):
Interesting approach to solve migration problem.
I am wondering if something like this would be required for other apps like TheGraph that want to migrate from namedpubsubTopic to sharding.

cmd/waku/relay.go
jm-clius: This is a very good point. In fact, the message id should in fact be the deterministic message hash defined in https://rfc.vac.dev/spec/14/#deterministic-message-hashing. Previously, though, we changed message ID to match go-waku because the impact of a message ID mismatch in a network is quite large. Perhaps the time has come though to bite the bullet and synchronise client upgrades for TWN?

richard-ramos: Yeah, i remember! it was very surprising how using different hashing function had such a large impact :worried: 
The only concern I have is that at this point we do not have a Waku message available yet, and we'd need to decode it from the protobuffer stored in the `data` attribute.

The fields i could see populated are these, based on the config we set for gossipsub:
```
message Message {
	optional string from = 1;
	optional bytes data = 2;    <--------------
	optional bytes seqno = 3;
        required string topic = 4;  <--------------
	optional bytes signature = 5;
	optional bytes key = 6;
}
```

cmd/waku/relay.go
richard-ramos: The problem is that a different content topic is generated per each channel you have joined so it's not possible to have a list of content topics to filter.

cmd/waku/relay.go
richard-ramos: Messages sent via relay will only have the payload and the topic; gossipsub was configured to not include signaturesm peer information or sequence numbers. So it should be okay, nodes that have already seen a messageID for some message they receive would not forward this message further

chaitanyaprem (approved):
LGTM

jm-clius (approved):
LGTM

PR #856, opened 2023-11-01
feat(rest-filterv2): get message

harsh-98:
# Description
Get messages endpoint. Related to: https://github.com/waku-org/go-waku/issues/746

# Changes

- [x]  add two endpoints based on discussion with richard. 
- [x] add test 

# Tests

- [x] new and previous tets are passing.

### DIFF
--- a/waku/v2/protocol/filter/client.go
+++ b/waku/v2/protocol/filter/client.go
@@ -481,8 +481,10 @@ func (wf *WakuFilterLightNode) Subscriptions() []*subscription.SubscriptionDetai
 	_, subs := wf.subscriptions.GetSubscription("", protocol.ContentFilter{})
 	return subs
 }
+
 func (wf *WakuFilterLightNode) IsListening(pubsubTopic, contentTopic string) bool {
 	return wf.subscriptions.IsListening(pubsubTopic, contentTopic)
+
 }
 
 // UnsubscribeWithSubscription is used to close a particular subscription


cmd/waku/flags.go
```diff
@@ -551,6 +551,13 @@ var (
 		Destination: &options.RESTServer.RelayCacheCapacity,
 		EnvVars:     []string{"WAKUNODE2_REST_RELAY_CACHE_CAPACITY"},
 	})
+	RESTFilterCacheCapacity = altsrc.NewIntFlag(&cli.IntFlag{
```
richard-ramos, line 554: @NagyZoltanPeter , it might be interesting to add this flag to nwaku, or maybe we should merge both message caches for both filter and relay? what do you think, @harsh-98

cmd/waku/server/rest/waku_rest.go
```diff
@@ -18,10 +18,11 @@ type WakuRest struct {
 
 	log *zap.Logger
 
-	relayService *RelayService
+	relayService  *RelayService
+	filterService *FilterService
 }
 
-func NewWakuRest(node *node.WakuNode, address string, port int, enablePProf bool, enableAdmin bool, relayCacheCapacity int, log *zap.Logger) *WakuRest {
+func NewWakuRest(node *node.WakuNode, address string, port int, enablePProf bool, enableAdmin bool, relayCacheCapacity, filterCacheCapacity int, log *zap.Logger) *WakuRest {
```
richard-ramos, line 25: I'm thinking it's time to refactor this function to receive a struct with the configuration instead of adding more functions! I'll open an issue so we can do that on a separate PR!

cmd/waku/server/rest/waku_rest.go
```diff
@@ -72,6 +77,7 @@ func (r *WakuRest) Start(ctx context.Context, wg *sync.WaitGroup) {
 
 	if r.node.Relay() != nil {
 		go r.relayService.Start(ctx)
+		go r.filterService.Start(ctx)
```
richard-ramos, line 80: Shouldn't this go in line 69 instead of here? looking at filter service implementation i don't see it using Relay()

cmd/waku/server/rest/filter_cache.go
```diff
@@ -0,0 +1,78 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	pubSubTopicMap, _ := protocol.ContentFilterToPubSubTopicMap(contentFilter)
+	for pubsubTopic, contentTopics := range pubSubTopicMap {
+		if c.data[pubsubTopic] == nil {
+			c.data[pubsubTopic] = make(map[string][]*pb.WakuMessage)
+		}
+		for _, topic := range contentTopics {
+			utils.Logger().Info(fmt.Sprintf("pb %s ct %s ", pubsubTopic, topic))
```
richard-ramos, line 35: Is this a leftover?
```suggestion
```

richard-ramos (approved):
:rocket: 

Left few comments/questions

cmd/waku/server/rest/waku_rest.go
richard-ramos: Tracking it here https://github.com/waku-org/go-waku/issues/859

PR #858, opened 2023-11-02
refactor: discv5 codebase

harsh-98:
# Description
Structure options/cli flags of discv5 in a single File. This will allow moving logic for creating discv5 to discv5 module. 
# Changes

- [x] discoveredNodes are set on `WakuNodeParameters` using `WithDiscoveredNOdes`, and this is then passed to Discv5.
- [x] Change type of discv5 in `WakuNode` from Service to `discv5.Discovery`. Reason golang a variable is nil if its type and value are both nil. `discv5Parameters.Service` returns nil of type `disv5.discovery` when service is not enabled. So, it is not equal to nil. And instance property of Service interface is not used for any purpose. I did this change. 
Another way to fix this is move service to common folder and change `discv5Parameters.Service` return type to Service.
- [x] Less context information for discv5 is needed outside discv5, like `disv5.enabled`. if discoveryv5 of WakuNode is nil, then the service is not being used. Currently `AutoUpdate` and `UdpPort` are exposed on `discoveryv5Params` as they are needed for localNode service. 

# Tests

### DIFF
--- a/waku/v2/discv5/option.go
+++ b/waku/v2/discv5/option.go
@@ -1,0 +1,132 @@
+package discv5
+
+import (
+	"crypto/ecdsa"
+
+	"github.com/ethereum/go-ethereum/p2p/enode"
+	"github.com/multiformats/go-multiaddr"
+	"github.com/prometheus/client_golang/prometheus"
+	cli "github.com/urfave/cli/v2"
+	"github.com/urfave/cli/v2/altsrc"
+	"github.com/waku-org/go-waku/waku/v2/dnsdisc"
+	"go.uber.org/zap"
+)
+
+type DiscV5Parameters struct {
+	// used by localNode
+	AutoUpdate bool
+	UdpPort    uint
+	// for cli
+	clinodes cli.StringSlice
+	//
+	enable        bool
+	autoFindPeers bool
+	bootnodes     []*enode.Node
+	advertiseAddr []multiaddr.Multiaddr
+	loopPredicate func(*enode.Node) bool
+}
+
+func (params *DiscV5Parameters) CliFlags() []cli.Flag {
+	return []cli.Flag{
+		altsrc.NewBoolFlag(&cli.BoolFlag{
+			Name:        "discv5-discovery",
+			Usage:       "Enable discovering nodes via Node Discovery v5",
+			Destination: &params.enable,
+			EnvVars:     []string{"WAKUNODE2_DISCV5_DISCOVERY"},
+		}), altsrc.NewStringSliceFlag(&cli.StringSliceFlag{
+			Name:        "discv5-bootstrap-node",
+			Usage:       "Text-encoded ENR for bootstrap node. Used when connecting to the network. Option may be repeated",
+			Destination: &params.clinodes,
+			EnvVars:     []string{"WAKUNODE2_DISCV5_BOOTSTRAP_NODE"},
+		}), altsrc.NewUintFlag(&cli.UintFlag{
+			Name:        "discv5-udp-port",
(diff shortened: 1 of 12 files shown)


waku/v2/discv5/option.go
```diff
@@ -0,0 +1,132 @@
...
+			Destination: &params.UdpPort,
+			EnvVars:     []string{"WAKUNODE2_DISCV5_UDP_PORT"},
+		}), altsrc.NewBoolFlag(&cli.BoolFlag{
+			Name:        "discv5-enr-auto-update",
+			Usage:       "Discovery can automatically update its ENR with the IP address as seen by other nodes it communicates with.",
+			Destination: &params.AutoUpdate,
+			EnvVars:     []string{"WAKUNODE2_DISCV5_ENR_AUTO_UPDATE"},
+		}),
+	}
+}
```
richard-ramos, line 54: I'm not sure I like this. The cli-flags are a concern for the `waku` command and not for the `waku` package itself (the library should not really be aware of how it is consumed). I'd rather have this as part of a package within cmd/waku

waku/v2/node/wakuoptions.go
```diff
@@ -373,12 +372,15 @@ func WithPeerStoreCapacity(capacity int) WakuNodeOption {
...
 	return func(params *WakuNodeParameters) error {
-		params.enableDiscV5 = true
-		params.udpPort = udpPort
-		params.discV5bootnodes = bootnodes
-		params.discV5autoUpdate = autoUpdate
+		params.discv5Params = discv5Params
+		return nil
+	}
+}
+func WithDiscoveredNodes(discoveredNodes []dnsdisc.DiscoveredNode) WakuNodeOption {
```
richard-ramos, line 381: IMO we're unnecesarily coupling DNSDiscovery with Discv5. DiscV5 is only concerned with ENRs for its bootnode setup

waku/v2/node/wakunode2.go
```diff
@@ -526,7 +525,8 @@ func (w *WakuNode) Stop() {
 	w.legacyFilter.Stop()
 	w.filterFullNode.Stop()
 
-	if w.opts.enableDiscV5 {
+	if w.discoveryV5 != nil {
+		utils.Logger().Info(fmt.Sprintf("%v", w.discoveryV5))
```
richard-ramos, line 529:
```suggestion
```

waku/v2/node/wakunode2.go
```diff
@@ -92,7 +93,7 @@ type WakuNode struct {
 
 	relay           Service
 	lightPush       Service
-	discoveryV5     Service
+	discoveryV5     *discv5.DiscoveryV5
```
richard-ramos, line 96: What's the advantage of using the specific type here compared to using the `Service` interface?

waku/v2/node/wakunode2.go
```diff
@@ -265,11 +266,9 @@ func New(opts ...WakuNodeOption) (*WakuNode, error) {
 		w.log.Error("creating peer connection strategy", zap.Error(err))
 	}
 
-	if w.opts.enableDiscV5 {
-		err := w.mountDiscV5()
-		if err != nil {
-			return nil, err
-		}
+	w.discoveryV5, err = w.opts.discv5Params.Service(w.opts.privKey, w.localNode, w.peerConnector, w.opts.prometheusReg, w.log, w.opts.discoveredNodes, w.opts.advertiseAddrs)
```
richard-ramos, line 269: This function is not very elegant due to the number of parameters it has (it's a code smell).

PR #860, opened 2023-11-02
feat: dont' filter out bootnodes

richard-ramos:
# Description
This avoids the problem described in https://github.com/waku-org/nwaku/issues/2162

### DIFF
--- a/waku/v2/discv5/discover.go
+++ b/waku/v2/discv5/discover.go
@@ -51,7 +51,7 @@ type DiscoveryV5 struct {
 type discV5Parameters struct {
 	autoUpdate    bool
 	autoFindPeers bool
-	bootnodes     []*enode.Node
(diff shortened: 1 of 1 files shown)


chaitanyaprem (approved):
LGTM

richard-ramos:
> lgtm. One question, why is defaultPredicate exposed?

I was thinking that someone using the `PeerIterator` function might be interested in also have a predicate that does the filtering of shards discv5 does by default. Perhaps it could even be used here https://github.com/waku-org/go-waku/blob/02f2800b046094f73d1011081daef7d897126687/waku/v2/protocol/peer_exchange/protocol.go#L143 as I see there's no particular filtering being applied to the nodes being returned by discv5

jm-clius:
👍

PR #862, opened 2023-11-03
fix: decode url param in relay rest API

chaitanyaprem:
# Description
Topic URL param is not being read properly in relay REST APIs which is causing 500 error. 

```
Hi, I'm trying to use relay via REST API in go-waku (wakuorg/go-waku:latest) as described here https://nagyzoltanpeter.github.io/nwaku-api/index.html#post-/relay/v1/subscriptions but I get errors for relay/v1/messages endpoint
 For example trying to publish a message I get a 500
 curl -vv -X POST "http://127.0.0.1:63970/relay/v1/messages/%2Fwaku%2F2%2Fdefault-waku%2Fproto" -H "content-type: application/json" -d '{"payload":"ZXhhbXBsZQ==","contentTopic":"string","version":0,"timestamp":0}'
 When querying messages I get 404 with not subscribed to topic
 curl -vv -X GET "http://127.0.0.1:63970/relay/v1/messages/%2Fwaku%2F2%2Fdefault-waku%2Fproto" -H "accept: application/json" 
 
If I use nwaku instead of gowaku it works fine. Also If I use gowaku with RPC instead of REST it works fine so the container setup seems correct. 
Also trying to subscribe the node to a topic seems to work fine, I get 200 here:
curl -vv -X POST "http://127.0.0.1:63970/relay/v1/subscriptions" -H "accept: text/plain" -H "content-type: application/json" -d '["/waku/2/default-waku/proto"]' 
```

# Changes

- [ ] decode URL param for post and get messages relay REST API

# Tests

Verified both the APIs with /waku/2/default-waku/proto topic and they seem to work fine.

### DIFF
--- a/cmd/waku/server/rest/relay.go
+++ b/cmd/waku/server/rest/relay.go
@@ -216,7 +216,6 @@ func (r *RelayService) postV1Message(w http.ResponseWriter, req *http.Request) {
 	}
 	defer req.Body.Close()
 
-	//var err error
 	if topic == "" {
 		topic = relay.DefaultWakuTopic
 	}
--- a/cmd/waku/server/rest/relay.go
+++ b/cmd/waku/server/rest/relay.go
@@ -171,8 +171,13 @@ func (r *RelayService) getV1Messages(w http.ResponseWriter, req *http.Request) {
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
-
-	var err error
+	topic, err := url.QueryUnescape(topic)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		_, err = w.Write([]byte("invalid topic format"))
+		r.log.Error("writing response", zap.Error(err))
+		return
+	}
 
 	r.messagesMutex.Lock()
 	defer r.messagesMutex.Unlock()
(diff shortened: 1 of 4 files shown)


cmd/waku/server/rest/relay.go
```diff
@@ -171,8 +171,13 @@ func (r *RelayService) getV1Messages(w http.ResponseWriter, req *http.Request) {
...
 	}
-
-	var err error
+	topic, err := url.QueryUnescape(topic)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		_, err = w.Write([]byte("invalid topic format"))
+		r.log.Error("writing response", zap.Error(err))
+		return
+	}
```
richard-ramos, line 180: @chaitanyaprem, @harsh-98,  what do you think if we move the code from https://github.com/waku-org/go-waku/pull/856/files#diff-66b9a518fabf46f0f294417e83bdd3e2305e7db3f7bc3a82a0425fa142968617R367-R381 into rest/utils.go so both filter and relay can use it?

cmd/waku/server/rest/relay.go
```diff
@@ -205,7 +216,7 @@ func (r *RelayService) postV1Message(w http.ResponseWriter, req *http.Request) {
 	}
 	defer req.Body.Close()
 
-	var err error
+	//var err error
```
richard-ramos, line 219:
```suggestion
```

cmd/waku/server/rest/relay.go
chaitanyaprem: Makes sense, did not think this code was already there. 
Will do the change.

richard-ramos: That PR is approved, we can go ahead merge it if it looks okay to you, and then rebase this one to contain that change! In the meantime i'll approve this, since this fix solves the problem anyway

richard-ramos (approved):

Issue #861, opened 2023-11-03
chore: Refactor relay REST API to use new subscriptions


chaitanyaprem:
### Background
The existing relay REST and RPC API's use a local message cache to read from subcription and store messages. Since the subscription has been updated to cache messages, this additional cache is not required anymore and can be removed. 

### Details

- [ ] Modify older relay REST API to use cache from subscription
- [ ] Modify relay RPC API to use cache from subscription
- [ ] Add unit tests for autosharding RPC and REST API

### Acceptance criteria
A list of tasks that need to be done for the issue to be considered resolved.

PR #866, opened 2023-11-06
chore: update relay REST and RPC API's and fix unit tests

chaitanyaprem:
While updating relay REST API's to remove duplicate message cache as part of #861 , have noticed that REST and RPC unit tests are not run as part of CI and many of them are failing. 

@harsh-98 did you not run and verify the Unit tests while working on #865 ? 

# Description

# Changes

- [x] Update old relay REST API's to use new subscription
- [x] fix relay REST & RPC unit tests 
- [x] fix RPC admin unit test
- [x] fix filter RPC and REST unit tests
- [x] Enable RPC and REST unit tests by default

Note: race condition failure should also be fixed by #867 

# Tests


cmd/waku/server/rpc/filter_test.go
```diff
@@ -50,6 +50,7 @@ func makeFilterService(t *testing.T, isFullNode bool) *FilterService {
 }
 
 func TestFilterSubscription(t *testing.T) {
+	t.Skip("skipping since it is legacy filter")
```
richard-ramos, line 53: TIL!

cmd/waku/server/rest/relay.go
```diff
@@ -44,11 +35,8 @@ func NewRelayService(node *node.WakuNode, m *chi.Mux, cacheCapacity int, log *za
 		node:          node,
 		log:           log.Named("relay"),
 		cacheCapacity: cacheCapacity,
```
richard-ramos, line 37: Are we not going to use a cache for the messages?  if that's the case, then we can remove the `cacheCapacity` parameter from this function and the attribute from `RelayService` struct. I have no preference between one vs the other, just wondering about it because this would be differnt from nwaku.

The difference between having the cache instead of relying on the subscription was that you could always have the last N more recent messages in the cache (since it worked kinda like a ring buffer), compared to using the subscription in which you'll always have the oldest 1000 messages, (iirc)

chaitanyaprem: There is another PR #869  where i had updated to use this cacheCapacity to pass it to subscription itself. Since now subscription itself acts like a cache, I don't see the need to maintain this second cache at REST level.

richard-ramos: But the difference still remains. The subscription will fill itself up with the first `cacheCapacity` messages and then stop receiving messages, while other mechanism you'd always have the last `cacheCapacity` mssages received.

Dunno if it might cause problems like being disconnected due to being a `subscriber too slow`, since perhaps the REST method will not be called frequently enough

chaitanyaprem: Aha, then maybe we can enhance subscription to take an option of over-riding older msgs in case of full?
Because if consumer is slow, it would make sense to take some action at subscription level itself. This way REST API or a regular API user gets same behaviour. 
I am also thinking we should align the behaviour between REST API and programmatic (golang or c-bindings) API.

richard-ramos: > maybe we can enhance subscription to take an option of over-riding older msgs in case of full

This sounds great! I'll open an issue for this and approve the PR then!

richard-ramos (approved):

cmd/waku/server/rpc/filter_test.go
chaitanyaprem: Had to search for what TIL meant. :)

PR #873, opened 2023-11-07
fix: panic if it is not possible to obtain the merkle root

### LINKED ISSUE #738: bug: handle errors generated in gm.rootTracker.UpdateLatestRoot(pair.Key.(uint64))
In https://github.com/waku-org/go-waku/pull/726#discussion_r1322478800 it was reported that an error updating the latest root should be considered critical. A failure in this function means that interacting with RLN failed, so we will probably end up with an inconsistent merkle tree

richard-ramos:
# Description
Panics if while retrieving the merkle root, there's a failure. Having an error in RLN at this stage will probably mean that we have an inconsistent state in the merkle tree, so it's better to stop the program execution

## Issue

closes #738

### DIFF
--- a/waku/v2/protocol/rln/group_manager/root_tracker.go
+++ b/waku/v2/protocol/rln/group_manager/root_tracker.go
@@ -4,7 +4,9 @@ import (
 	"bytes"
 	"sync"
 
+	"github.com/waku-org/go-waku/waku/v2/utils"
 	"github.com/waku-org/go-zerokit-rln/rln"
+	"go.uber.org/zap"
 )
 
 // RootsPerBlock stores the merkle root generated at N block number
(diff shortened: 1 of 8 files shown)


waku/v2/protocol/rln/group_manager/root_tracker.go
```diff
@@ -102,18 +101,18 @@ func (m *MerkleRootTracker) IndexOf(root [32]byte) int {
...
 // IDCommitment is received so we can keep track of the merkle root change
-func (m *MerkleRootTracker) UpdateLatestRoot(blockNumber uint64) (rln.MerkleNode, error) {
+func (m *MerkleRootTracker) UpdateLatestRoot(blockNumber uint64) rln.MerkleNode {
 	m.Lock()
 	defer m.Unlock()
 
 	root, err := m.rln.GetMerkleRoot()
 	if err != nil {
-		return [32]byte{}, err
+		utils.Logger().Named("root-tracker").Panic("could not retrieve merkle root", zap.Error(err))
```
chaitanyaprem, line 110: What would be the recovery here? 
If it just restarting the node, I am wondering if we should retry once more before declaring panic.

richard-ramos: The recovery is a restart. I'm not sure if we should retry. I'm worried about RLN being in some weird state after running into this error, and a restart wouldn't hurt much the sync process since it would still start from the last record synced instead of from scratch

chaitanyaprem (approved):
LGTM

PR #878, opened 2023-11-08
fix : issues with get messages API

chaitanyaprem:
# Description
@fbarbu15 reported few issues he faced while testing RPC and REST API for relay with latest go-waku image.
These are fixes for some of the issues noticed. 

# Changes

- [x] Do not use request context in subscribe API as that would finish as soon as request is done which would close the subscription channel as well.
- [x] Add checks for channel open status and return error for RPC and REST APIs to avoid panic
- [x] Get Subscription API to not return subscription which has noConsume flag set(which indicates that this subscription is not created via an API rather only during node initialization) 

# Tests

### DIFF
--- a/cmd/waku/server/rest/relay.go
+++ b/cmd/waku/server/rest/relay.go
@@ -91,7 +91,7 @@ func (r *RelayService) postV1Subscriptions(w http.ResponseWriter, req *http.Requ
 		} else {
 			topicToSubscribe = topic
 		}
-		_, err = r.node.Relay().Subscribe(req.Context(), protocol.NewContentFilter(topicToSubscribe), relay.WithCacheSize(r.cacheCapacity))
+		_, err = r.node.Relay().Subscribe(r.node.Relay().Context(), protocol.NewContentFilter(topicToSubscribe), relay.WithCacheSize(r.cacheCapacity))
 
 		if err != nil {
 			r.log.Error("subscribing to topic", zap.String("topic", strings.Replace(topicToSubscribe, "\n", "", -1)), zap.Error(err))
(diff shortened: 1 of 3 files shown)


cmd/waku/server/rpc/relay.go
```diff
@@ -206,8 +210,13 @@ func (r *RelayService) GetV1Messages(req *http.Request, args *TopicArgs, reply *
 	if err != nil {
 		return err
 	}
+	fmt.Println("subscription is ", sub)
```
richard-ramos, line 213:
```suggestion
```

cmd/waku/server/rest/relay.go
```diff
@@ -126,7 +126,14 @@ func (r *RelayService) getV1Messages(w http.ResponseWriter, req *http.Request) {
 	}
 	var response []*pb.WakuMessage
 	select {
-	case msg := <-sub.Ch:
+	case msg, open := <-sub.Ch:
+		if !open {
+			r.log.Error("consume channel is closed for subscription", zap.String("pubsubTopic", topic))
+			w.WriteHeader(http.StatusNotFound)
+			_, err = w.Write([]byte("consume channel is closed for subscription"))
+			r.log.Error("writing response", zap.Error(err))
```
richard-ramos, line 134:
```suggestion
			_, err = w.Write([]byte("consume channel is closed for subscription"))
			if err != nil {
				r.log.Error("writing response", zap.Error(err))
			}
```

cmd/waku/server/rpc/relay.go
```diff
@@ -150,7 +151,11 @@ func (r *RelayService) GetV1AutoMessages(req *http.Request, args *TopicArgs, rep
 		return err
 	}
 	select {
-	case msg := <-sub.Ch:
+	case msg, open := <-sub.Ch:
+		if !open {
+			r.log.Error("consume channel is closed for subscription", zap.String("pubsubTopic", args.Topic))
+			return errors.New("consume channel is closed for subscription")
```
richard-ramos, line 157: Probably makes sense to extract this error to a global var, and use it in here and in line 218

richard-ramos (approved):
LGTM, just left few observations

PR #879, opened 2023-11-08
chore: add test utils string generators

romanzac:
Description
Utility functions to generate strings of various nature for advanced testing

Tests available to run:
cd go-waku/tests
go test -run TestStringGenerators


richard-ramos (approved):
Seems okay to me.
I do wonder if there is value in having tests for test utils.

romanzac:
> Seems okay to me. I do wonder if there is value in having tests for test utils.

Testing string generators to minimize potential doubt; the tests I am about to write will run with different data for each iteration.

PR #880, opened 2023-11-08
fix: WakuMessage json encoding

richard-ramos:
# Description
Fixes an issue reported by @fbarbu15 in which the waku message content topic field was not being decoded correctly in the REST API

### DIFF
--- a/waku/v2/protocol/pb/codec.go
+++ b/waku/v2/protocol/pb/codec.go
@@ -1,0 +1,29 @@
+package pb
+
+import (
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+func (m *WakuMessage) MarshalJSON() ([]byte, error) {
+	return (protojson.MarshalOptions{}).Marshal(m)
+}
+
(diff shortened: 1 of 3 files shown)


fbarbu15 (approved):
LGTM, thanks

waku/v2/protocol/pb/codec.go
```diff
@@ -0,0 +1,29 @@
+package pb
+
+import (
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+func (m *WakuMessage) MarshalJSON() ([]byte, error) {
+	return (protojson.MarshalOptions{}).Marshal(m)
```
chaitanyaprem, line 9: The issue noticed was here https://github.com/waku-org/go-waku/blob/43412c9da5927d16c8befcb19d1511e707a9ef70/waku/v2/protocol/pb/message.pb.go#L31

where `content_topic` was being picked instead of `contentTopic` during decode.
How did both got assigned to the field? Was there some manual edit to the pb.go file post generation?

chaitanyaprem (approved):
LGTM

PR #881, opened 2023-11-08
fix: noisy peer exchange log

richard-ramos:
# Description
Removes some noise from the logs, by only adding a log when we discover a new peer


chaitanyaprem (approved):
LGTM

PR #883, opened 2023-11-08
chore: remove `--store-message-db-vacuum` 

richard-ramos:
# Description
`VACUUM`, as described in https://github.com/waku-org/nwaku/pull/2165 is an operation that takes a substantial amount of time to be executed depending on the number of rows in the database. Based on the experience so far since the introduction of this flag, It's for the better to execute this command as an operation done via cron or manually, instead of having it be part of the code of waku.

# Changes

- [x] Remove flag
- [x] Remove DB specific implementation of vacuum command
- [x] Left the `DBSettings` struct, since we might end up having to setup some RDBMS specific setting in the future?

### DIFF
--- a/cmd/waku/flags.go
+++ b/cmd/waku/flags.go
@@ -330,12 +330,6 @@ var (
 		Destination: &options.Store.DatabaseURL,
 		EnvVars:     []string{"WAKUNODE2_STORE_MESSAGE_DB_URL"},
 	})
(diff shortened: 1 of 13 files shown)


yakimant (approved):
Another option is https://sqlite.org/pragma.html#pragma_auto_vacuum
But I agree in general - not a part of waku code.

chaitanyaprem (approved):
LGTM

Issue #786, opened 2023-09-29
feat: Changes to Store protocol APIs for Autosharding


chaitanyaprem:
### Feature
Store API to be updated to take only contentTopics and derive pubSubTopic based on autosharding.

Probably reuse contentFilter which is extracted from Filter protocol for the same.

TBD - More analysis to be done to see what other changes are required apart from StoreClient API.

@SionoiS , Doesn't store API also require change to work with only contentTopics for autosharding? Did not see this being tracked.

SionoiS:
The protocol doesn't have to change because it already allow content topics without pubsub topics. I did not change anything in nwaku for STORE yet, but the implementation could be improved.

chaitanyaprem:
> The protocol doesn't have to change because it already allow content topics without pubsub topics. I did not change anything in nwaku for STORE yet, but the implementation could be improved.

Ok, was not aware of that. 
I meant the Store-Client side of the code. Wouldn't there be changes required to do peerSelection based on contentTopic alone i.e using autosharding to derive the pubSubTopic?

SionoiS:
> Wouldn't there be changes required to do peerSelection based on contentTopic alone i.e using autosharding to derive the pubSubTopic?

Yes that part I'm working on right now. https://github.com/waku-org/nwaku/pull/2092 but that just about finding a Store peer using discv5. Otherwise use the peer that come with the request.

> And also on the StoreNode side, wouldn't we need to derive the pubSubTopic for the contentTopic using autosharding (if pubSubTopic is not specified) to retrieve messages? Looks like there will be some work related to Store and Autosharding.

If the request don't specify pubsub topic we can ignore sharding and only filter messages by content topic.

The only problem is when the request has both pubsub and content topics. Now that's complicated. All content topic could fall on the same shard, each have their own shard or a mix. May require protocol changes not sure yet.

chaitanyaprem:
> The only problem is when the request has both pubsub and content topics. Now that's complicated. All content topic could fall on the same shard, each have their own shard or a mix. May require protocol changes, I haven't dived deep yet.
Hmm, To simplify this...maybe we can take an approach similar to what was done for Filter to handle autosharding. 
On Store-client side, split them to multiple requests i.e 1 per pubSubTopic and fire them towards StoreNode.

chaitanyaprem:
> The protocol doesn't have to change because it already allow content topics without pubsub topics. I did not change anything in nwaku for STORE yet, but the implementation could be improved.

[...]

PR #885, opened 2023-11-09
feat: update store client Query API for autosharding

chaitanyaprem:
# Description
Store API changes to support auto-sharding as per #786 

# Changes

- [x] Updated store API to support queries based on contentTopics where peer selection is done based on (autosharding)
- [x] Update peer selection to support finding peers supporting multiple pubsubTopics
- [x] Cleanup Store REST API  

# Tests

### DIFF
--- a/cmd/waku/server/rest/store.go
+++ b/cmd/waku/server/rest/store.go
@@ -14,7 +14,6 @@ import (
 	"github.com/waku-org/go-waku/waku/v2/node"
 	"github.com/waku-org/go-waku/waku/v2/protocol/store"
 	"github.com/waku-org/go-waku/waku/v2/protocol/store/pb"
-	"github.com/waku-org/go-waku/waku/v2/utils"
 )
 
 type StoreService struct {
@@ -56,7 +55,7 @@ func NewStoreService(node *node.WakuNode, m *chi.Mux) *StoreService {
 	return s
 }
 
-func getStoreParams(r *http.Request) (multiaddr.Multiaddr, *store.Query, []store.HistoryRequestOption, error) {
+func getStoreParams(r *http.Request) (*store.Query, []store.HistoryRequestOption, error) {
 	query := &store.Query{}
 	var options []store.HistoryRequestOption
 	var err error
@@ -65,15 +64,9 @@ func getStoreParams(r *http.Request) (multiaddr.Multiaddr, *store.Query, []store
 	if peerAddrStr != "" {
 		m, err = multiaddr.NewMultiaddr(peerAddrStr)
 		if err != nil {
-			return nil, nil, nil, err
+			return nil, nil, err
 		}
-
-		peerID, err := utils.GetPeerID(m)
-		if err != nil {
-			return nil, nil, nil, err
-		}
-
-		options = append(options, store.WithPeer(peerID))
+		options = append(options, store.WithPeerAddr(m))
 	}
 	query.PubsubTopic = r.URL.Query().Get("pubsubTopic")
 
@@ -86,7 +79,7 @@ func getStoreParams(r *http.Request) (multiaddr.Multiaddr, *store.Query, []store
 	if startTimeStr != "" {
 		startTime, err := strconv.ParseInt(startTimeStr, 10, 64)
 		if err != nil {
-			return nil, nil, nil, err
+			return nil, nil, err
 		}
 		query.StartTime = &startTime
 	}
@@ -95,7 +88,7 @@ func getStoreParams(r *http.Request) (multiaddr.Multiaddr, *store.Query, []store
 	if endTimeStr != "" {
 		endTime, err := strconv.ParseInt(endTimeStr, 10, 64)
 		if err != nil {
-			return nil, nil, nil, err
+			return nil, nil, err
 		}
 		query.EndTime = &endTime
 	}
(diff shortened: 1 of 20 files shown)


waku/v2/peerstore/waku_peer_store.go
```diff
@@ -207,7 +208,38 @@ func (ps *WakuPeerstoreImpl) PubSubTopics(p peer.ID) ([]string, error) {
...
+				peerTopicMap[topic] = struct{}{}
+			}
+			for _, topic := range pubSubTopics {
+				if _, ok := peerTopicMap[topic]; !ok {
+					match = false
+					break
+				}
+			}
+			if match {
+				result = append(result, p)
```
SionoiS, line 235: Could you not store a shard set for each peers then use set operations?

There are probably fancier way to store this data for easy access but would be good enough IMO.

SionoiS (approved):
I don't fully understand GO but LGTM.

waku/v2/peermanager/peer_manager.go
```diff
@@ -420,7 +420,7 @@ func (pm *PeerManager) addPeer(ID peer.ID, addrs []ma.Multiaddr, origin wps.Orig
 }
 
 // AddPeer adds peer to the peerStore and also to service slots
-func (pm *PeerManager) AddPeer(address ma.Multiaddr, origin wps.Origin, pubSubTopics []string, protocols ...protocol.ID) (peer.ID, error) {
+func (pm *PeerManager) AddPeer(address ma.Multiaddr, origin wps.Origin, pubSubTopics []string, connectNow bool, protocols ...protocol.ID) (peer.ID, error) {
```
richard-ramos, line 423: To avoid adding an extra parameter to the function to connect (which is only used in a single place), what do you think if we refactor it to return the `PeerData` and also split it into two separate functions, and modify AddrInfoToPeerData so it looks like this:


```go

func AddrInfoToPeerData(origin wps.Origin, peerID peer.ID, host host.Host, pubsubTopics ...string ) *service.PeerData {
	addrs := host.Peerstore().Addrs(peerID)
	if len(addrs) == 0 {
		//Addresses expired, remove peer from peerStore
		host.Peerstore().RemovePeer(peerID)
		return nil
	}
	return &service.PeerData{
		Origin: origin,
		AddrInfo: peer.AddrInfo{
			ID:    peerID,
			Addrs: addrs,
		},
               PubsubTopics: pubsubTopics,
	}
}

func (pm *PeerManager) AddPeer(address ma.Multiaddr, origin wps.Origin, pubSubTopics []string, protocols ...protocol.ID) (service.PeerData, error) {
	//Assuming all addresses have peerId
	info, err := peer.AddrInfoFromP2pAddr(address)
	if err != nil {
		return "", err
	}
	//Add Service peers to serviceSlots.
	for _, proto := range protocols {
		pm.addPeerToServiceSlot(proto, info.ID)
	}
	//Add to the peer-store
	err = pm.addPeer(info.ID, info.Addrs, origin, pubSubTopics, protocols...)
	if err != nil {
		return "", err
	}

        return AddrInfoToPeerData(origin, info.ID, pm.host, pubsubTopics...)
        return info, nil
}

func (pm *PeerManager) Connect(peerData service.PeerData) {
    go pm.peerConnector.PushToChan(peerData)
}
```
 

Then in waku_store_client.go, line 284, you'd do something like this:
```go
		peerData, err := store.pm.AddPeer(params.peerAddr, peerstore.Static, pubsubTopics, true, StoreID_v20beta4)
			if err != nil {
				return nil, err
			}

                        store.pm.Connect(peerData)

			params.selectedPeer = peerData.AddrInfo.ID
```

waku/v2/peerstore/waku_peer_store.go
```diff
@@ -207,7 +208,38 @@ func (ps *WakuPeerstoreImpl) PubSubTopics(p peer.ID) ([]string, error) {
...
+// If specifiPeers are listed, filtering is done from them otherwise from all peers in peerstore
+func (ps *WakuPeerstoreImpl) PeersByPubSubTopics(pubSubTopics []string, specificPeers ...peer.ID) peer.IDSlice {
+	if specificPeers == nil {
+		specificPeers = ps.Peers()
+	}
+	var result peer.IDSlice
+	for _, p := range specificPeers {
+		topics, err := ps.PubSubTopics(p)
+		if err == nil {
+			//Convoluted and crazy logic to find subset of topics
```
richard-ramos, line 221: lol

cmd/waku/server/rest/store.go
```diff
@@ -198,13 +199,9 @@ func (d *StoreService) getV1Messages(w http.ResponseWriter, r *http.Request) {
 
 	ctx, cancel := context.WithTimeout(r.Context(), 5*time.Second)
 	defer cancel()
-
-	_, err = d.node.AddPeer(peerAddr, peerstore.Static, d.node.Relay().Topics())
-	if err != nil {
-		writeStoreError(w, http.StatusInternalServerError, err)
-		return
+	if peerAddr != nil {
+		options = append(options, store.WithPeerAddr(peerAddr))
```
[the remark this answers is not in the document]
chaitanyaprem: I just moved it down from REST API thinking this could be a useful addition for all API's.
Yeah, created an issue #894 for having this option for other protocols.

chaitanyaprem:
@richard-ramos , looks like conventional commits check is failing for commit messages in this PR. 
Can we allow for existing commits to not follow this? 
I am not sure if it is a good idea to change older commit messages.

waku/v2/peermanager/peer_manager.go
chaitanyaprem: Good idea, i remember wanting to add a connect method but for some reason backed out if it and did this.

waku/v2/peerstore/waku_peer_store.go
chaitanyaprem: Yeah, this should get addressed once we move to our own peerstore implementation as we can choose the DS how we want it to be.

richard-ramos:
> looks like conventional commits check is failing for commit messages in this PR.

The fix is very easy :)

```
git fetch
git rebase origin/master        # This will make the commit history lineal and get rid also of the awful merge commits
```
and afterwards:
```
git rebase -i HEAD~5
```

And edit lines 1,2 and 4 so they look like this:
```
reword a6387751 rename topic to pubsubtopic in storeQuery                         
reword d8e9ce06 draft store api changes for autosharding                          
pick 6340702f feat: update peer selection to be based on multiple pubsubTopics, update store API
reword 904647dd connect to the store peer that is added via API                   
reword 086bfac8 chore: address review comments                                                   
```                                                 
and then, fix the messages so they look nice and tidy!

Changing older commits might be an issue only on `master` because other PRs might depend on those, but in this case since this PR is not the base of any other, it should be fine

richard-ramos (approved):

PR #886, opened 2023-11-09
chore: add messages logging subsystem

vitvly:
Adds a separate logging subsystem for relay/filter messages, to rely on https://github.com/ipfs/go-log#golog_log_level.

Also removes `logging.HexString` in favour of `logging.HexBytes`.

### DIFF
--- a/waku/v2/utils/logger.go
+++ b/waku/v2/utils/logger.go
@@ -9,6 +9,7 @@ import (
 )
 
 var log *zap.Logger
+var messageLoggers map[string]*zap.Logger
 
 // Logger creates a zap.Logger with some reasonable defaults
 func Logger() *zap.Logger {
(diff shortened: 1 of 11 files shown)


waku/v2/utils/logger.go
```diff
@@ -18,6 +19,15 @@ func Logger() *zap.Logger {
 	return log
 }
 
+// MessagesLogger returns a logger used for debug logging of receivent/sent messages
+func MessagesLogger() *zap.Logger {
```
chaitanyaprem, line 23: I was not aware we can use this logging to control levels at each subsystem this way.
Love the idea, but maybe it would be good to have subsystem at a protocol level which gives more fine-grain control. We can have msgLogging enabled either at relay or filter separately.

This function can take a prefix as an argument which each protocol can specify during initialization.

chaitanyaprem (commented):
Looks good
Left a minor comment. 

Also, is the logLevel configurable via a config option as well other than using env variable?

chaitanyaprem (approved):
Approving this as the comment is a minor one.

waku/v2/utils/logger.go
chaitanyaprem: Not sure if it is practically possible to maintain same logs across implementations. But maybe that could be done for specific trace logging like this one for messages. 


Yup, that way log level for each protocol can be changed/set to facilitate protocol specific debugging.

vitvly:
@chaitanyaprem i just added an extra change that i forgot to push yesterday: https://github.com/waku-org/go-waku/pull/886/commits/b469dae9f5cf48836e846e0b4e3e6883a4a0098e . I think this line is redundant.

waku/v2/utils/logger.go
vitvly: Pushed a change adding prefix to message loggers.

PR #888, opened 2023-11-09
fix(rest): use custom struct for messages instead of protobuffer

richard-ramos:
# Description
Previously, marshalling the protobuffer to json, ended up including a lot of fields that are not part of the rest api described in .yml files. With this PR, we use a struct `RestWakuMessage` which contains the fields described in the API

### DIFF
--- a/cmd/waku/server/rest/message.go
+++ b/cmd/waku/server/rest/message.go
@@ -1,0 +1,46 @@
+package rest
+
+import (
+	"errors"
+
+	"github.com/waku-org/go-waku/cmd/waku/server"
+	"github.com/waku-org/go-waku/waku/v2/protocol/pb"
+)
+
+type RestWakuMessage struct {
+	Payload      server.Base64URLByte `json:"payload"`
+	ContentTopic string               `json:"contentTopic"`
+	Version      *uint32              `json:"version,omitempty"`
+	Timestamp    *int64               `json:"timestamp,omitempty"`
+	Meta         []byte               `json:"meta,omitempty"`
+}
+
+func (r *RestWakuMessage) FromProto(input *pb.WakuMessage) error {
+	if err := input.Validate(); err != nil {
+		return err
+	}
+
+	r.Payload = input.Payload
+	r.ContentTopic = input.ContentTopic
+	r.Timestamp = input.Timestamp
+	r.Version = input.Version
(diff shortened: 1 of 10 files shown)


cmd/waku/server/rest/message.go
```diff
@@ -0,0 +1,46 @@
+package rest
+
+import (
+	"errors"
+
+	"github.com/waku-org/go-waku/cmd/waku/server"
+	"github.com/waku-org/go-waku/waku/v2/protocol/pb"
+)
+
+type RestWakuMessage struct {
```
chaitanyaprem, line 10: I am little confused as to the original issue we faced which forced us to change this. 
Refer https://github.com/golang/protobuf/issues/1388#issuecomment-1644112541 
I looks like if either of the json tag matches, then the decoding should happen. 
Also, omitEmpty should have taken care of fields not being filled into json if they are empty. 

Wondering if we root-caused the issue correctly or missing something.

chaitanyaprem: Ah, and there is no way for us to specify to ignore these fields in protobuf generation.
Then i guess we don't have a choice.

chaitanyaprem, line 10: Suggestion: Would be nice if we can generate this from yaml, but that can be taken up later once yamls get aligned across implementations.

chaitanyaprem (approved):
LGTM

cmd/waku/server/rest/message.go
richard-ramos: There's another option, we can use the protobuffer, and send an update to the waku-rest-api repo to include all the WakuMessage fields, including the rln proof!  I actually prefer that since it means less code for us to maintain

chaitanyaprem: Hmm, yeah..not a bad idea. We give all fields that are received and user can choose to ignore them if they don't want.

richard-ramos: I agree! it should be possible to even generate the handlers for REST endpoints from the yaml. These are some projects i have looked that we could consider for this task:
- https://github.com/contiamo/openapi-generator-go (this one is cool because it also uses `chi`)
- https://github.com/ogen-go/ogen
- https://github.com/OpenAPITools/openapi-generator

PR #891, opened 2023-11-09
chore: indicate that `--dns-discovery-url` can be repeated

richard-ramos:


jakubgs (approved):
I think repeatable options usually have a plural name, but this is good for now.

PR #898, opened 2023-11-14
fix: rest get messages not returning more than 1 message

chaitanyaprem:
# Description
Getting relay messages via REST API was not working correctly as it would only return a single message due to having a select with a default case without having a loop to drain the channel.

# Changes

- [ ] Added a loop to drain the channel until cacheCapacity or context is done
- [ ] Increase default relay cache capacity

# Tests

Tested locally using default pubsubtopic and noticed that messages were returned until channel is drained.


richard-ramos (approved):
Nice. I'll close the other PR since this is simpler!

PR #901, opened 2023-11-14
fix: allow to use `postgres` in DB url

richard-ramos:
# Description
Allow using `postgres` in the db connection string to match nwaku. 
(`postgresql` and `postgres` are valid db url prefixes according to `libpq`)


richard-ramos:
I'm merging this PR immediatly because the change is trivial

PR #906, opened 2023-11-15
fix: use subscription peerIds instead of separate peer slice

richard-ramos:
This fixes an issue that happened when unsubscribing from a full node, since the peer slice would not contain peer information when an specific peer was being selected


chaitanyaprem (approved):
LGTM

Would like to know specific scenario where we noticed the issue, because there must still be an underlying issue in some code-flow that must have caused the original issue that triggered this change. Probably somewhere in the flow, some data-structure update is missed?

richard-ramos:
I noticed this happening while running the filter example and looking at an error about that happened while trying to unsubsccribe, since the peerId we were using was empty!

PR #908, opened 2023-11-16
chore(ci): build using github actions

richard-ramos:
# Description
Replace some Jenkinfiles by Github actions


richard-ramos:
@chaitanyaprem, with this PR, we've got the following build times for all the targets I defined (still missing .deb/.rpm)
image

While working on it, I realized that this is a good time to revisit the decision of building all assets with each PR: is it worth it to build the mobile libraries, and linux packages (.rpm/.deb) always. or should we do this only for releases? WDYT?
The advantages of one vs the other is that it will speed up the PR feedback cycle, and also we don't really use all the files that are generated.

richard-ramos:
Added actions to build the .deb and .rpm too.  It probably makes sense to upload the macos builds both for amd64 and arm

chaitanyaprem:
> @chaitanyaprem, with this PR, we've got the following build times for all the targets I defined (still missing .deb/.rpm) image
> 
> While working on it, I realized that this is a good time to revisit the decision of building all assets with each PR: is it worth it to build the mobile libraries, and linux packages (.rpm/.deb) always. or should we do this only for releases? WDYT? The advantages of one vs the other is that it will speed up the PR feedback cycle, and also we don't really use all the files that are generated.

I agree that we don't need to build all packages for every PR, mainly only run the tests. But i did find that linux build helps identify issues which sometimes are not reported in other actions like test. I don't remember which ones actually, but maybe we can leave 1 linux build. 

Also, is there a way for a user to trigger these builds manually via github actions for a PR? If so, based on the need for a specific PR we can generate these.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,316 @@
...
+          common:
+          - '.github/workflows/**'
+          - 'Makefile'
+          - 'libs/**'
+
+          v2:
+          - 'waku/**'
+          - 'cmd/**'
+          - 'library/**'
+          - 'tests/**'
```
chaitanyaprem, line 39: How about examples? 
We may have to rerun some jobs on changes to examples right.
Especially the linter and build jobs.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,316 @@
...
+
+      - name: Upload asset
+        uses: actions/upload-artifact@v3
+        with:
+          name: gowaku-${{ steps.waku_version.outputs.version }}-ios.tar.gz
+          path: ./build/lib/gowaku-${{ steps.waku_version.outputs.version }}-ios.tar.gz
+          if-no-files-found: error
+
+
+  test:
```
chaitanyaprem, line 271: Wondering if it makes sense to keep all jobs in 1 file called ci.yml or have separate files and refer them in main file.

chaitanyaprem (approved):
LGTM

One question i have is in order to merge this PR, you will have to modify branch protection rules and merge. 
Also, can we update the checks to point to these individual actions if they are all defined in same file?

richard-ramos:
I'm going to revert deleting the Jenkinsfile and split everything in separate files in another PR. Once that's completed, We can delete again the jenkinfiles and setup branch protection rules to use github actions

richard-ramos:
> we update the checks to point to these individual actions if they are all defined in same file?

Yes!

richard-ramos:
> is there a way for a user to trigger these builds manually via github actions for a PR? If so, based on the need for a specific PR we can generate these.

Through the Actions tab, using the [Run Workflow] dropdown

image

PR #910, opened 2023-11-16
test: extend message timeout

romanzac:
Changes
Adding more time to wait for a message sent during the test. Messages could be up to 1MB and  it may take longer on slower computers to receive them.


richard-ramos (approved):

romanzac:
It is based on old commit not compatible with  naming conventions.

PR #912, opened 2023-11-16
fix(test): postgres

richard-ramos:
# Description
It seems that postgres tests were not working. When I executed `make test-postgres` I noticed that no tests were being executed, once the makefile was fixed, I noticed some compilation errors and failures to connect to the postgresdb.


chaitanyaprem (approved):
LGTM

PR #916, opened 2023-11-21
chore: filter v2 tests push invalid payload

romanzac:
Description
Batch of tests for Waku Filter v2 - PUSH with invalid payloads

Changes
Tests available to run:
cd go-waku/waku/v2/protocol/filter
go test


romanzac:
I wonder about timestamp field in Waku messages. There is no protection against insider attack. When someone wants to send messages in the past and alter order of messages, we let him do it. Is it difficult to verify common time flow (to have consensus on time) ? What are the trade offs here ? 

I imagine what would happen, if postal offices and government offices unite to collect fine fees with penalty stating something happened in the past. And I imagine they don't do it, because such attack is complex and slow. Not with Waku, though. Order can be altered quickly and wrong decisions or claims could be made at the same speed. 

Interested in your thoughts.

chaitanyaprem:
> I wonder about timestamp field in Waku messages. There is no protection against insider attack. When someone wants to send messages in the past and alter order of messages, we let him do it. Is it difficult to verify common time flow (to have consensus on time) ? What are the trade offs here ?
> 
> I imagine what would happen, if postal offices and government offices unite to collect fine fees with penalty stating something happened in the past. And I imagine they don't do it, because such attack is complex and slow. Not with Waku, though. Order can be altered quickly and wrong decisions or claims could be made at the same speed.
> 
> Interested in your thoughts.

This should get sorted out once RLN gets applied to light protocols as well. Since RLN is timestamp dependent that should take care of issues with timestamps (if someone tries to push older message, they get penalized). You can refer to discussion here https://github.com/waku-org/research/issues/45

waku/v2/protocol/filter/filter_proto_ident_test.go
```diff
@@ -225,7 +225,7 @@ func (s *FilterTestSuite) TestIncorrectSubscribeIdentifier() {
 	s.ctx, s.ctxCancel = context.WithTimeout(context.Background(), 10*time.Second) // Test can't exceed 10 seconds
 
 	s.testTopic = "/waku/2/go/filter/test"
-	s.testContentTopic = "TopicA"
+	s.testContentTopic = "/test/10/my-app"
```
chaitanyaprem, line 228: Better to use a const instead of copy pasting the same value in multiple places which is error prone.

waku/v2/protocol/filter/filter_push_test.go
```diff
@@ -141,3 +144,130 @@ func (s *FilterTestSuite) TestLargePayloadsUTF8() {
...
 
 }
+
+func (s *FilterTestSuite) TestEmptyPayload() {
+
+	// Subscribe
+	s.subDetails = s.subscribe(s.testTopic, s.testContentTopic, s.fullNodeHost.ID())
+
+	// Should get rejected
+	_, err := s.relayNode.Publish(s.ctx, tests.CreateWakuMessage(s.testContentTopic, utils.GetUnixEpoch(), ""), relay.WithPubSubTopic(s.testTopic))
```
chaitanyaprem, line 154: This is not a test for filter, because this is testing publish for a relayNode. Ideally we should have this test under relay tests.

chaitanyaprem (changes requested):
All the tests documented, none of them actually test anything wrt filter Push, rather only relay publish. 

I would expect these to be part of testing relay API than filter. 

What is the objective of the tests though if at all we are thinking of for filter protocol?

waku/v2/protocol/filter/filter_proto_ident_test.go
romanzac: I agree, fixed at [8c755ff](https://github.com/waku-org/go-waku/pull/916/commits/8c755ffc1ffcc70142b5f151899cf94d540bbd0a)

waku/v2/protocol/filter/filter_push_test.go
romanzac: There is no way how lightNode(filter) could send PUSH message, correct ?

chaitanyaprem: No, lightNode is a receiver of PUSH messaage. That can only be sent by a filterFullNode.

If you want to simulate pushMessage, you will have to use an internal method from filterFullNode `pushMessage`. 

But as my question remains, what are we trying to validate here? Because filterPush is not an API someone can use to push a message. It is automatically done by a filterFullNode on receiving any message on relay that is of interest to a filter subscriber.

romanzac: Understand, I would suggest to move negative tests to relay module and confirm positive tests with reception on the filter side. Does it look reasonable ?

romanzac:
> All the tests documented, none of them actually test anything wrt filter Push, rather only relay publish.
> 
> I would expect these to be part of testing relay API than filter.
> 
> What is the objective of the tests though if at all we are thinking of for filter protocol?

Changed according to your suggestions. Please have a look. Thanks.

waku/v2/protocol/relay/waku_relay_test.go
```diff
@@ -302,3 +305,53 @@ func TestWakuRelayAutoShard(t *testing.T) {
 	require.NoError(t, err)
 
 }
+
+func TestInvalidPushMessages(t *testing.T) {
```
chaitanyaprem, line 309: Minor one: rename to `TestInvalidMessagePublish`

waku/v2/protocol/filter/filter_test.go
```diff
@@ -29,6 +29,9 @@ func TestFilterSuite(t *testing.T) {
 	suite.Run(t, new(FilterTestSuite))
 }
 
+const defaultPubSubTopic = "/waku/2/go/filter/test"
```
chaitanyaprem, line 32: Better to rename to `defaultTestPubsubTopic` to avoid confusion. There is already a defaultpubsubTopic as "/waku/2/default-waku/proto" in pubsubtopic.go .

waku/v2/protocol/filter/filter_push_test.go
```diff
@@ -141,3 +144,84 @@ func (s *FilterTestSuite) TestLargePayloadsUTF8() {
...
+	// Subscribe
+	s.subDetails = s.subscribe(s.testTopic, s.testContentTopic, s.fullNodeHost.ID())
+
+	// Set time in the future
+	timeDelta, _ := time.ParseDuration("200h")
+	futureTime := proto.Int64(time.Now().UnixNano() + timeDelta.Nanoseconds())
+	message := tests.CreateWakuMessage(s.testContentTopic, futureTime, "test_payload")
+
+	// Should get accepted
+	_, err := s.relayNode.Publish(s.ctx, message, relay.WithPubSubTopic(s.testTopic))
```
chaitanyaprem, line 159: Not able to understand the objective of this test. 

Maybe this test should be put under RLNRelay where publish should get failed.

waku/v2/protocol/filter/filter_push_test.go
```diff
@@ -141,3 +144,84 @@ func (s *FilterTestSuite) TestLargePayloadsUTF8() {
...
+func (s *FilterTestSuite) TestNegativeTimestamp() {
+
+	// Subscribe
+	s.subDetails = s.subscribe(s.testTopic, s.testContentTopic, s.fullNodeHost.ID())
+
+	negTimestamp := int64(-100)
+	message := tests.CreateWakuMessage(s.testContentTopic, &negTimestamp, "test_payload")
+
+	// Should get accepted
+	_, err := s.relayNode.Publish(s.ctx, message, relay.WithPubSubTopic(s.testTopic))
```
chaitanyaprem, line 198: Again, trying to understand the objective of this test. 

Is it just to verify that message can be pushed wit negative timestamp?

waku/v2/protocol/filter/filter_push_test.go
romanzac: You are right, not much value of these tests without RLNRelay. I've removed them from PR. I'll study RLNRelay and reuse them later.

waku/v2/protocol/filter/filter_push_test.go
romanzac: Yes, not much of the value, testing grey zone. Removed for this PR.

waku/v2/protocol/relay/waku_relay_test.go
romanzac: Done at https://github.com/waku-org/go-waku/pull/916/commits/c076e36cec70d43bf80266b6ae33b175ce655860

waku/v2/protocol/filter/filter_test.go
romanzac: Done at https://github.com/waku-org/go-waku/pull/916/commits/c076e36cec70d43bf80266b6ae33b175ce655860

chaitanyaprem (approved):
LGTM

PR #918, opened 2023-11-22
chore: extract linux build to separate workflow

richard-ramos:
# Description
This workflow can be triggered via the [Actions] tab, once PR is merged
It also removes unused jenkinsfile that are no longer required due to https://github.com/waku-org/go-waku/pull/908


chaitanyaprem (approved):
LGTM

Wondering if we should have linux build also run by default as part of CI checks. 

And only leave mobile builds optional.

PR #920, opened 2023-11-22
fix: do not start metadata protocol unless required

richard-ramos:
## Changes
- [ ] Metadata protocol is not started unless clusterId is specified
- [ ] Simplifies peer disconnection logic, if an error is returned when doing a metadata request or if there is a clusterID mismatch

### DIFF
--- a/waku/v2/protocol/metadata/waku_metadata.go
+++ b/waku/v2/protocol/metadata/waku_metadata.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"math"
-	"strings"
 
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/libp2p/go-libp2p/core/host"
@@ -57,6 +56,7 @@ func (wakuM *WakuMetadata) SetHost(h host.Host) {
 func (wakuM *WakuMetadata) Start(ctx context.Context) error {
 	if wakuM.clusterID == 0 {
 		wakuM.log.Warn("no clusterID is specified. Protocol will not be initialized")
+		return nil
 	}
 
 	ctx, cancel := context.WithCancel(ctx)
@@ -135,16 +135,21 @@ func (wakuM *WakuMetadata) Request(ctx context.Context, peerID peer.ID) (*protoc
 	stream.Close()
 
 	if response.ClusterId == nil {
-		return nil, nil // Node is not using sharding
+		return nil, errors.New("node did not provide a waku clusterid")
 	}
 
-	result := &protocol.RelayShards{}
-	result.ClusterID = uint16(*response.ClusterId)
+	rClusterID := uint16(*response.ClusterId)
+	var rShardIDs []uint16
 	for _, i := range response.Shards {
-		result.ShardIDs = append(result.ShardIDs, uint16(i))
+		rShardIDs = append(rShardIDs, uint16(i))
 	}
 
-	return result, nil
+	rs, err := protocol.NewRelayShards(rClusterID, rShardIDs...)
+	if err != nil {
+		return nil, err
+	}
+
+	return &rs, nil
 }
 
 func (wakuM *WakuMetadata) onRequest(ctx context.Context) func(network.Stream) {
@@ -209,6 +214,15 @@ func (wakuM *WakuMetadata) ListenClose(n network.Network, m multiaddr.Multiaddr)
 	// Do nothing
 }
 
+func (wakuM *WakuMetadata) disconnectPeer(peerID peer.ID, reason error) {
+	logger := wakuM.log.With(logging.HostID("peerID", peerID))
+	logger.Error("disconnecting from peer", zap.Error(reason))
+	wakuM.h.Peerstore().RemovePeer(peerID)
+	if err := wakuM.h.Network().ClosePeer(peerID); err != nil {
+		logger.Error("could not disconnect from peer", zap.Error(err))
+	}
+}
+
 // Connected is called when a connection is opened
 func (wakuM *WakuMetadata) Connected(n network.Network, cc network.Conn) {
 	go func() {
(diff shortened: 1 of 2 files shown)


waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -225,14 +240,13 @@ func (wakuM *WakuMetadata) Connected(n network.Network, cc network.Conn) {
 		shard, err := wakuM.Request(wakuM.ctx, peerID)
 		if err == nil {
 			if shard == nil {
-				err = errors.New("no shard reported")
+				shouldDisconnect = false // No shard reported
```
chaitanyaprem, line 243: This doesn't seem right. 
The Request method can return nil, nil in case metadata protocol is not supported by the peer. Doesn't this mean we should disconnect from the peer because our own clusterID is non-Zero?

waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -225,14 +240,13 @@ func (wakuM *WakuMetadata) Connected(n network.Network, cc network.Conn) {
...
 			} else if shard.ClusterID != wakuM.clusterID {
 				err = errors.New("different clusterID reported")
 			}
 		} else {
 			// Only disconnect from peers if they support the protocol
-			// TODO: open a PR in go-libp2p to create a var with this error to not have to compare strings but use errors.Is instead
-			if strings.Contains(err.Error(), "protocols not supported") {
+			if isProtocolNotSupported(err) {
```
chaitanyaprem, line 249: Same comment as above, if clusterID is !=0 for localNode, then shouldn't we disconnect peers that don't support the metadata protocol?

Trying to understand what would be the case where we would need to be connected to such peer.


Also, this check seems duplicate to the one already done in lines 103-105 in this PR.

waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -89,11 +90,20 @@ func (wakuM *WakuMetadata) getClusterAndShards() (*uint32, []uint32, error) {
...
+	return errors.Is(err, notSupportedErr)
+}
+
 func (wakuM *WakuMetadata) Request(ctx context.Context, peerID peer.ID) (*protocol.RelayShards, error) {
 	logger := wakuM.log.With(logging.HostID("peer", peerID))
 
 	stream, err := wakuM.h.NewStream(ctx, peerID, MetadataID_v1)
 	if err != nil {
+		if isProtocolNotSupported(err) {
+			return nil, nil
```
chaitanyaprem, line 104: This return type is misleading for this function as the stream opening failed and we are still returning err non nil.
Better to add such behvaiour in comments so that caller is aware.

waku/v2/protocol/metadata/waku_metadata.go
kaichaosun, line 243: Here the logic is pretty confusing, is it possible to make it like this?
```
shard, err := wakuM.Request(wakuM.ctx, peerID)

if err != nil {
    logError()
    // disconnect from peers if error happens in request and they support the protocol
    if protocolSupported(err) {
        removeAndClosePeer()
    }
}

if shard != nil && shard.ClusterID != wakuM.clusterID {
    // disconnect if nodes in different cluster
     removeAndClosePeer()
}

return
```

In `Request` method, no longer need this
```
if isProtocolNotSupported(err) {
    return nil, nil
}
```
and can just return the error.

richard-ramos: Yes, this was a misunderstanding I had about the metadata protocol that I cleared today. Will update the PR

richard-ramos: I used a variant of your suggestion, as verifying for the protocol is no longer needed, since we are mounting waku_metadata only if the clusterID != 0, and then we expect that peers support the protocol, otherwise we can drop them.

waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -134,17 +144,22 @@ func (wakuM *WakuMetadata) Request(ctx context.Context, peerID peer.ID) (*protoc
 
 	stream.Close()
 
-	if response.ClusterId == nil {
+	if response.GetClusterId() == 0 {
```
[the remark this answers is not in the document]
richard-ramos: I *think* that spec is outdated now, since we're assuming in nwaku too that clusterID 0 means no support for metadata protocol. 
cc: @alrevuelta @jm-clius

waku/v2/protocol/metadata/waku_metadata.go
richard-ramos: With last commit, this observation no longer applies

waku/v2/protocol/metadata/waku_metadata.go
richard-ramos: fixed in last commit

waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -135,16 +135,21 @@ func (wakuM *WakuMetadata) Request(ctx context.Context, peerID peer.ID) (*protoc
 	stream.Close()
 
 	if response.ClusterId == nil {
-		return nil, nil // Node is not using sharding
+		return nil, errors.New("node not subscribed to any shard")
```
chaitanyaprem, line 138:
```suggestion
		return nil, errors.New("node did not provide a waku clusterid")
```

waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -209,6 +214,15 @@ func (wakuM *WakuMetadata) ListenClose(n network.Network, m multiaddr.Multiaddr)
 	// Do nothing
 }
 
+func (wakuM *WakuMetadata) disconnectPeer(peerID peer.ID, reason error) {
+	logger := wakuM.log.With(logging.HostID("peerID", peerID))
+	logger.Error("disconnecting from peer", zap.Error(reason))
+	wakuM.h.Peerstore().RemovePeer(peerID)
+	if err := wakuM.h.Network().ClosePeer(peerID); err != nil {
+		logger.Error("could not disconnect from peer", zap.Error(err))
```
chaitanyaprem, line 222: This error is interesting, wondering what could be the reason for this failure.
If it is just that remote-peer has disconnected before us, then maybe we shouldn't log error.
If not, would adding a retry here make it more fool-proof?

waku/v2/protocol/metadata/waku_metadata.go
```diff
@@ -219,30 +233,14 @@ func (wakuM *WakuMetadata) Connected(n network.Network, cc network.Conn) {
 
 		peerID := cc.RemotePeer()
 
-		logger := wakuM.log.With(logging.HostID("peerID", peerID))
-
-		shouldDisconnect := true
 		shard, err := wakuM.Request(wakuM.ctx, peerID)
-		if err == nil {
-			if shard == nil {
```
chaitanyaprem, line 227: Is this error handled by new changes?
Or is shard support something we are not worried about as of now?

chaitanyaprem (approved):
LGTM

Left few comments.

kaichaosun (approved):
LGTM

waku/v2/protocol/metadata/waku_metadata.go
richard-ramos: It is already handled by the changes. Instead of returning a null shard, I just return an error.

waku/v2/protocol/metadata/waku_metadata.go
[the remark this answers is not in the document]
jm-clius: I wouldn't view this as a spec change - just a limited implementation. In other words, so far we have only implemented the Metadata protocol for TWN which is on cluster ID = `1`. This is to avoid breaking backwards compatibility for now, but this does not mean that the protocol fundamentally views `0` as "no support for protocol".

PR #929, opened 2023-11-27
feat(c-bindings): support creating multiple instances

### LINKED ISSUE #847: chore(c-bindings): multiple nodes
Currently it's only possible to run a single node with the bindings. We need to modify all the API to allow multiple nodes, with `waku_new` returning an instance that can be passed to all the API functions to modify that instance / interact with it

richard-ramos:
# Description
Adds support to create multiple instances of nodes when using c-bindings

## Issue

closes #847


library/node.go
```diff
@@ -51,10 +63,66 @@ func randomHex(n int) (string, error) {
 	return hex.EncodeToString(bytes), nil
 }
 
+func Init() *WakuInstance {
+	id := uint(len(wakuInstances))
```
chaitanyaprem, line 67: Need synchronization for accessing and updating this map, otherwise it can lead to issues. 

What if Init gets called in 2 threads and both return same len of the map?

library/node.go
```diff
@@ -51,10 +63,66 @@ func randomHex(n int) (string, error) {
 	return hex.EncodeToString(bytes), nil
 }
 
+func Init() *WakuInstance {
```
chaitanyaprem, line 66: Wondering rather than returning pointers across language, why can't we just use instanceID (a number).

I am envisioning a flow like this:

App(C code) -> calls waku_createNode(config) which returns a node along with instanceID of the wakuNode.

All further invocations from the app  would just pass the instanceID for invoking further methods. Same instanceID is mentioned in callback as well.

the library shall maintain instanceID -> actual WakuNode pointer mapping to resolve.

richard-ramos: Curiously enough, the original version of the c-bindings supported the creation of multiple instances with a `nodeID` variable being required https://github.com/waku-org/go-waku/pull/212 but it was removed based on this discussion: https://github.com/vacp2p/rfc/pull/501#discussion_r833329214

Ivansete-status: I understand the approach and that would require an internal management of the waku_node instances. Nevertheless, what is the problem of returning the waku_node address? Even though the "integer-returning" approach can work well and is acceptable from my point of view, I wouldn't invest in it unless we find a clear reason to adopt it.

chaitanyaprem: The waku_node instances have to be managed anyway right, even if you pass pointer? 
Or is that not the case with nwaku?

chaitanyaprem: This i am fine with. 
Was just wondering if we could reduce the number of calls if we use an instanceId. 
As of now a user has to do `waku_init`, then set callback and then call `waku_new`. 

The flow looks non-intuitive. Ideally it should just be either waku_init or waku_new which returns instance(pointer or id) and also takes callback as an arg. If this can be achieved with either approach then i am not particular about either.

richard-ramos: I see. In that case assuming we want to use a void pointer, it would become very annoying to implement because of the way we use callbacks to return results in functions (which I personally don't like since i found that the original version in which we called `waku_free` for each value returned was easier to use than the callback mechanism we use now)

With the callback based approach we don't return values directly but a return code to indicate if there was an error or not, and we need to pass a `WakuCallback` which is defined as follows:
```cpp
typedef void (*WakuCallBack) (int ret_code, const char* msg, void * user_data);
```
Do note that `msg` parameter which would contain the response is a `const char*`, so if we wanted to use a void pointer, we'd have to define a different type of callback, while using an integer id wouldn't be so problematic as converting an int to a string is not a big deal.

But on the other hand, I would not expect developers to really use the c-bindings directly but to try instead to abstract away these complexities under some developer friendly wrappers where they can just do something like:
```rs
let result = waku::new(theConfig, onMessageCallback)
match result {
        Ok(node) => do something with the node,
        Err(error) => panic!("could not instantiate the node: {:?}", error),
    };
```

Ivansete-status (approved):
LGTM! Thanks for it!

library/node.go
richard-ramos: Latest commit removes the need of `waku_init`, with `waku_new` matching nwaku function definition

PR #930, opened 2023-11-28
chore: allow custom resolver for dns discovery

kaichaosun:
# Description

Add opt to allow custom resolver, use the dnsdisc.Resolver interface instead of the net.Resolver data type.

Related to: https://github.com/status-im/status-go/pull/4381

### DIFF
--- a/waku/v2/dnsdisc/enr.go
+++ b/waku/v2/dnsdisc/enr.go
@@ -15,6 +15,7 @@ import (
 
 type dnsDiscoveryParameters struct {
 	nameserver string
+	resolver   dnsdisc.Resolver
 }
 
 type DNSDiscoveryOption func(*dnsDiscoveryParameters)
@@ -23,6 +24,12 @@ type DNSDiscoveryOption func(*dnsDiscoveryParameters)
 func WithNameserver(nameserver string) DNSDiscoveryOption {
 	return func(params *dnsDiscoveryParameters) {
 		params.nameserver = nameserver
+	}
+}
+
+func WithResolver(resolver dnsdisc.Resolver) DNSDiscoveryOption {
+	return func(params *dnsDiscoveryParameters) {
+		params.resolver = resolver
 	}
 }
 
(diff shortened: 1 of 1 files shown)


waku/v2/dnsdisc/enr.go
```diff
@@ -26,6 +27,12 @@ func WithNameserver(nameserver string) DNSDiscoveryOption {
 	}
 }
 
+func WithResolver(resolver dnsdisc.Resolver) DNSDiscoveryOption {
```
chaitanyaprem, line 30: I am wondering if this should be mutually exclusive with `WithNameserver` , as only one of them would get used eventually. In that case maybe we can return error. 

You can take a look at how this is done in https://github.com/waku-org/go-waku/blob/d7249fc123d3a27e3eb60b85d58d4d7a51df64a1/waku/v2/protocol/filter/options.go#L73

kaichaosun: Thanks for point out other options. @chaitanyaprem 
The error handling provides here seems much valuable for critical check. 
It seems the mutually exclusive check in params will not reduce the check in `RetrieveNodes` function. For such optional parameters, constrain the logic to less places probably makes code more readable.

chaitanyaprem: Did not get your point here.
What i was trying to say is If we check for the 2 params to be mutually exclusive in the option itself, then error would be thrown in retrieveNodes if both the params are set. 
Something like below:

```suggestion
func WithResolver(resolver dnsdisc.Resolver) DNSDiscoveryOption {
	return func(params *dnsDiscoveryParameters) error{
	        if params.nameserver!= ""{
	              return errors.New("nameserver and dnsresolver are mutually exclusive, any of the one params have to be set")
	        }
		params.resolver = resolver
	}
```

Note that this is not important, but a nice a to have. 
If you want to take it up separately, that is also fine. Don't want to block you from proceeding. Hence approving this PR.

chaitanyaprem (approved):
Approving not to block.

Comment can be addressed separately as well.

PR #931, opened 2023-11-28
chore: filter v2 tests coverage improvement

romanzac:
Description
Batch of tests for Waku Filter v2 to improve coverage

Changes
Tests available to run:
cd go-waku/waku/v2/protocol/filter
go test

### DIFF
--- a/waku/v2/protocol/filter/filter_subscribe_test.go
+++ b/waku/v2/protocol/filter/filter_subscribe_test.go
@@ -398,3 +398,31 @@ func (s *FilterTestSuite) TestIsSubscriptionAlive() {
 	s.Require().Error(err)
 
 }
+
+func (s *FilterTestSuite) TestFilterSubscription() {
+	contentFilter := protocol.ContentFilter{PubsubTopic: s.testTopic, ContentTopics: protocol.NewContentTopicSet(s.testContentTopic)}
+
+	// Subscribe
+	s.subDetails = s.subscribe(s.testTopic, s.testContentTopic, s.fullNodeHost.ID())
+
+	// Returns no error and SubscriptionDetails for existing subscription
+	_, err := s.lightNode.FilterSubscription(s.fullNodeHost.ID(), contentFilter)
+	s.Require().NoError(err)
+
+	otherFilter := protocol.ContentFilter{PubsubTopic: "34583495", ContentTopics: protocol.NewContentTopicSet("sjfa402")}
+
+	// Returns error and nil SubscriptionDetails for non existent subscription
+	nonSubscription, err := s.lightNode.FilterSubscription(s.fullNodeHost.ID(), otherFilter)
+	s.Require().Error(err)
+	s.Require().Nil(nonSubscription)
+
+	// Create new host/peer - not related to any node
+	host, err := tests.MakeHost(context.Background(), 54321, rand.Reader)
+	s.Require().NoError(err)
+
+	// Returns error and nil SubscriptionDetails for unrelated host/peer
+	nonSubscription, err = s.lightNode.FilterSubscription(host.ID(), contentFilter)
+	s.Require().Error(err)
+	s.Require().Nil(nonSubscription)
+
+}
--- a/waku/v2/protocol/filter/filter_subscribe_test.go
+++ b/waku/v2/protocol/filter/filter_subscribe_test.go
@@ -426,3 +426,27 @@ func (s *FilterTestSuite) TestFilterSubscription() {
 	s.Require().Nil(nonSubscription)
 
 }
+
+func (s *FilterTestSuite) TestHandleFilterSubscribeOptions() {
+	contentFilter := protocol.ContentFilter{PubsubTopic: s.testTopic, ContentTopics: protocol.NewContentTopicSet(s.testContentTopic)}
+
+	// Subscribe
+	s.subDetails = s.subscribe(s.testTopic, s.testContentTopic, s.fullNodeHost.ID())
+
+	// With valid peer
+	opts := []FilterSubscribeOption{WithPeer(s.fullNodeHost.ID())}
+
+	// Positive case
+	_, _, err := s.lightNode.handleFilterSubscribeOptions(s.ctx, contentFilter, opts)
+	s.Require().NoError(err)
+
+	addr := s.fullNodeHost.Addrs()[0]
+
+	// Combine mutually exclusive options
+	opts = []FilterSubscribeOption{WithPeer(s.fullNodeHost.ID()), WithPeerAddr(addr)}
+
+	// Should fail on wrong option combination
+	_, _, err = s.lightNode.handleFilterSubscribeOptions(s.ctx, contentFilter, opts)
+	s.Require().Error(err)
+
+}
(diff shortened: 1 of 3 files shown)


romanzac:
@chaitanyaprem I would like to wrap up filter tests with covering certain error cases. I wonder if you see the same value in https://github.com/waku-org/go-waku/pull/931/commits/49435a833f0cc937d8c86992dde11ff060c6c223 I'd say integrator might appreciate to receive error when sending UnsubscribeAll with wrong peer.ID (host.ID) by mistake. Being an attacker would  probably give up looking at peer ID length. So I would say benefit >> security risk to provide error?

chaitanyaprem:
> @chaitanyaprem I would like to wrap up filter tests with covering certain error cases. I wonder if you see the same value in [49435a8](https://github.com/waku-org/go-waku/commit/49435a833f0cc937d8c86992dde11ff060c6c223) I'd say integrator might appreciate to receive error when sending UnsubscribeAll with wrong peer.ID (host.ID) by mistake. Being an attacker would probably give up looking at peer ID length. So I would say benefit >> security risk to provide error?

Good one, looks like as of now the result will just be an empty instead of an error indicating no subscription found for the peer which is not a great dev ex.
This could be addressed so that the API returns appropriate error, but I don't understand how it is some secuirty risk.

romanzac:
> > @chaitanyaprem I would like to wrap up filter tests with covering certain error cases. I wonder if you see the same value in [49435a8](https://github.com/waku-org/go-waku/commit/49435a833f0cc937d8c86992dde11ff060c6c223) I'd say integrator might appreciate to receive error when sending UnsubscribeAll with wrong peer.ID (host.ID) by mistake. Being an attacker would probably give up looking at peer ID length. So I would say benefit >> security risk to provide error?
> 
> Good one, looks like as of now the result will just be an empty instead of an error indicating no subscription found for the peer which is not a great dev ex. This could be addressed so that the API returns appropriate error, but I don't understand how it is some secuirty risk.

It is not a security risk, or it is just a negligible one. To compare, there is also no error when we unsubscribe from non existent topic. Here, topic could be very short string, and could be potentially guessed by executing enough unsubscribe calls. Attacker could then listen to conversation or spam (up to rate the limit). It is good we don't return any error code at that case.

I've created improvement https://github.com/waku-org/go-waku/issues/933

chaitanyaprem:
> > > @chaitanyaprem I would like to wrap up filter tests with covering certain error cases. I wonder if you see the same value in [49435a8](https://github.com/waku-org/go-waku/commit/49435a833f0cc937d8c86992dde11ff060c6c223) I'd say integrator might appreciate to receive error when sending UnsubscribeAll with wrong peer.ID (host.ID) by mistake. Being an attacker would probably give up looking at peer ID length. So I would say benefit >> security risk to provide error?
> > 
> > 
> > Good one, looks like as of now the result will just be an empty instead of an error indicating no subscription found for the peer which is not a great dev ex. This could be addressed so that the API returns appropriate error, but I don't understand how it is some secuirty risk.
> 
> It is not a security risk, or it is just a negligible one. To compare, there is also no error when we unsubscribe from non existent topic. Here, topic could be very short string, and could be potentially guessed by executing enough unsubscribe calls. Attacker could then listen to conversation or spam (up to rate the limit). It is good we don't return any error code at that case.
> 
> I've created improvement #933

We need to return the proper errors so that dev ex is better. 
Wrt an attacker listening to conversation, they can listen to any messages as long as they are subscribed to a pubsubTopic using relay (since this is a gossipsub network), which is why it is recommended for applications to encrypt their data so that 3rd parties cannot read the actual content. 
Also note that there is no rate limit when it comes to listening/receiving messages, the only rate-limit is for publishing messages to the network (in order to avoid spam).

chaitanyaprem (approved):
LGTM

richard-ramos (approved):

romanzac:
> > > > @chaitanyaprem I would like to wrap up filter tests with covering certain error cases. I wonder if you see the same value in [49435a8](https://github.com/waku-org/go-waku/commit/49435a833f0cc937d8c86992dde11ff060c6c223) I'd say integrator might appreciate to receive error when sending UnsubscribeAll with wrong peer.ID (host.ID) by mistake. Being an attacker would probably give up looking at peer ID length. So I would say benefit >> security risk to provide error?
> > > 
> > > 
> > > Good one, looks like as of now the result will just be an empty instead of an error indicating no subscription found for the peer which is not a great dev ex. This could be addressed so that the API returns appropriate error, but I don't understand how it is some secuirty risk.
> > 
> > 
> > It is not a security risk, or it is just a negligible one. To compare, there is also no error when we unsubscribe from non existent topic. Here, topic could be very short string, and could be potentially guessed by executing enough unsubscribe calls. Attacker could then listen to conversation or spam (up to rate the limit). It is good we don't return any error code at that case.
> > I've created improvement #933
> 
> We need to return the proper errors so that dev ex is better. Wrt an attacker listening to conversation, they can listen to any messages as long as they are subscribed to a pubsubTopic using relay (since this is a gossipsub network), which is why it is recommended for applications to encrypt their data so that 3rd parties cannot read the actual content. Also note that there is no rate limit when it comes to listening/receiving messages, the only rate-limit is for publishing messages to the network (in order to avoid spam).

@chaitanyaprem  Would you like to fix https://github.com/waku-org/go-waku/issues/933 some time soon ? This PR has a test which depends on fixing #933, so it won't pass required checks before fixing.

chaitanyaprem:
> > > > > @chaitanyaprem I would like to wrap up filter tests with covering certain error cases. I wonder if you see the same value in [49435a8](https://github.com/waku-org/go-waku/commit/49435a833f0cc937d8c86992dde11ff060c6c223) I'd say integrator might appreciate to receive error when sending UnsubscribeAll with wrong peer.ID (host.ID) by mistake. Being an attacker would probably give up looking at peer ID length. So I would say benefit >> security risk to provide error?
> > > > 
> > > > 
> > > > Good one, looks like as of now the result will just be an empty instead of an error indicating no subscription found for the peer which is not a great dev ex. This could be addressed so that the API returns appropriate error, but I don't understand how it is some secuirty risk.
> > > 
> > > 
> > > It is not a security risk, or it is just a negligible one. To compare, there is also no error when we unsubscribe from non existent topic. Here, topic could be very short string, and could be potentially guessed by executing enough unsubscribe calls. Attacker could then listen to conversation or spam (up to rate the limit). It is good we don't return any error code at that case.
> > > I've created improvement #933
> > 
> > 
> > We need to return the proper errors so that dev ex is better. Wrt an attacker listening to conversation, they can listen to any messages as long as they are subscribed to a pubsubTopic using relay (since this is a gossipsub network), which is why it is recommended for applications to encrypt their data so that 3rd parties cannot read the actual content. Also note that there is no rate limit when it comes to listening/receiving messages, the only rate-limit is for publishing messages to the network (in order to avoid spam).
> 
> @chaitanyaprem Would you like to fix #933 some time soon ? This PR has a test which depends on fixing #933, so it won't pass required checks before fixing.

Just merged the fix on to master. If you update your branch, you should have the fix in it and ci would pass :)

PR #934, opened 2023-11-29
chore: error handling in params construction

kaichaosun:
# Description

The change add error handling based on this comment https://github.com/waku-org/go-waku/pull/930#discussion_r1407134171
~This seems a bit tedius to me and brings less value. How do you think?~

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,11 +1,3 @@
-version: "2"
-
-checks:
-  return-statements:
-    enabled: true
-    config:
-      threshold: 6
-
 plugins:
 #  golint:
 #    enabled: true
@@ -15,7 +7,6 @@ plugins:
     enabled: true
 #  golangci-lint:
 #    enabled: true
-
 exclude_patterns:
   - "."
   - "**/*.pb.go"
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,3 +1,11 @@
+version: "2"
+
+checks:
+  return-statements:
+    enabled: true
+    config:
+      threshold: 6
+
 plugins:
 #  golint:
 #    enabled: true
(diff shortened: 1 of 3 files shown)


richard-ramos (approved):
I like the change. It will for sure improve the dev experience, since then a developer will know for sure that they cannot use both options at the same time (before they could but then the nameservers would have been ignored, so this will potentially save them debugging time and for us, having to give support in case such scenario happens)

chaitanyaprem (approved):
LGTM

kaichaosun:
codeclimate check is failed with excessive return statement, should I change it somewhere in their site or add the config in go-waku repo?
I have no clue why test-ci failed.

chaitanyaprem:
> codeclimate check is failed with excessive return statement, I loose the check for 6 return statements. I can also refactor the code to reduce the return statement, but have 4 return statement looks pretty strict.
> 
> Any clue why test-ci failed? It's happy now. 😂

Codeclimate check is not mandatory as it has some weird checks like this return statements which are not possible to enforce all the time. 

test-ci failure could be due to a flaky test, have noticed this sometimes before as well. 

But, I don't think increasing return statements to 6 helps much, i would suggest if we can make codeclimate ignore this in total. @richard-ramos WDYT?

richard-ramos:
> Any clue why test-ci failed? It's happy now.

Probably due to the flaky postgresql tests

chaitanyaprem:
> > But, I don't think increasing return statements to 6 helps much, i would suggest if we can make codeclimate ignore this in total. @richard-ramos WDYT?
> 
> We have 3 options here:
> 
>     1. Disable it
> 
>     2. Mark as wont fix in code climate
> 
> 
> I kinda like option 2 because having multiple returns could be a sign of code smell, or a function that needs to be split up.

Let's go with option-2. Even i like it better. 

@kaichaosun , please revert the codeclimate check change and then you are good to merge this PR.

PR #937, opened 2023-11-29
refactor: reuse query and cursor instead of specifying each pb field

richard-ramos:
# Description
While debugging store queries with @Ivansete-status, i realized that specifying each field for subsequent queries is not a good idea because maybe the protocol will change in the future, and we might end up missing attributes here. With this change we reuse the query fields and the cursor


chaitanyaprem (approved):
LGTM

PR #940, opened 2023-11-30
feat: fill msg timestamp while appending RLN proof it is not set

### LINKED ISSUE #926: feat: use current timestamp when unset in rest api
### Problem

When publishing a message via the REST API without setting a timestamp, default to the current one instead of 0 when RLN is enabled.

 curl -X POST "http://127.0.0.1:8645/relay/v1/auto/messages" \
 -H "content-type: application/json" \
 -d '{"payload":"XQ==","contentTopic":"/my-app/2/chatroom-1/proto"}

Note: Also need to make this mandatory in Relay publish API if RLN is enabled, else need to set it to currentTimestamp.

chaitanyaprem:
# Description

If message timestamp is not set, then as part of appending RLN proof to the message, the timestamp is also set based on which the RLN proof is calculated.

Fixes #926 and not just REST API but for all API's.

cc @waku-org/nwaku-developers This would have to be addressed in nwaku REST API as well.

### DIFF
--- a/waku/v2/protocol/rln/waku_rln_relay.go
+++ b/waku/v2/protocol/rln/waku_rln_relay.go
@@ -217,6 +217,10 @@ func (rlnRelay *WakuRLNRelay) AppendRLNProof(msg *pb.WakuMessage, senderEpochTim
 	}
 
 	msg.RateLimitProof = b
+	//If msgTimeStamp is not set, then set it to timestamp of proof
+	if msg.Timestamp == nil {
+		msg.Timestamp = proto.Int64(senderEpochTime.Unix())
+	}
 
 	return nil
 }


richard-ramos:
The code looks okay, but I'm wondering why is it necessary to set the timestamp in the message? 
AFAIK it's not used in the RLN validator, but instead the validation is done against the epoch attribute part of the RLN Proof

chaitanyaprem:
> The code looks okay, but I'm wondering why is it necessary to set the timestamp in the message? AFAIK it's not used in the RLN validator, but instead the validation is done against the epoch attribute part of the RLN Proof

True, but for Store queries to be consistent across nodes having a senderTimestamp helps with it.
Also this is already being tracked as an issue in nwaku at https://github.com/waku-org/nwaku/issues/2193

More context regarding this can be found at https://discord.com/channels/1110799176264056863/1111540684575477821/1177574753100378123 as further improvements in store protocol also expects timestamp to be part of message. Hence it would help making it mandatory.

richard-ramos (approved):

PR #942, opened 2023-11-30
feat: make `criteriaFN` public

richard-ramos:
# Description
This change is required to be able to override the default factory store


chaitanyaprem (approved):
LGTM

PR #944, opened 2023-12-05
fix(store): set pagesize behavior match nwaku

richard-ramos:
# Description
While chatting with @igorsirotin, he pointed out the difference in behavior in storenode implementations between go-waku and nwaku. This PR fixes this difference


chaitanyaprem (approved):
LGTM

PR #945, opened 2023-12-05
fix: take into account the ENR seq to determine if a peer is new or not

richard-ramos:
# Description
We were assuming that if a peer already existed on the store, it was already seen, regardless of the ENR being updated or not.


chaitanyaprem (approved):
LGTM

only one thought though, will addpeer override the existing peer record or do we have to remove and re add for it to take new param support

richard-ramos:
> only one thought though, will addpeer override the existing peer record or do we have to remove and re add for it to take new param support

It will override the peer info

PR #948, opened 2023-12-06
fix: number of connected peers per topic

richard-ramos:
# Description
Make sure that the number of connected peers by topic is correct 
(I was not able to reproduce the issue reported on which after 8hrs or so, no new connections happen, but this change seems to improve things?)


chaitanyaprem (approved):
LGTM

PR #950, opened 2023-12-08
Don't merge: Added getmessages endpoint

t-one-000:
The report is attached as a pdf:
[WakuMessageRetrievalResponseAnalysisReport.pdf](https://github.com/waku-org/go-waku/files/13608980/WakuMessageRetrievalResponseAnalysisReport.pdf)

I couldn't figure out how to access the Query method in the persistence/store class so I had to hardcode a connection to the specific DB to instantiate a DBStore.

### DIFF
--- a/cmd/waku/server/rest/getmessages.go
+++ b/cmd/waku/server/rest/getmessages.go
@@ -1,0 +1,67 @@
+package rest
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/go-chi/chi/v5"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/waku-org/go-waku/waku/persistence"
+	"github.com/waku-org/go-waku/waku/persistence/sqlite"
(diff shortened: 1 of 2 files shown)


chaitanyaprem:
> The report is attached as a pdf: [WakuMessageRetrievalResponseAnalysisReport.pdf](https://github.com/waku-org/go-waku/files/13608980/WakuMessageRetrievalResponseAnalysisReport.pdf)
> 
> I couldn't figure out how to access the Query method in the persistence/store class so I had to hardcode a connection to the specific DB to instantiate a DBStore.

You could have just used the existing [REST API](https://github.com/waku-org/go-waku/blob/c403388ec2146eb93155ee47f92f8eb030190857/cmd/waku/server/rest/store_api.yaml#L14-L16) to fetch messages from Store. 

No need to have this new API. 

Also, would like to know where you are trying to use go-waku.
You can hop in to our [discord](https://discord.waku.org/) if you have any queries.

chaitanyaprem:
Closing this PR as this should not be merged.

PR #952, opened 2023-12-09
chore: intro go workspaces, upgrade to go1.21

Nasfame:
# Description

# Changes

- [ ] ignore ide configs
- [ ] intro go.work
- [ ] upgrade go1.21 tool chain [quick go support]

# Tests


go.work
Nasfame, line 1: Has to check if toolchain directive is supported in go workspaces. I think not 🚫

PR #964, opened 2023-12-18
chore: stop light filter node by default

kaichaosun:
# Description

Related to issue https://github.com/status-im/status-go/issues/4459

# Changes

- stop filter light node by default
- run unsubscribeAll in goroutine

### DIFF
--- a/waku/v2/protocol/filter/client.go
+++ b/waku/v2/protocol/filter/client.go
@@ -113,7 +113,9 @@ func (wf *WakuFilterLightNode) Stop() {
 		wf.h.RemoveStreamHandler(FilterPushID_v20beta1)
 		if wf.subscriptions.Count() > 0 {
 			go func() {
-				defer func() { recover() }()
+				defer func() {
+					_ = recover()
+				}()
 				res, err := wf.unsubscribeAll(wf.Context())
 				if err != nil {
 					wf.log.Warn("unsubscribing from full nodes", zap.Error(err))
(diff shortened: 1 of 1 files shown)


waku/v2/protocol/filter/client.go
```diff
@@ -112,19 +112,20 @@ func (wf *WakuFilterLightNode) Stop() {
...
 		if wf.subscriptions.Count() > 0 {
-			res, err := wf.unsubscribeAll(wf.Context())
-			if err != nil {
-				wf.log.Warn("unsubscribing from full nodes", zap.Error(err))
-			}
-
-			for _, r := range res.Errors() {
-				if r.Err != nil {
-					wf.log.Warn("unsubscribing from full nodes", zap.Error(r.Err), logging.HostID("peerID", r.PeerID))
+			go func() {
```
richard-ramos, line 115: wouldn't it be possible for this go-routine to panic in status-go if the wakunode is garbage collected in the logout before `wf.unsubscribeAll()` finishes execution?  Perhaps we should use a `recover()` just in case.

cc: @cammellos, because he's much more knowledgeable about panics and goroutines!

richard-ramos: This has the drawback of making the logout process in status-go take longer than usual, which ends up causing a bad UX (the user would have to wait until status-go unsubscribes from the filter nodes before being able to logout). 

This is mainly the reason why the filter client was not stopped in the first place, and left to the full nodes to drop the connection eventually

PR #965, opened 2023-12-18
chore: use nwaku's db format

richard-ramos:
The only issue is that once you 'migrate' from nwaku to go-waku, the PRAGMA schema.user_version wont be used anymore. Probably i'll introduce a version table in nwaku to avoid this


waku/persistence/store.go
```diff
@@ -300,17 +300,20 @@ func (d *DBStore) Validate(env *protocol.Envelope) error {
...
+	stmt, err := d.db.Prepare("INSERT INTO message (id, messageHash, storedAt, timestamp, contentTopic, pubsubTopic, payload, version) VALUES ($1, $2, $3, $4, $5, $6, $7, $8)")
 	if err != nil {
 		d.metrics.RecordError(insertFailure)
 		return err
 	}
 
-	cursor := env.Index()
-	dbKey := NewDBKey(uint64(cursor.SenderTime), uint64(cursor.ReceiverTime), env.PubsubTopic(), env.Index().Digest)
+	storedAt := env.Message().GetTimestamp()
+	if storedAt == 0 {
```
chaitanyaprem, line 311: Why can't storedAt always be receiverTimestamp? 
Since msgTimestamp is anyways stored separately, wouldn't it be better to use this field for time at which message is received.

Also, wondering if this field is really required since store query should ideally only rely on senderTimestamp for better query response gaurantees across different store nodes. Maybe i am missing something. Why was the receiverTime included in store query at all since different nodes in the network can receive message at differnt times (considering granularity of 1 sec).

waku/persistence/sqlite/nwaku.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"database/sql"
+	"errors"
+	"fmt"
+)
+
+const minSupportedNWAKUversion = 8
+const maxSupportedNWAKUversion = 8
+
+func handleNWakuPreMigration(db *sql.DB) (bool, error) {
```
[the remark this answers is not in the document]
chaitanyaprem: If i am not wrong, this code is for scenarios where go-waku is running as service-node and not used as a client SDK(as in status-desktop). 
If someone is running an nwaku node and wants to migrate to go-waku node, then this would be useful as the database need not be cleaned.
@richard-ramos can confirm

waku/persistence/sqlite/nwaku.go
```diff
@@ -0,0 +1,82 @@
...
+	if err != nil && errors.Is(err, sql.ErrNoRows) {
+		return false, fmt.Errorf("could not obtain schema_migrations data while attempting to migrate nwaku database: %w", err)
+	}
+
+	if nwakuDBVersion == 0 {
+		// not a nwaku db
+		return false, nil
+	}
+
+	if nwakuDBVersion < minSupportedNWAKUversion || nwakuDBVersion > maxSupportedNWAKUversion {
```
chaitanyaprem, line 33: Seeing there are a good number of conditions here, are these going to go away once nwaku also aligns with same db strategy as go-waku? 
If not, we may have to document somewhere that on noticing abc errors, please delete the db and continue.

chaitanyaprem (approved):
LGTM

Left some minor comments.

waku/persistence/sqlite/nwaku.go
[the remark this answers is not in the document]
richard-ramos: Correct. It would be for such cases. Although ideally they would migrate to postgresql instead.

waku/persistence/store.go
richard-ramos: The message timestamp is an optional value so we can't rely on it for queries. I agree that the way it's done is confusing, in part because the `storedAt` column shouldnt be called that, because the data it contains does not refer just to the receiver timestamp, but also to the sender timestamp as seen in nwaku here:  https://github.com/waku-org/nwaku/blob/ebad0385ef61c55cf0699a114fb64cb6e88d9680/waku/waku_store/client.nim#L159-L160 in which the storedAt will contain the sender timestamp if available, otherwise the current timestamp.

Perhaps this is something that we could rename to something different? like `queryTimestamp` or dunno. (cc: @Ivansete-status @jm-clius)

Regarding the second point, `storedAt` is currently being used for retrieiving query results within a time range. See https://github.com/waku-org/nwaku/blob/ebad0385ef61c55cf0699a114fb64cb6e88d9680/waku/waku_archive/driver/sqlite_driver/queries.nim#L266-L274

PR #966, opened 2023-12-20
chore: add retry mechanism for rpc calls and rename db function 

richard-ramos:


chaitanyaprem (approved):
LGTM

Minor comment, why not just use `retry.Do` as we are not passing any data to internal functions?

richard-ramos:
Ah! it's because with retry.WithData I can return an additional value (the result of the RPC call), and not only the error,   i.e.:
```
func Do(retryableFunc RetryableFunc, opts ...Option) error
``` 
vs
```
func DoWithData[T any](retryableFunc RetryableFuncWithData[T], opts ...Option) (T, error)
```

PR #973, opened 2023-12-26
fix: use node's clusterId when adding peer from admin REST API & don't subscribe to default pubsub topic for non-zero clusterID

chaitanyaprem:
# Changes

- Admin REST API to add peer was using default clusterID rather than node's configured clusterID.
- If clusterID is set to non-zero, do not subscribe to default pubsubTopic `/waku/2/default-waku/proto`


richard-ramos (approved):
good catch!

PR #978, opened 2023-12-27
fix: ENR seq check to skip adding already existing peer

chaitanyaprem:
# Description
While running local go-waku node, had noticed that same peer was getting added again and again despite checks for duplicates and ENR updates.
Noticed that ENR update check was failing even if sequence number of ENR is not updated.

### DIFF
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -350,8 +350,9 @@ func (pm *PeerManager) AddDiscoveredPeer(p service.PeerData, connectNow bool) {
 	_, err := pm.host.Peerstore().(wps.WakuPeerstore).Origin(p.AddrInfo.ID)
 	if err == nil {
 		enr, err := pm.host.Peerstore().(wps.WakuPeerstore).ENR(p.AddrInfo.ID)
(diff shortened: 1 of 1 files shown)


waku/v2/peermanager/peer_manager.go
```diff
@@ -351,9 +351,11 @@ func (pm *PeerManager) AddDiscoveredPeer(p service.PeerData, connectNow bool) {
 	if err == nil {
 		enr, err := pm.host.Peerstore().(wps.WakuPeerstore).ENR(p.AddrInfo.ID)
 		// Verifying if the enr record is more recent (DiscV5 and peer exchange can return peers already seen)
-		if err == nil && enr.Record().Seq() > p.ENR.Seq() {
+		if err == nil && enr.Record().Seq() >= p.ENR.Seq() {
 			return
 		}
+		pm.logger.Info("peer already found in peerstore, but re-adding it as ENR sequence is higher than locally stored",
```
richard-ramos, line 357: I wonder if this should be a .Debug?

richard-ramos (approved):

waku/v2/peermanager/peer_manager.go
chaitanyaprem: Thought about it and left it as Info as this should not be a frequent occurence for peers. 
Do you see a scenario where this log becomes bothersome?

richard-ramos: Not sure. I know that go-waku does a enr seq update as soon as it obtains an external address. Let's keep it on Info and see how it behaves :)

PR #985, opened 2024-01-03
fix: return error while serializing rln metadata in case chainID is 0

chaitanyaprem:


richard-ramos (commented):
I'm curious about this change. Why isn't it possible to use RLN with chainID=0?

chaitanyaprem:
> I'm curious about this change. Why isn't it possible to use RLN with chainID=0?

I was under the assumption that chainID 0 is not valid and definitely state of RLN should always be associated with a chain. 
Is there a case where this is not associated with a chain?

richard-ramos:
Ah you're right. I confused ChainID with ClusterID! sorry!

richard-ramos (approved):

Issue #819, opened 2023-10-19
chore: create docs for setting up a systemd service and exit codes


richard-ramos:

richard-ramos:
**Weekly Update**

- _achieved_: Added systemd template and docs
- _next_: Add env variables suggested by infra

chaitanyaprem:
Closing the issue as it is descoped for go-waku.

PR #989, opened 2024-01-03
feat: systemd

richard-ramos:
# Description
Adds docs on how to setup a systemd service, and also adds systemd configuration to .deb/.rpm
Note: since there's no `--rpm-systemd` flag in `fpm`, I ended up creating a .deb package with `fpm`, and later converted it to .rpm with the same tool

Fixes https://github.com/waku-org/go-waku/issues/819


scripts/linux/waku.service
```diff
@@ -0,0 +1,44 @@
...
+# List of non-recoverable error codes
+RestartPreventExitStatus=166
+
+ExecStart=/usr/bin/waku \
+  --cluster-id=${CLUSTER_ID} \
+  --port=${TCP_PORT} \
+  --rest=${REST_ENABLED} \
+  --rest-port=${REST_PORT} \
+  --metrics=${METRICS_ENABLED} \
+  --metrics-port=${METRICS_PORT}
```
richard-ramos, line 44: @chaitanyaprem I'm not sure what flags to use here by default. Should I use the same from go-waku-compose ?

chaitanyaprem: go-waku-compose is meant for TWN so cluster-id is mentioned as 1. 
But for status since they use static sharding cluster-id is 16. So maybe we can indicate that as default and add a note to change it accordingly. 

Wrt rest, we better leave it enabled as it is useful to query info from a node. 
Metrics also should be better enabled by default. 

Wrt rest and metrics ports, I am wondering if it is better to not specify and leave them as default. Since this is required by Infra team, maybe better to check with them.

richard-ramos: @jakubgs what would you like to see in the systemd template for go-waku?

jakubgs: You can use our docker compose as a starting point for what flags to include:
https://github.com/status-im/infra-role-nim-waku/blob/master/templates/docker-compose.yml.j2

As for systemd service config, you can take a look at our beacon node one:
https://github.com/status-im/infra-role-beacon-node-linux/blob/master/templates/beacon-node.service.j2

And include fields like `Documentation`, `Restart`, and `EnvironmentFile`.
With `EnvironmentFile` you could put the defaults under `/etc` and allow the user to easily modify them.

jakubgs:
Are we sure we want to lose this work? Seems useful to me.

PR #994, opened 2024-01-04
feat: parameterizable number of connections per IP

richard-ramos:
# Description
Matches https://github.com/waku-org/nwaku/pull/2323

### DIFF
--- a/waku/v2/peermanager/connection_gater.go
+++ b/waku/v2/peermanager/connection_gater.go
@@ -16,18 +16,20 @@ import (
 // the number of connections per IP address
 type ConnectionGater struct {
 	sync.Mutex
-	logger  *zap.Logger
-	limiter map[string]int
+	logger        *zap.Logger
+	limiter       map[string]int
+	maxConnsPerIP int
 }
 
-const maxConnsPerIP = 10
-
 // NewConnectionGater creates a new instance of ConnectionGater
-func NewConnectionGater(logger *zap.Logger) *ConnectionGater {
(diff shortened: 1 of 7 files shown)


waku/v2/peermanager/connection_gater.go
```diff
@@ -103,7 +103,7 @@ func (c *ConnectionGater) validateInboundConn(addr multiaddr.Multiaddr) bool {
 	c.Lock()
 	defer c.Unlock()
 
-	if currConnections := c.limiter[ip.String()]; currConnections+1 > maxConnsPerIP {
+	if currConnections := c.limiter[ip.String()]; c.maxConnsPerIP > 0 && currConnections+1 > c.maxConnsPerIP {
 		return false
```
chaitanyaprem, line 107: Wondering if we ought to add some info/debug log here to indicate we are limiting connections as per config.

chaitanyaprem (approved):
LGTM, left a minor comment.

chaitanyaprem:
> The IP based rate limit seems better apply to store (maybe also relay) protocol, as it mostly deploys in a server. For other protocols, when used in client app like status, the user usually behind a NAT or VPN, which means same IP assigns to many users. Wondering what's the best practice for such scenarios?

Here the limit is for any peer that is connecting to the node ir-respective of the protocol involved. This is to ensure that someone doesn't spawn too many peers in order to attack a node or mislead/coerce a node.

Wrt your query of nodes running behind a NAT or VPN, it is still fine. Because there would be other nodes in the network you can connect to as long as the network is sufficiently decentralized and has lot of nodes. This just prevents from more than `n` nodes with same IP connecting to same peer.

Issue #988, opened 2024-01-03
bug/regression: invalid memory address or nil pointer dereference when trying to connect nodes


fbarbu15:
This is a regression as it didn't happen last week

**To Reproduce**
1. Start 1st node (relay + filter): `['--listen-address=0.0.0.0', '--rpc=true', '--rpc-admin=true', '--rest=true', '--rest-admin=true', '--websocket-support=true', '--log-level=DEBUG', '--rest-relay-cache-capacity=100', '--websocket-port=50764', '--rpc-port=50762', '--rest-port=50761', '--tcp-port=50763', '--discv5-udp-port=50765', '--rpc-address=0.0.0.0', '--rest-address=0.0.0.0', '--nat=extip:172.18.163.70', '--peer-exchange=true', '--discv5-discovery=true', '--cluster-id=0', '--min-relay-peers-to-publish=0', '--legacy-filter=false', '--rest-filter-cache-capacity=50', '--relay=true', '--filter=true', '--nodekey=30348dd51465150e04a5d9d932c72864c8967f806cce60b5d26afeca1e77eb68']`
[...]

PR #996, opened 2024-01-05
fix: use both filter and legacyFilter flags to determine if filter is enabled and set Waku2 ENR field

chaitanyaprem:
Florin ran into an issue which has 2-fold solution. 

As per https://github.com/waku-org/go-waku/issues/988#issuecomment-1876810340 , the Filter-Subscribe REST API was timing out in node2 eventhough it is connected to node1(which has Filterv2 enabled).

But since node1's ENR is not settings Filter bit in Waku2 field, node2 doesn't consider node1 as filter node as it is only found via discovery.

Second problem is his request times-out because node2 goes into on-demand discovery and does till REST API times out.
This will have to improved for dev-ex, by having an internal timeout for on-demand discovery and responding back with error.

### DIFF
--- a/waku/v2/node/wakunode2.go
+++ b/waku/v2/node/wakunode2.go
@@ -200,7 +200,7 @@ func New(opts ...WakuNodeOption) (*WakuNode, error) {
 	w.log = params.logger.Named("node2")
 	w.wg = &sync.WaitGroup{}
 	w.keepAliveFails = make(map[peer.ID]int)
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):
Change looks good to me!, however I'd go even further and just use w.opts.enableFilterFullNode since we should think of dropping legacy filter support, as discussed here: https://discord.com/channels/1110799176264056863/1192407282420551742/1192408493790085240 

However, we can go ahead and merge this, since once legacy filter is removed, this flag will be gone as well :)

PR #999, opened 2024-01-05
feat: ping cache

richard-ramos:
# Description
Cache that contains the ping times for peers, for reusing the value instead of pinging a peer each time you want to select the fastest peer.  Ping times are verified every 10 seconds.

### DIFF
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -84,6 +84,7 @@ type PeerManager struct {
 	discoveryService       *discv5.DiscoveryV5
 	wakuprotoToENRFieldMap map[protocol.ID]WakuProtoInfo
 	TopicHealthNotifCh     chan<- TopicHealthStatus
+	rttCache               *FastestPeerSelector
 }
 
 // PeerSelection provides various options based on which Peer is selected from a list of peers.
@@ -188,6 +189,7 @@ func NewPeerManager(maxConnections int, maxPeers int, metadata *metadata.WakuMet
 		subRelayTopics:         make(map[string]*NodeTopicDetails),
 		maxPeers:               maxPeers,
 		wakuprotoToENRFieldMap: map[protocol.ID]WakuProtoInfo{},
+		rttCache:               NewFastestPeerSelector(logger),
 	}
 	logger.Info("PeerManager init values", zap.Int("maxConnections", maxConnections),
 		zap.Int("maxRelayPeers", maxRelayPeers),
@@ -206,6 +208,7 @@ func (pm *PeerManager) SetDiscv5(discv5 *discv5.DiscoveryV5) {
 // SetHost sets the host to be used in order to access the peerStore.
 func (pm *PeerManager) SetHost(host host.Host) {
 	pm.host = host
+	pm.rttCache.SetHost(host)
 }
 
 // SetPeerConnector sets the peer connector to be used for establishing relay connections.
@@ -215,7 +218,6 @@ func (pm *PeerManager) SetPeerConnector(pc *PeerConnectionStrategy) {
 
 // Start starts the processing to be done by peer manager.
 func (pm *PeerManager) Start(ctx context.Context) {
-
 	pm.RegisterWakuProtocol(relay.WakuRelayID_v200, relay.WakuRelayENRField)
 
 	pm.ctx = ctx
--- a/go.sum
+++ b/go.sum
@@ -564,7 +564,6 @@ github.com/go-latex/latex v0.0.0-20210118124228-b3d85cf34e07/go.mod h1:CO1AlKB2C
 github.com/go-logfmt/logfmt v0.3.0/go.mod h1:Qt1PoO58o5twSAckw1HlFXLmHsOX5/0LbT9GBnD5lWE=
 github.com/go-logfmt/logfmt v0.4.0/go.mod h1:3RMwSq7FuexP4Kalkev3ejPJsZTpXXBr9+V4qmtdjCk=
 github.com/go-logfmt/logfmt v0.5.0/go.mod h1:wCYkCAKZfumFQihp8CzCvQ3paCTfi41vtzG1KdI/P7A=
-github.com/go-logfmt/logfmt v0.5.1/go.mod h1:WYhtIu8zTZfxdn5+rREduYbwxfcBr/Vr6KEVveWlfTs=
 github.com/go-logr/logr v0.1.0/go.mod h1:ixOQHD9gLJUVQQ2ZOR7zLEifBX6tGkNJF4QyIY7sIas=
 github.com/go-logr/logr v0.2.0/go.mod h1:z6/tIYblkpsD+a4lm/fGIIU9mZ+XfAiaFtq7xTgseGU=
 github.com/go-logr/logr v0.4.0/go.mod h1:z6/tIYblkpsD+a4lm/fGIIU9mZ+XfAiaFtq7xTgseGU=
--- a/waku/v2/node/wakuoptions.go
+++ b/waku/v2/node/wakuoptions.go
@@ -12,7 +12,6 @@ import (
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/libp2p/go-libp2p"
-	mplex "github.com/libp2p/go-libp2p-mplex"
 	pubsub "github.com/libp2p/go-libp2p-pubsub"
 	"github.com/libp2p/go-libp2p/config"
 	"github.com/libp2p/go-libp2p/core/crypto"
@@ -559,7 +558,6 @@ var DefaultLibP2POptions = []libp2p.Option{
 	libp2p.UserAgent(UserAgent),
 	libp2p.ChainOptions(
 		libp2p.Muxer("/yamux/1.0.0", yamux.DefaultTransport),
-		libp2p.Muxer("/mplex/6.7.0", mplex.DefaultTransport),
 	),
 	libp2p.EnableNATService(),
 	libp2p.ConnectionManager(newConnManager(200, 300, connmgr.WithGracePeriod(0))),
--- a/waku/v2/peermanager/fastest_peer_selector.go
+++ b/waku/v2/peermanager/fastest_peer_selector.go
@@ -1,0 +1,172 @@
+package peermanager
+
+import (
+	"context"
+	"errors"
+	"sort"
+	"sync"
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/host"
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
+	"github.com/waku-org/go-waku/logging"
+	"go.uber.org/zap"
+)
+
+type FastestPeerSelector struct {
+	sync.RWMutex
+
+	host host.Host
+
+	logger *zap.Logger
+}
+
+func NewFastestPeerSelector(logger *zap.Logger) *FastestPeerSelector {
+	return &FastestPeerSelector{
+		logger: logger.Named("rtt-cache"),
+	}
+}
+
+func (r *FastestPeerSelector) SetHost(h host.Host) {
+	r.host = h
+}
+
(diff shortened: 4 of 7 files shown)


waku/v2/peermanager/peer_manager.go
```diff
@@ -106,6 +109,8 @@ func NewPeerManager(maxConnections int, maxPeers int, logger *zap.Logger) *PeerM
 		subRelayTopics:         make(map[string]*NodeTopicDetails),
 		maxPeers:               maxPeers,
 		wakuprotoToENRFieldMap: map[protocol.ID]WakuProtoInfo{},
+		timesource:             timesource,
+		rttCache:               NewRTTCache(timesource, logger),
```
chaitanyaprem, line 113: I am wondering if a timesource is not passed, should we not initialize this cache and not have fastest selection feature or control this by a config option? This maybe useful for someone who doesn't want constant pings done to peers.

waku/v2/peermanager/rtt_cache.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (r *RTTCache) Start(ctx context.Context) {
+	go r.start(ctx, 10*time.Second)
+}
+
+func (r *RTTCache) pingPeers(ctx context.Context) {
+	r.RLock()
+	var toPing peer.IDSlice
+	if len(r.pingQueue) == 0 {
+		toPing = r.host.Peerstore().Peers()
```
chaitanyaprem, line 68: Shouldn't we only use connectedPeers , because otherwise the ping may open a connection to the peer un-necessarily.

waku/v2/peermanager/peer_manager.go
chaitanyaprem: Especially for a serviceNode use-case, we may not want to start this RTT Cache itself.

waku/v2/peermanager/rtt_cache.go
```diff
@@ -0,0 +1,256 @@
...
+	} else {
+		for _, p := range r.pingQueue {
+			if p.IsStale() {
+				toPing = append(toPing, p.PeerID)
```
chaitanyaprem, line 72: We may also have to handle removal of peers that are no longer in peerStore or peers that were previously connected, but not disconnected.

waku/v2/peermanager/rtt_cache.go
```diff
@@ -0,0 +1,256 @@
...
+
+type RTTCache struct {
+	sync.RWMutex
+
+	host host.Host
+
+	peers     map[peer.ID]*PeerRTT
+	pingQueue PingQueue
+
+	verificationInterval time.Duration
```
chaitanyaprem, line 42: Shouldn't this be set somewhere?

waku/v2/peermanager/rtt_cache.go
```diff
@@ -0,0 +1,256 @@
...
+		peers:      make(map[peer.ID]*PeerRTT),
+	}
+}
+
+func (r *RTTCache) SetHost(h host.Host) {
+	r.host = h
+}
+
+func (r *RTTCache) Start(ctx context.Context) {
+	go r.start(ctx, 10*time.Second)
```
chaitanyaprem, line 61: Shouldn't this be based on verificationInterval?

waku/v2/peermanager/rtt_cache.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (r *RTTCache) getOrDefault(peerID peer.ID) (*PeerRTT, bool) {
+	peerRTT, ok := r.peers[peerID]
+	exists := true
+
+	if !ok {
+
+		rngSrc := rand.NewSource(rand.Int63())
+		minBackoff, maxBackoff := time.Second*5, time.Hour
+		bkf := backoff.NewExponentialBackoff(minBackoff, maxBackoff, backoff.FullJitter, time.Second, 5.0, 0, rand.New(rngSrc))()
```
chaitanyaprem, line 198: Maybe the backoff growth could be based on verificationInterval rather than hrdcoded to 5secs?

waku/v2/peermanager/rtt_cache.go
```diff
@@ -0,0 +1,256 @@
...
+	case <-ctx.Done():
+		return nil, ctx.Err()
+
+	case result := <-ping.Ping(ctx, r.host, peer):
+		r.Lock()
+		defer r.Unlock()
+
+		peerRTT, exists := r.getOrDefault(peer)
+		if result.Error == nil {
+			peerRTT.PingRTT = result.RTT
```
chaitanyaprem, line 123: The code in if and else condition is pretty much similar except the values set and returned, maybe can make this into an function to avoid duplication.

waku/v2/peermanager/rtt_cache.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (r *RTTCache) PeersRTT(ctx context.Context, peers peer.IDSlice) []*PeerRTT {
+	wg := sync.WaitGroup{}
+	pingCh := make(chan *PeerRTT, len(peers))
+	wg.Add(len(peers))
+	for _, p := range peers {
+		go func(p peer.ID) {
+			defer wg.Done()
+			peerRTT, err := r.PingPeer(ctx, p)
```
chaitanyaprem, line 154: Wondering if we need to ping the peers during selection, since we have a periodic ping going on which should update the RTT for the peers. This may cause problems even if 1 peer is stale. 
Wouldn't it be better to be optimistic and return the peer with shortest ping for which we already have RTT recorded?

waku/v2/peermanager/rtt_cache.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (r *RTTCache) FastestPeer(ctx context.Context, peers peer.IDSlice) (*PeerRTT, error) {
+	var min *PeerRTT
+	for _, p := range r.PeersRTT(ctx, peers) {
+		if min == nil {
+			min = p
+		} else {
+			if p.PingRTT < min.PingRTT {
+				min = p
```
chaitanyaprem, line 179: Confused here as to why min has to be checked, I am assuming we are using a heap and can sort the peers in the heap itself based on ping times arrived at during regular pings itself.
I maybe missing something here though.

chaitanyaprem:
Another thought comes to my mind is since messages would anyways be sent/received from Filter node either via FilterPing or MessagePush. Can we somehow also use those messages RTT (from transport layer) and in that case Ping would only get used in idle periods. 

Makes me wonder if we are creating lot of traffic in the network just for RTT maintenance.
Need to do some more study regarding if this is something that libp2p provides.

waku/v2/peermanager/rtt_cache.go
richard-ramos: For sure it would simplify the code

richard-ramos:
> Can we somehow also use those messages RTT (from transport layer) and in that case Ping would only get used in idle periods.

Would be nice if libp2p had something to let us know how long it took to open a stream. I did try to find this unsuccesfully tho :(

waku/v2/peermanager/rtt_cache.go
richard-ramos: But aren't we're interested in pinging all the peers to know their reply time regardless of them being connected or not?

waku/v2/peermanager/rtt_cache.go
richard-ramos: I ended up checking if the peer to ping has info in the peerStore before adding it to `toPing`

richard-ramos:
@chaitanyaprem,  I found something interesting in yamux!!:

Looks like it measures RTT for all the open sessions:
- https://github.com/libp2p/go-yamux/blob/48aa3a7bb46c354076969f5b3ac7f4b805744dcb/session.go#L164C1-L164C24
- https://github.com/libp2p/go-yamux/blob/48aa3a7bb46c354076969f5b3ac7f4b805744dcb/session.go#L340
meaning that instead of doing a Ping ourselves, we can use `Peerstore().LatencyEWMA(peer.ID) time.Duration` to obtain an exponentially-weighted moving avg. I think we can use this, since we're dropping support mplex!  (we would still need to ping those peers that do not have a latency since it means we have never pinged them before).

There's also a keepAlive with a configurable interval: 
https://github.com/libp2p/go-yamux/blob/48aa3a7bb46c354076969f5b3ac7f4b805744dcb/session.go#L430C1-L430C37
which i seems interesting too because maybe I can use this instead of the keepAlive code in go-waku to handle session connectivity

waku/v2/peermanager/rtt_cache.go
chaitanyaprem: Yeah, but we keep discovered peers in the store in general ir-respective of whether we connect to them or not. Because we keep pruning the connections for relay if we end up connecting to more, but at the same time they still exist in the peerStore for future use, in case a new shard subscription occurs or the peers we are connected to disconnect from us. 
Hence, i was wondering if we can only do this ping for peers that we are already connected to. Although, if the application wants to select new peers for a service, then we may not have peers with best RTT to select. 
Wondering how we can get around that.

And at the same time, it will not be a good idea to keep connecting to peers just to get their RTT and not use the connections.

waku/v2/peermanager/rtt_cache.go
chaitanyaprem: Then let's keep it simple then and not do ping during selection :)

chaitanyaprem:
> @chaitanyaprem, I found something interesting in yamux!!:
> 
> Looks like it measures RTT for all the open sessions:
> 
>     * https://github.com/libp2p/go-yamux/blob/48aa3a7bb46c354076969f5b3ac7f4b805744dcb/session.go#L164C1-L164C24
> 
>     * https://github.com/libp2p/go-yamux/blob/48aa3a7bb46c354076969f5b3ac7f4b805744dcb/session.go#L340
>       meaning that instead of doing a Ping ourselves, we can use `Peerstore().LatencyEWMA(peer.ID) time.Duration` to obtain an exponentially-weighted moving avg. I think we can use this, since we're dropping support mplex!  (we would still need to ping those peers that do not have a latency since it means we have never pinged them before).
> 
> 
> There's also a keepAlive with a configurable interval: https://github.com/libp2p/go-yamux/blob/48aa3a7bb46c354076969f5b3ac7f4b805744dcb/session.go#L430C1-L430C37 which i seems interesting too because maybe I can use this instead of the keepAlive code in go-waku to handle session connectivity
> 
> Seems to me that this should simplify the code quite a bit :)

That's awesome, definitely will simply a lot of overhead at our layer.
And since mplex is being dropped, it makes sense to use this.

waku/v2/peermanager/rtt_cache.go
chaitanyaprem: Ah.ok.

waku/v2/peermanager/rtt_cache.go
richard-ramos: I wonder if we should use instead an ICMP Ping?

richard-ramos:
I did a new version using the latency. It is more efficient than the current peer selector on `master` branch because it uses the latency metric recorded by yamux,  but still does a Ping() for those peers which have no latency recorded.

waku/v2/peermanager/rtt_cache.go
richard-ramos: I did a change in the code, which still requires doing the ping but the code is less complex now

waku/v2/peermanager/peer_manager.go
richard-ramos: This is no longer an issue since there's no periodic ping being executed now.

waku/v2/peermanager/rtt_cache.go
chaitanyaprem: > I wonder if we should use instead an ICMP Ping?

That is not a bad idea, but generally ICMP ping also is blocked by firewalls so we can't rely on it.

chaitanyaprem (approved):
LGTM

Left one comment

astronaut1712:
when can it be merged? version `v0.9.0` will be got error because `libp2p` has removed `muxer/mplex` package

richard-ramos:
Yamux is currently not used in nwaku. Let's wait for this PR to be merged https://github.com/waku-org/nwaku/pull/2397 and for the waku/status fleets to be updated with a version that includes this change.

chaitanyaprem:
@richard-ramos , time to merge this PR i guess. CI issues need to be fixed though
It would be useful for peer management for status.

PR #1008, opened 2024-01-16
chore: remove RPC server

richard-ramos:
# Description
Removes the JSON RPC functionality from go-waku


chaitanyaprem (approved):
LGTM

danisharora099:
hmm.. this might break our tests suite on js-waku

richard-ramos:
Could js-waku use REST instead?

danisharora099:
> Could js-waku use REST instead?

yes, we could do it -- what's the priority of this deprecation from go-waku side? 
@waku-org/js-waku-developers can discuss this and account for this change in the coming PM call on Wednesday

danisharora099:
@richard-ramos i think we can safely merge this PR considering js-waku does not do CI interop tests with go-waku anymore :)

weboko (approved):
`js-waku` stopped running pipelines against `go-waku` 😢

PR #1013, opened 2024-01-25
fix: update lastProcessedBlock even if no RLN membership event is present

chaitanyaprem:
Update lastProcessedBlock even if no RLN membership event is present or membership removed event is present.

This is an optimization to handle a scenario of not requiring to sync on restart from blocks which were already processed but not updated due to no events being processed.

### DIFF
--- a/waku/v2/protocol/rln/group_manager/dynamic/dynamic.go
+++ b/waku/v2/protocol/rln/group_manager/dynamic/dynamic.go
@@ -78,6 +78,9 @@ func (gm *DynamicGroupManager) handler(events []*contracts.RLNMemberRegistered, 
 			lastBlockProcessed = event.Raw.BlockNumber
 		}
 	}
+	if len(events) == 0 {
+		lastBlockProcessed = latestProcessBlock
+	}
 
 	err := gm.RemoveMembers(toRemoveTable)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


waku/v2/protocol/rln/group_manager/dynamic/dynamic.go
```diff
@@ -73,10 +73,9 @@ func (gm *DynamicGroupManager) handler(events []*contracts.RLNMemberRegistered,
 			}
 			eventsPerBlock = append(eventsPerBlock, event)
 			toInsertTable.Set(event.Raw.BlockNumber, eventsPerBlock)
-
-			if event.Raw.BlockNumber > lastBlockProcessed {
-				lastBlockProcessed = event.Raw.BlockNumber
-			}
+		}
+		if event.Raw.BlockNumber > lastBlockProcessed {
```
alrevuelta, line 77: Not sure this fully solves the issue. Imagine a case where blocks 1000 to 2000 are processed, but `events` is empty. In this case I think `lastBlockProcessed` will still be wrong.

I would say what matter here is
`gm.lastBlockProcessed = lastBlockProcessed`

When no events happens `latestProcessBlock` should be the latest processed block

chaitanyaprem: True, will take a relook at whole logic once.

chaitanyaprem: Addressed in [99a3510](https://github.com/waku-org/go-waku/pull/1013/commits/99a351057083ff2a7d186abcf922b9209f36a87c)

PR #1015, opened 2024-01-26
fix: minor issues in connectionstatus report 

chaitanyaprem:
While going through code noticed few changes that needed to be done in connection status reporting
- legacy_filter is used instead of filterV2 to report connectionStatus. 
- filter and store are not needed to report relay connection status
- Store Status need not be always connected to a store node, as in status-go connection is opened on demand.Hence we check in peerStore if a store node is present to indicate the same.

Also noticed one more thing, while going through status-go code which needs to be improved and has impact on different parts of the code from go-waku till status app. 
Created and issue for the same [here](https://github.com/status-im/status-go/issues/4628). Do give your comments @richard-ramos


waku/v2/node/connectedness.go
```diff
@@ -160,20 +161,38 @@ func (w *WakuNode) Status() (isOnline bool, hasHistory bool) {
...
+				break
+			}
+		}
 	}
 
 	if w.opts.enableFilterLightNode && !w.opts.enableRelay {
 		isOnline = hasLightPush && hasFilter
 	} else {
-		isOnline = hasRelay || hasLightPush && (hasStore || hasFilter)
+		isOnline = hasRelay
```
richard-ramos, line 195: This change makes sense to do now. The reason why it checked for store before was because we always kept a connection to a store node (this code was added particularly for status-go back then). Not sure why we checked for filter tho? Maybe @vitvly remembers!

waku/v2/node/connectedness.go
```diff
@@ -160,20 +161,38 @@ func (w *WakuNode) Status() (isOnline bool, hasHistory bool) {
...
+			for _, protocol := range protocols {
+				if !hasStore && protocol == store.StoreID_v20beta4 {
+					hasStore = true
+					break
+				}
+			}
+			if hasStore {
+				break
+			}
+		}
```
richard-ramos, line 189: We're actually not using the logic from line 170 to 189, right?  I'm thinking if we should remove that

richard-ramos (approved):

waku/v2/node/connectedness.go
richard-ramos: Also, i think 161-163 is not needed either

chaitanyaprem: > We're actually not using the logic from line 170 to 189, right? I'm thinking if we should remove that

I added them thinking that atleast if the peerStore has a storeNode we can indicate it. But yeah, we have on-demand discovery as well so it shouldn't matter i guess.

chaitanyaprem: > Also, i think 161-163 is not needed either

Hmm, true. Will remove them

richard-ramos:
Weird! this is approved already, yet it's asking me to approve 
image

richard-ramos (approved):

chaitanyaprem:
> Weird! this is approved already, yet it's asking me to approve image

I have sent a re-review request as i have updated some things. 
But now, i am wondering if this PR is even required as anyways we are working on https://github.com/waku-org/go-waku/pull/1027

Shall i go ahead and close this?

Issue #667, opened 2023-08-23
feat: Enforce service specific rate limits for Lightpush


chaitanyaprem:
### Problem

As discussed here  https://github.com/waku-org/go-waku/issues/594#issuecomment-1621665668 , we should enforce some rate limits for heavy protocols such as store. Otherwise a node can be DoSed by store requests. Example: only allow to serve x request/second, or bytes/second. Beyond that, temporally blacklist.

### Suggested solution

We can use go-libp2p scaling limit config. e.g:  limits set by go-libp2p for the default services: https://github.com/libp2p/go-libp2p/blob/master/limits.go#L15 . We can control maximum number of streams allocated to each service and memory.
Provide configuration for these rate-limits and good defaults as well.

### Alternatives considered
None

### Acceptance criteria
Validate by pumping service traffic and verifying rate limits are applied as configured.

chaitanyaprem:
Rate-limiting would be required for the following protocols:
    Store
    Filter
    Lightpush
    Peer exchange
    Rendezvous


@jm-clius , @alrevuelta  Since RLN would be enabled, I don't think we would require to enforce additional rate-limiting on lightPush. Do confirm this.

This task would require analyzing what sort of limits are to be applied, req/sec or max streams, max connections etc. We should also need to come up with required configurations that we want to expose to users.

@jm-clius : Is this something that is required for the launch of The Waku network.

fryorcraken:
RLN rules need to be applied on light push on server side so that the node relaying doesn't get descored on gossipsub for forwarding messages with incorrect or no proof (when RLN applies)

chaitanyaprem:
> RLN rules need to be applied on light push on server side so that the node relaying doesn't get descored on gossipsub for forwarding messages with incorrect or no proof (when RLN applies)

Makes sense. cc @richard-ramos in case it is not already being done. We can take this up as part of #655?

chaitanyaprem:
> I'd say yes, at least in a basic form. One thing here is that we may not want to make this overly configurable at first. Could choose some value (e.g. max 5 requests per second) and design something simply around that.

Agreed, we can start simple without any configurationsby enforcing below 2 types of limits.Do provide any comments.
1. Number of connections/peers per service. (This can be derived as servicePeers/no of services.).
2. Allowed msgRate from a peer per service (this can be set by default as 5 req/sec).


[...]

PR #1024, opened 2024-02-01
chore: add rate limiter option to lightpush

richard-ramos:
# Description
Adds an optional rate limit (that we should set in status-go).
Fixes https://github.com/waku-org/go-waku/issues/667

### DIFF
--- a/examples/chat2/chat.go
+++ b/examples/chat2/chat.go
@@ -324,7 +324,7 @@ func (c *Chat) publish(ctx context.Context, message string) error {
 	}
 
 	if c.options.LightPush.Enable {
-		lightOpt := []lightpush.Option{lightpush.WithDefaultPubsubTopic()}
+		lightOpt := []lightpush.RequestOption{lightpush.WithDefaultPubsubTopic()}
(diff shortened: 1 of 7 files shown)


chaitanyaprem (approved):
LGTM

chaitanyaprem:
I am wondering if we should also provide an option to specify rate-limit config like WithRateLimit(<req/sec>) itself for cross language scenarios where limiter cannot be passed from application

richard-ramos:
> cross language scenarios
you mean, interop with other languages via bindings? if that's the case, it's a nice idea and worth adding to nwaku c-bindings.

chaitanyaprem:
> > cross language scenarios
> 
> you mean, interop with other languages via bindings? if that's the case, it's a nice idea and worth adding to nwaku c-bindings.

Yup, but i am thinking it would be better to add it to the golang API as well so that it would be uniform across all language bindings. WDYT?

richard-ramos:
Not sure! I'm wary of adding new code to go-waku that is not going to be used, since we're replacing it by nwaku

PR #1025, opened 2024-02-01
feat: add support for multiple public keys per topic

### LINKED ISSUE #1022: feat: Support message verification using multiple public keys in same pubsubTopic
# Problem

As of now, in status, a public key is used to verify all messages received on a topic in order to determine if they are valid or not. 

But if a single shard/pubsubTopic has to be shared by multiple communities, we need to support registering multiple keys which would be used sequentially to verify or validate messages and consider them as valid even if one of the keys work.

# Acceptance Criteria

Have tests to verify multiple keys being used for validation.

# Notes

Need to analyze performance impact for using multiple keys.

richard-ramos:
Fixes #1022

### DIFF
--- a/waku/v2/protocol/relay/validators.go
+++ b/waku/v2/protocol/relay/validators.go
@@ -85,10 +85,16 @@ func (w *WakuRelay) topicValidator(topic string) func(ctx context.Context, peerI
 }
 
 // AddSignedTopicValidator registers a gossipsub validator for a topic which will check that messages Meta field contains a valid ECDSA signature for the specified pubsub topic. This is used as a DoS prevention mechanism
-func (w *WakuRelay) AddSignedTopicValidator(topic string, publicKey *ecdsa.PublicKey) error {
-	w.log.Info("adding validator to signed topic", zap.String("topic", topic), zap.String("publicKey", hex.EncodeToString(elliptic.Marshal(publicKey.Curve, publicKey.X, publicKey.Y))))
+func (w *WakuRelay) AddSignedTopicValidator(topic string, publicKeys []*ecdsa.PublicKey) error {
 
-	fn := signedTopicBuilder(w.timesource, publicKey)
+	var hexPubKeys []string
+	for _, publicKey := range publicKeys {
+		hexPubKey := hex.EncodeToString(elliptic.Marshal(publicKey.Curve, publicKey.X, publicKey.Y))
+		hexPubKeys = append(hexPubKeys, hexPubKey)
+	}
+	w.log.Info("adding validator to signed topic", zap.String("topic", topic), zap.Strings("publicKeys", hexPubKeys))
+
+	fn := signedTopicBuilder(w.timesource, publicKeys)
 
 	w.RegisterTopicValidator(topic, fn)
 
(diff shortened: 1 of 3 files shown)


waku/v2/protocol/relay/validators.go
```diff
@@ -85,10 +85,16 @@ func (w *WakuRelay) topicValidator(topic string) func(ctx context.Context, peerI
 }
 
 // AddSignedTopicValidator registers a gossipsub validator for a topic which will check that messages Meta field contains a valid ECDSA signature for the specified pubsub topic. This is used as a DoS prevention mechanism
-func (w *WakuRelay) AddSignedTopicValidator(topic string, publicKey *ecdsa.PublicKey) error {
-	w.log.Info("adding validator to signed topic", zap.String("topic", topic), zap.String("publicKey", hex.EncodeToString(elliptic.Marshal(publicKey.Curve, publicKey.X, publicKey.Y))))
+func (w *WakuRelay) AddSignedTopicValidator(topic string, publicKeys []*ecdsa.PublicKey) error {
```
chaitanyaprem, line 88: I am wondering if we should rename it to `SetSignedTopicValidator` and only have 1 signedValidator per topic? 
I am thinking of a scenario where this function gets called twice for same topic from application, then validation could fail for one and not for the others which defeats the purpose of supporting multiple keys as ultimately the message validation fails because validators are invoked in sequence. 

Instead we should let app only add publicKeys which would be then appended to list of keys mapped against each topic so that a single validator function loops through all keys are validates.

chaitanyaprem: I see another issue of using signedTopicValidators when multiple communities are on same pubsubTopic. 
e.g; 
Status user has joined a community C1 which is on pubsubTopic P1 and hence the AddSignedTopicValidator method would have been called with PK of C1. 

If there is already a community C2 which is on P1 but this user is not part of, then all messages received for C2 will fail validation as key is not available with this user. Which means a lot of relay messages would fail and peers would be descored (leading to relay connectivity issues).

richard-ramos: Hmmm that's a very good point. It makes me think that the private key generation in status-go perhaps should be created deterministically based on the pubsub topic value, instead of being created randomly for each community. Otherwise we won't be able to safely have multiple communities with DoS protection running on the same shard.

chaitanyaprem: Yeah, but then anyone can push messages to that pubsubTopic which defeats the purpose of signing messages and validating them at pubsub layer. 
Which means as of now probably the only way to share pubsubTopics is via RLN then?
cc @jm-clius

fryorcraken: > Otherwise we won't be able to safely have multiple communities with DoS protection running on the same shard.

The proposal is to use an application key. A key set in Status software at building time.
We know that such key could be extracted with reverse engineering.
But in general, opt-in message signing is a weak DOS protection mechanisms. RLN beying the stronger one.

There could be a flow where an operator set the key for a shard, then communicate this key (private and public) to the community owner that then broadcast it to community members. But again, in this case, supporting several keys at the asme time would not be that useful.

> Which means as of now probably the only way to share pubsubTopics is via RLN then?

When sharing a shard, RLN is indeed the preferred way but using an application key could be a fair start to get going.

richard-ramos:
Closing this PR due to the issues described in comments.

Issue #1021, opened 2024-02-01
feat: Report shard/pubsubTopic specific connection health/status


chaitanyaprem:
### Problem

As of now peer connection is being reported to status-go based on number of peers connected ir-respective of shards.
Once multiple shards are to be supported either for communities or common ones like for 1:1 chat, the health should be reported separately per shard. 

### Suggested solution
As per initial discussion, following approach is to be taken:
1. Report relay peer health per shard based from peerManager(as it already tracks relay connections)
2. Report Filter and lightpush health from within status-go as peers are managed by it. This shall later be moved into SDK layer or within go-waku itself.

TBD: Analyse the approach currently followed where on every peer connection or disconnected status is reported.

### Alternatives considered
None

### Additional context
Ref issue https://github.com/status-im/status-go/issues/4628

### Acceptance criteria
Unit tests to verify health reporting

PR #1027, opened 2024-02-06
feat: topic health reporting

chaitanyaprem:
# Description
Implements https://github.com/waku-org/go-waku/issues/1021

# Changes

- [x] Store gossipsub scores in WakuPeerStore
- [x] Use PeerScores and DMin to determine topic-health
- [x] Report topic health to channel
- [x] Unit tests

# Tests

### DIFF
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -37,7 +37,7 @@ const (
 func (t TopicHealth) String() string {
 	switch t {
 	case UnHealthy:
-		return "Healthy"
+		return "UnHealthy"
 	case MinimallyHealthy:
 		return "MinimallyHealthy"
 	case SufficientlyHealthy:
--- a/waku/v2/peermanager/peer_manager.go
+++ b/waku/v2/peermanager/peer_manager.go
@@ -34,6 +34,11 @@ const (
 	SufficientlyHealthy = 2
 )
 
+type TopicHealthStatus struct {
+	topic  string
+	health TopicHealth
+}
+
 // NodeTopicDetails stores pubSubTopic related data like topicHandle for the node.
 type NodeTopicDetails struct {
 	topic        *pubsub.Topic
@@ -63,6 +68,7 @@ type PeerManager struct {
 	subRelayTopics         map[string]*NodeTopicDetails
 	discoveryService       *discv5.DiscoveryV5
 	wakuprotoToENRFieldMap map[protocol.ID]WakuProtoInfo
+	TopicHealthNotifCh     chan<- TopicHealthStatus
 }
 
 // PeerSelection provides various options based on which Peer is selected from a list of peers.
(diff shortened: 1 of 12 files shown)


waku/v2/peermanager/peer_manager.go
```diff
@@ -87,6 +102,55 @@ func inAndOutRelayPeers(relayPeers int) (int, int) {
...
+			pThreshold, err := pm.host.Peerstore().(wps.WakuPeerstore).Score(p)
+			if err == nil && pThreshold < relay.PeerPublishThreshold {
+				pm.logger.Debug("peer score below publish threshold", logging.HostID("peer", p), zap.Float64("score", pThreshold))
+			} else {
+				//For now considering peer as healthy if we can't fetch score.
+				healthyPeerCount++
+			}
+			if err != nil {
+				pm.logger.Warn("failed to fetch peer score ", zap.Error(err), logging.HostID("peer", p))
+			}
```
richard-ramos, line 120: I find this somewhat confusing, what do you think if we change it like this
```suggestion
			if err == nil {
			    if pThreshold < relay.PeerPublishThreshold {
				pm.logger.Debug("peer score below publish threshold", logging.HostID("peer", p), zap.Float64("score", pThreshold))
			    } else {
				    //For now considering peer as healthy if we can't fetch score.
				    healthyPeerCount++
			    }
		       } else {
				pm.logger.Warn("failed to fetch peer score ", zap.Error(err), logging.HostID("peer", p))
			}
```

waku/v2/peermanager/peer_manager.go
```diff
@@ -87,6 +102,55 @@ func inAndOutRelayPeers(relayPeers int) (int, int) {
...
+		}
+	}
+	//Update topic's health
+	oldHealth := topic.healthStatus
+	if healthyPeerCount < 1 { //Ideally this check should be done with minPeersForRelay, but leaving it as is for now.
+		topic.healthStatus = UnHealthy
+	} else if healthyPeerCount < 4 {
```
richard-ramos, line 127: Should we replace the 4 for
```suggestion
	} else if healthyPeerCount < waku_proto.GossipSubDMin {
```
?

waku/v2/peermanager/peer_manager.go
```diff
@@ -87,6 +102,55 @@ func inAndOutRelayPeers(relayPeers int) (int, int) {
...
+		topic.healthStatus = MinimallyHealthy
+	} else {
+		topic.healthStatus = SufficientlyHealthy
+	}
+
+	if oldHealth != topic.healthStatus {
+		//Check old health, and if there is a change notify of the same.
+		pm.logger.Debug("topic health has changed", zap.String("pubsubtopic", topic.topic.String()), zap.Int("health", int(topic.healthStatus)))
```
richard-ramos, line 135: I think we could make this more readable by adding this function:
```go
func (t TopicHealth) String() string {
    switch e {
    case UnHealthy:
        return "Healthy"
    case MinimallyHealthy:
        return "MinimallyHealthy"
    case SufficientlyHealthy:
        return "SufficientlyHealthy"
    }
}
```
and replacing `zap.Int` for `zap.Stringer`

richard-ramos (approved):

waku/v2/peermanager/peer_manager.go
chaitanyaprem: Looks like i missed handling a case while merging conditions in order to optimize. 
Let me clean it up. 

My idea was to consider a peer as healthy even if we can't fetch its score and increase healthyPeerCount.

PR #1035, opened 2024-02-17
chore: peermanager tests coverage improvement

romanzac:
### Description
Batch of tests for Waku Peermanager to improve coverage

### Changes
Tests available to run:
cd go-waku/waku/v2/peermanager
go test


waku/v2/peermanager/topic_event_handler_test.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	// peerEvt to find: relay.PEER_JOINED, relay.PEER_LEFT
+	peerEvt := relay.EvtPeerTopic{
+		PubsubTopic: pubSubTopic,
+		PeerID:      hosts[1].ID(),
+		State:       relay.PEER_JOINED,
+	}
+
+	emitter, err := eventBus.Emitter(new(relay.EvtPeerTopic))
```
chaitanyaprem, line 198: We don't need to emit these events as a peer joins a topic, the event get emitted from gossipsub layer

waku/v2/peermanager/topic_event_handler_test.go
```diff
@@ -0,0 +1,297 @@
...
+		log.Info("New topic subscribed", zap.String("topic", peerTopic.topic.String()))
+	}
+	pm.checkAndUpdateTopicHealth(peerTopic)
+	require.Equal(t, TopicHealth(UnHealthy), peerTopic.healthStatus)
+	time.Sleep(2 * time.Second)
+
+	// Process second to fourth PEER_JOINED event
+	for i := 2; i < 4; i++ {
+
+		peerEvt2 := relay.EvtPeerTopic{
```
chaitanyaprem, line 247: Similar comment to above, no need to emit events rather just let peers subscribe to a topic.

waku/v2/peermanager/topic_event_handler_test.go
```diff
@@ -0,0 +1,297 @@
...
+		require.NoError(t, err)
+	}
+
+	// Create peer manager instance with the first hosts
+	pm, eventBus := makePeerManagerWithEventBus(t, relays[0], &hosts[0])
+	pm.ctx = ctx
+	pm.RegisterWakuProtocol(relay.WakuRelayID_v200, relay.WakuRelayENRField)
+
+	// Connect host[0] with all other hosts to reach 4 connections
+	for i := 1; i < 5; i++ {
```
chaitanyaprem, line 168: Rather than adding peers this way, you can use PeerManager's AddPeer method which takes care of a bunch of these.

waku/v2/peermanager/topic_event_handler_test.go
romanzac: I've simplified event handling test to focus on adding peers to topic and connectedness only. Getting to health  would require full stack and this could be tested better at interop level.

waku/v2/peermanager/topic_event_handler_test.go
romanzac: I've simplified event handling test to focus on adding peers to topic and connectedness only. Getting to health  would require full stack and this could be tested better at interop level.

waku/v2/peermanager/topic_event_handler_test.go
romanzac: I've simplified event handling test to focus on adding peers to topic and connectedness only. Getting to health  would require full stack and this could be tested better at interop level.

waku/v2/peermanager/connection_gater_test.go
```diff
@@ -0,0 +1,86 @@
...
+	require.NoError(t, err)
+
+	peerA := h1.ID()
+
+	remoteAddr1 := ma.StringCast("/ip4/1.2.3.4/tcp/1234")
+
+	connGater := NewConnectionGater(2, log)
+
+	// Test peer blocking
+	allow := connGater.InterceptPeerDial(peerA)
```
chaitanyaprem, line 49: Did not understand the need for this, as the function InterceptPeerDial doesn't do anything but just returns.

Infact many of these intercept* functions don't have any implementation other than a blind return, so no point in writing tests for them

waku/v2/peermanager/connection_gater_test.go
```diff
@@ -0,0 +1,86 @@
...
+	require.True(t, allow)
+
+	// Test connection was secured and upgraded
+	allow = connGater.InterceptSecured(network.DirInbound, peerA, &mockConnMultiaddrs{local: nil, remote: nil})
```
chaitanyaprem, line 53: Same comment as above.

waku/v2/peermanager/topic_event_handler_test.go
```diff
@@ -0,0 +1,213 @@
...
+	ctxEventLoop := context.Background()
+	go pm.peerEventLoop(ctxEventLoop)
+
+	// Prepare emitter
+	emitter, err := eventBus.Emitter(new(relay.EvtPeerTopic))
+	require.NoError(t, err)
+
+	// Send PEER_JOINED events for hosts 2-5
+	for i := 1; i < 5; i++ {
+		err = emitTopicEvent(pubSubTopic, hosts[i].ID(), emitter, relay.PEER_JOINED)
```
chaitanyaprem, line 190: As i had mentioned before, there is no need to emit these events.
Just subscribe all the 5 hosts to same pubsubTopic and these events would automatically get generated.

waku/v2/peermanager/topic_event_handler_test.go
```diff
@@ -0,0 +1,213 @@
...
+	require.Equal(t, 4, len(pm.host.Peerstore().(*wps.WakuPeerstoreImpl).PeersByPubSubTopic(pubSubTopic)))
+
+	// Check all hosts have been connected
+	for _, peer := range pm.host.Peerstore().(*wps.WakuPeerstoreImpl).PeersByPubSubTopic(pubSubTopic) {
+		require.Equal(t, network.Connected, pm.host.Network().Connectedness(peer))
+	}
+
+	// Send PEER_LEFT events for hosts 2-5
+	for i := 1; i < 5; i++ {
+		err = emitTopicEvent(pubSubTopic, hosts[i].ID(), emitter, relay.PEER_LEFT)
```
chaitanyaprem, line 205: Same comment as above, just unsubscribe each node from the pubsubTopic and leave event would be generated automatically.

waku/v2/peermanager/service_slot_test.go
```diff
@@ -22,12 +22,31 @@ func TestServiceSlot(t *testing.T) {
...
+	peerID3 := peer.ID("peerId3")
+
+	//
+	slots.getPeers(protocol).add(peerID2)
+	slots.getPeers(protocol).add(peerID3)
+	//
+
+	fetchedPeers, err = slots.getPeers(protocol).getRandom(2)
+	require.NoError(t, err)
+	require.Equal(t, 2, len(maps.Keys(fetchedPeers)))
```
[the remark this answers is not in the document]
romanzac: Check added at https://github.com/waku-org/go-waku/pull/1035/commits/dfdde56fd61209c567132fa067088db373bd2193

waku/v2/peermanager/connection_gater_test.go
romanzac: These functions are tested for consistency. It is not possible to cover all functionality at once. When I will return back either because of a problem, or because I want to add a test, I will be able to quickly see where are untested spots. More spots -> more complicated to asses what to do about them. I may not remember connection gater had so simple implementation.

waku/v2/peermanager/topic_event_handler_test.go
romanzac: Would be interesting to know why it doesn't work like you have described.

romanzac: My approach is to get as minimal moving parts as possible to get my unit moving in expected direction. I need either 100% clarity for unit test on required pieces or rather move test to interop level, if there is something unexplaineable.

waku/v2/protocol/filter/filter_push_test.go
```diff
@@ -16,7 +16,7 @@ func (s *FilterTestSuite) TestValidPayloadsASCII() {
 	s.subDetails = s.subscribe(s.testTopic, s.testContentTopic, s.fullNodeHost.ID())
 
 	// Prepare data
-	messages := prepareData(100, false, false, true, tests.GenerateRandomASCIIString)
+	messages := prepareData(50, false, false, true, tests.GenerateRandomASCIIString)
```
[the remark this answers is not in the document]
chaitanyaprem: > I've reverted related commit. I'll create dedicated PR later. It would need to be merged before this PR. Not sure why Github runners become slower. [d6fb227](https://github.com/waku-org/go-waku/commit/d6fb227372a299eeec1c27655b10581fbdf53934)

Ah, ok...in that case it is fine.

waku/v2/peermanager/topic_event_handler_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	for i := 1; i < 5; i++ {
+		// Subscribe to Pubsub topic on first host only
+		_, err := relays[i].Subscribe(ctx, protocol.NewContentFilter(pubSubTopic))
+		require.NoError(t, err)
+	}
+
+	// Start event loop to listen to events
+	ctxEventLoop := context.Background()
+	go pm.peerEventLoop(ctxEventLoop)
```
chaitanyaprem, line 184: You should start eventloop before you subscribe, otherwise events might get missed.

waku/v2/peermanager/topic_event_handler_test.go
```diff
@@ -0,0 +1,218 @@
...
+
+	//// Send PEER_LEFT events for hosts 2-5
+	//for i := 1; i < 5; i++ {
+	//	err = emitTopicEvent(pubSubTopic, hosts[i].ID(), emitter, relay.PEER_LEFT)
+	//	require.NoError(t, err)
+	//	time.Sleep(100 * time.Millisecond)
+	//}
+	//
+	//// Check all hosts have left the topic
+	//require.Equal(t, 0, len(pm.host.Peerstore().(*wps.WakuPeerstoreImpl).PeersByPubSubTopic(pubSubTopic)))
```
chaitanyaprem, line 216: This can also be uncommented once you call unsubscribe

waku/v2/peermanager/topic_event_handler_test.go
chaitanyaprem: That could probably be happening because the eventLoop was started after subscribe.

romanzac: I've moved eventLoop() goroutine start ahead of subscription. https://github.com/waku-org/go-waku/pull/1035/commits/0372719e4713f5969ded5060c4974acb7460be97

Result is the same. I am not getting PEER_JOINED, relay.PEER_LEFT. There is only one place in code where those get emited: (w *WakuRelay) topicEventPoll(topic string, handler *pubsub.TopicEventHandler) which doesn't use (pm *PeerManager) handlerPeerTopicEvent(peerEvt relay.EvtPeerTopic). It looks we expect something to happen at pubsub (libp2p) level and then react on our (peermanager) side to it. I was not able to find what is that specifically. Your thoughts ?

waku/v2/peermanager/topic_event_handler_test.go
romanzac: Let's see later.

waku/v2/peermanager/topic_event_handler_test.go
romanzac: Commented above.

waku/v2/peermanager/topic_event_handler_test.go
chaitanyaprem: Interesting. 
One thing i had noticed is that relay[0] was not subscribed to pubsubTopic in the code. 
But even after that events are not getting generated. Not sure what could be causing the issue.

romanzac: How should we proceed with this PR ? I have a milestone to finish this week. I really like to crack mysteries, right now though I suggest:
1) Open new issue to look at the libp2p/pm event transitions later  
2) I revert test code back to test pm's side of the event loop only

What do you think ?

chaitanyaprem: OK, we can take that approach.

chaitanyaprem (commented):
CI issues need to be fixed.

romanzac:
> bug: metadata might not always be available #1055

I know. Right now we are blocked on https://github.com/waku-org/go-waku/issues/1055 I cannot make test pass without this issue to resolve. It is not big change I assume.

romanzac:
Flaky service slot tests fixed at https://github.com/waku-org/go-waku/pull/1058

PR #1037, opened 2024-02-19
chore: drop legacy filter support

richard-ramos:
# Description
Removes support for legacy filter.
cc: @jm-clius 

# Changes


jm-clius:
Great! So presumably this is safe to remove from nwaku too? cc @Ivansete-status

chaitanyaprem (approved):
LGTM

Do rerun CI as test seems to have failed.

PR #1041, opened 2024-02-21
refactor: use ttl of 5m for subscriptions

richard-ramos:
Filter subscribers need to ping at least once in the last 4:59m to keep their subscription alive (like in nwaku)


chaitanyaprem (approved):
LGTM

PR #1043, opened 2024-02-26
feat(peer-exchange): rate limit

### LINKED ISSUE #1042: feat: rate limit for PeerExchange Service
### Problem
As status app in relay mode is going to run peerExchange as a service, rate limiting needs to be introduced so that a user doesn't get DOSed.
This can be similar to what has been done for lightpush.

richard-ramos:
# Description
Adds rate limit option to peer exchange

## Issue

closes #1042


chaitanyaprem:
Created relevant issue  https://github.com/status-im/status-go/issues/4857 in status-go for tracking.

PR #1051, opened 2024-03-13
chore: discv5 tests coverage improvement

romanzac:
### Description
Batch of tests for Waku Discv5 to improve coverage

### Changes
Tests available to run:
cd go-waku/waku/v2/discv5
go test


waku/v2/discv5/discover_test.go
```diff
@@ -15,6 +20,51 @@ import (
 	"github.com/waku-org/go-waku/waku/v2/utils"
 )
 
+func discoverFilterOnDemand(iterator enode.Iterator, maxCount int) ([]service.PeerData, error) {
```
chaitanyaprem, line 23: This should already be covered in peer manager tests as https://github.com/waku-org/go-waku/blob/826c7fb924e1f070e414773d9bee76998ad479aa/waku/v2/peermanager/peer_discovery.go#L52-L103 is already doing same functionality. 

Is it not being show in coverage report?

waku/v2/discv5/discover_test.go
```diff
@@ -102,3 +152,276 @@ func TestDiscV5(t *testing.T) {
...
+	iterator3, err := d3.PeerIterator(FilterCapabilities(filterBitfield))
+	require.NoError(t, err)
+	require.NotNil(t, iterator3)
+
+	time.Sleep(2 * time.Second)
+
+	// Check node were discovered by automatic discovery
+	require.True(t, peerconn3.HasPeer(host1.ID()) && peerconn3.HasPeer(host2.ID()))
+
+	peers, err := discoverFilterOnDemand(iterator3, 1)
```
chaitanyaprem, line 224: I am wondering if we use existing `filterPredicate` option to filter discoveredPeers and let discovery run rather than writing similar method?

waku/v2/discv5/discover_test.go
```diff
@@ -102,3 +152,276 @@ func TestDiscV5(t *testing.T) {
...
+
+	// Host1 has store support while host2 hasn't
+	require.Equal(t, host1.ID().String(), peers[0].AddrInfo.ID.String())
+
+	d3.Stop()
+	peerconn3.Clear()
+
+}
+
+func TestDiscV5WithShardFilter(t *testing.T) {
```
chaitanyaprem, line 236: Shard filtering is alredy done in defaultPredicate. If we just write a test to use it, then we should be good.
https://github.com/waku-org/go-waku/blob/826c7fb924e1f070e414773d9bee76998ad479aa/waku/v2/discv5/discover.go#L385
Unless i am missing something else, this test is only checking discv5 filtering with shards right.

waku/v2/discv5/discover_test.go
```diff
@@ -102,3 +152,276 @@ func TestDiscV5(t *testing.T) {
...
+	// Store values into metric client struct
+	err := counter.Write(failures)
+	require.NoError(t, err)
+
+	// Check the count is in
+	require.Equal(t, 2, int(failures.GetCounter().GetValue()))
+
+}
+
+func TestDiscV5WithCustomPredicate(t *testing.T) {
```
chaitanyaprem, line 360: Nice test!

waku/v2/discv5/discover_test.go
romanzac: Yes, the main reason I come up with test cases for FilterShard() and FilterCapabilities() was they didn't show up as covered. Another reason why to test code which is already used somewhere else is the absence of any assertions. Unit tests serve also as a guard for regressions or could detect race conditions after unit got affected by changes somewhere else. I agree best value would be to come up with completely unique tests. This will eventually come later once I learn and seal the main flow.

waku/v2/discv5/discover_test.go
romanzac: It would probably work. FilterPredicate, DefaultPredicate were covered. My curiosity was: could I actually stack discovery and filter methods together ? First use "normal" discovery and check, second step use onDemand to filter out and check again. Now I understand filter just goes through already discovered nodes. Right ? Test code runs very fast, so these two step should yield same result with same nodes. If we add another node and let discovery to run longer, we may end up with the second check having different result.  I had brief discussion with @richard-ramos about the iterator being never ending. Perhaps it is on purpose, because we filter from dynamic pool of nodes.

waku/v2/discv5/discover_test.go
chaitanyaprem: Odd that they are not showing as covered, maybe the coverage report is not aggregating coverage from other tests is it? I thought i fixed this at some point.
Yeah, i just don;t want soo much code to be written to simulate things as it would become more maintenance prone as we need to change something especially if it is getting covered by some other test in another package.

waku/v2/discv5/discover_test.go
chaitanyaprem: > Now I understand filter just goes through already discovered nodes. Right ?

Not already discovered nodes. Filter is run on the nodes as they are being discovered.

chaitanyaprem: > Perhaps it is on purpose, because we filter from dynamic pool of nodes.

Yes, as nodes are discovered iterator is run on them..and discovery doesn't stop as long as discv5 is running.

waku/v2/discv5/discover_test.go
chaitanyaprem: Again coming back to my point that we don't need to simulate some code if similar code already exists as another unit. Otherwise test code will become more than actual code and it would be very difficult to maintain.

waku/v2/discv5/discover_test.go
romanzac: > > Perhaps it is on purpose, because we filter from dynamic pool of nodes.
> 
> Yes, as nodes are discovered iterator is run on them..and discovery doesn't stop as long as discv5 is running.

The behavior I saw is round robin across the same nodes which were already discovered. It was like discovery and iterator don't know about each other. For example 3 hosts and count param 6 in discoverFilterOnDemand, output will be  host1,host2,host3,host1,host2,host3. When count param is 2, no problem host1,host2 will get filtered. If n < count, caller needs to know hosts could potentially repeat and dedup on result is needed.

waku/v2/discv5/discover_test.go
romanzac: I have checked again:
1) Commented out my tests TestDiscV5WithShardFilter and TestDiscV5WithCapabilitiesFilter 
2) Run coverage report at /go-waku dir with "go test -coverprofile=coverage.out ./..."
3) FilterShard and FilterCapabilities haven't show up as covered

Looked up with GPT, tests from other packages could cover funcs in "local" package if:
1) There is explicit import and use of "remote" package
2) "go test -coverprofile=coverage.out ./..."   is used

discoverOnDemand() in peermanager package is not public, so it cannot be explicitly imported. And it is a normal code without any assertions. Even I believe is works well and no issues about it, I still need some code to build my test case to cover  FilterShard and FilterCapabilities. If we agree we would like to have better coverage which is traceable. 

I got your point, it really is more maintenance work. Coverage tests create sort of a cage limiting freedom of devs to be that fast. I briefly discussed with @fbarbu15 wether QA team could maintain tests broken with new changes and he agreed on it. I personally like to learn along the changes you devs introduce, so no problem on my side either. We would just need to hand over PRs from dev to QA and back. Please reach to you leaders if this model is we want. We should ideally get both quality and still deliver changes at decent speed?

chaitanyaprem: Ah..Thanks for finding it out.

Ok, in that case then it is probably ok to proceed with this.

waku/v2/discv5/discover_test.go
chaitanyaprem: That could be happening because the nodes are just sent regularly in discv5. 
IIRC, the discv5 doesn't keep a track of which nodes have been shared with which peer and hence it may end up sending an already shared node which is why you might be seeing this repeated behaviour of same nodes being reported agian and agian.

This is why the user of discv5 needs to store the nodes and ignore if any node is already reported. In our case that is done by using peerStore by the peerManager.

chaitanyaprem (approved):
LGTM

PR #1052, opened 2024-03-13
fix(metadata): only verify peer metadata on relay

richard-ramos:
# Description
As described here https://github.com/waku-org/nwaku/issues/2491 metadata protocol disconnects light clients.
With this change, instead of doing the metadata verification for all peers, it is done only for relay peers.


chaitanyaprem (commented):
Changes look good, but as per original issue the clusterID check has to be done for all peer connections ir-respective of relay or not. 
Could not find that being checked. Or i maybe missing something.

chaitanyaprem (approved):
LGTM.
If comment is addressed, then can be merged.

richard-ramos:
Good point. Let me add that back quickly

PR #1060, opened 2024-03-19
chore: sharding tests update

romanzac:
### Description
Batch of tests for Waku sharding functionality to extend existing tests 

### Changes
Tests available to run:
cd go-waku/waku/v2/node
go test


waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +335,94 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+func TestStaticShardingMultipleTopics(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	// Node1 with Relay
+	hostAddr1, err := net.ResolveTCPAddr("tcp", "0.0.0.0:0")
+	require.NoError(t, err)
+	wakuNode1, err := New(
+		WithHostAddress(hostAddr1),
+		WithWakuRelay(),
```
chaitanyaprem, line 348: If we are writing static sharding tests, we should specify clusterID when we create the Waku node.

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +335,94 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+	)
+	require.NoError(t, err)
+	err = wakuNode1.Start(ctx)
+	require.NoError(t, err)
+	defer wakuNode1.Stop()
+
+	pubSubTopic1 := protocol.NewStaticShardingPubsubTopic(uint16(0), uint16(0))
```
chaitanyaprem, line 355: ClusterID 0 cannot be used for static sharding. 0 would be considered as non-sharded default pubsubTopic.

Refer rfc https://rfc.vac.dev/spec/52/ for more details on shard allocation and usage.

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +335,94 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+		defer wg.Done()
+		env, ok := <-subs1[0].Ch
+		require.True(t, ok, "no message retrieved")
+		require.Equal(t, msg.Timestamp, env.Message().Timestamp)
+	}()
+
+	wg.Wait()
+
+	// Send another message to non-subscribed pubsub topic, but subscribed content topic
+	msg2 := tests.CreateWakuMessage(contentTopic1, utils.GetUnixEpoch(), "test message 2")
```
chaitanyaprem, line 407: This test looks like testing Waku relay with static sharding and it looks like only 1 node is used for the same. 
Not sure what additional code coverage are we achieving with this.

romanzac: I am learning and creating rather than trying to improve coverage in this PR. I saw some tests for sharding already and would not like to do the same things. I've marked the PR ready to review because I would need your help. We would eventually move out some of the tests once we vet them. Please note I mostly use public API. Even, if it was not indented this way,  it can still yield insights like https://github.com/waku-org/go-waku/issues/1070

waku/v2/node/wakunode2_test.go
romanzac: Are all 0-15 cluster index IDs reserved -> cannot be used with static sharding ?

waku/v2/node/wakunode2_test.go
romanzac: I see, so the clusterID could actually live, be useful, without sharded topic ?

waku/v2/node/wakunode2_test.go
chaitanyaprem: Reserved means can be used with static-sharding. It is just that someone else should not use these clusters other than Status/Waku.
But 0 is not supposed to be used for static-sharding as it is marked as global.

waku/v2/node/wakunode2_test.go
chaitanyaprem: If it is still in progress, then you can leave the PR in draft and assign me for review and ping me so that i can take a look.

waku/v2/node/wakunode2_test.go
chaitanyaprem: Did not get what you mean

romanzac: I curious about value of setting clusterID (node.WithClusterID ) without having intent to use any shard.

chaitanyaprem: But you are using a shard when you are subscribing the pubsubTopic that is part of the clusterid. 
the pubsubTopic format is `/waku/2/rs/<cluster-ID>/<shard>` which means a shard will be included.

romanzac: I see, it is used to create metadata when node instance is created: 

metadata := metadata.NewWakuMetadata(w.opts.clusterID, w.localNode, w.log)
w.metadata = metadata

Metadata won't even start without clusterID != 0:
```
// Start inits the metadata protocol
func (wakuM *WakuMetadata) Start(ctx context.Context) error {
	if wakuM.clusterID == 0 {
		wakuM.log.Warn("no clusterID is specified. Protocol will not be initialized")
		return nil
	}
	.
	.
	.
}
```
I've added WithClusterID and it has no effect on my test. Next interesting thing: node is a member of only one cluster from metadata perspective. Correct ? https://github.com/waku-org/go-waku/pull/1060/commits/4266513a252b3b30480bf2ccad6a85e57d7ee7f0

chaitanyaprem: `node is a member of only one cluster from metadata perspective. Correct ?`

Yes, a node can be part of a single cluster. And without clusterID metadata protocol won't be initialized.

As i mentioned above, since these are tests for sharding and as Waku supports 2 types static and auto sharding, we should try to use those formats.

romanzac:
Issue https://github.com/waku-org/go-waku/issues/1077 related to this PR

tests/utils.go
```diff
@@ -385,3 +388,36 @@ func GenerateRandomSQLInsert(maxLength int) (string, error) {
...
+}
+
+func WaitForTimeout(t *testing.T, ctx context.Context, wg *sync.WaitGroup, ch chan *protocol.Envelope) {
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		select {
+		case _, ok := <-ch:
+			require.False(t, ok, "should not retrieve message")
+		case <-time.After(1 * time.Second):
```
chaitanyaprem, line 415: minor suggest, maybe take timeout as an arg to the method so that it can be used in other tests as well.

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +321,222 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+		WithHostAddress(hostAddr1),
+		WithWakuRelay(),
+		WithClusterID(uint16(20)),
+	)
+	require.NoError(t, err)
+	err = wakuNode1.Start(ctx)
+	require.NoError(t, err)
+	defer wakuNode1.Stop()
+
+	pubSubTopic1 := protocol.NewStaticShardingPubsubTopic(uint16(21), uint16(0))
```
chaitanyaprem, line 342: Shouldn't 20 be passed as clusterID as node1 is being initialised with clusterID as 20?

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +321,222 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+	pubSubTopic1Str := pubSubTopic1.String()
+	contentTopic1 := "/test/2/my-app/sharded"
+
+	pubSubTopic2 := protocol.NewStaticShardingPubsubTopic(uint16(21), uint16(10))
```
chaitanyaprem, line 346: Same as above

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +321,222 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+		require.True(t, ok, "no message retrieved")
+		require.Equal(t, msg.Timestamp, env.Message().Timestamp)
+	}()
+
+	wg.Wait()
+
+	// Send another message to non-subscribed pubsub topic, but subscribed content topic
+	msg2 := tests.CreateWakuMessage(contentTopic1, utils.GetUnixEpoch(), "test message 2")
+
+	_, err = r.Publish(ctx, msg2, relay.WithPubSubTopic("/waku/2/rs/0/321"))
```
chaitanyaprem, line 398: Using clusterID 0 here? 
Better to use a different pubsunTopic in cluster 20 itself which is not subscribed.

chaitanyaprem (commented):
Did partial review, but looks like comments need to be addressed

tests/utils.go
romanzac: Thanks for suggestion. Added timeout param to both WaitFor funcs. https://github.com/waku-org/go-waku/pull/1060/commits/99e3f334cd8356d343d3f3f7ac750ca46508de05

waku/v2/node/wakunode2_test.go
romanzac: It is on purpose to check if sharded cluster ID could be different than native clusterID - message still delivered. I've added variables to display intent. Not sure if "native" and sharded are the good labels. https://github.com/waku-org/go-waku/pull/1060/commits/99e3f334cd8356d343d3f3f7ac750ca46508de05

waku/v2/node/wakunode2_test.go
romanzac: Good point, I've forgot 0 is already subscribed. Fixed at https://github.com/waku-org/go-waku/pull/1060/commits/99e3f334cd8356d343d3f3f7ac750ca46508de05

waku/v2/node/wakunode2_test.go
chaitanyaprem: Hmm, that is because clusterID is just a user abstraction for underlying pubsubTopic. 
When we do subscribe for relay, we don't check if the node is initialized with that clusterID.
Not sure if we will add such a check in go-waku, but it would be nice to not allow this.

cc @richard-ramos, something to be considered for nwaku bindings.

romanzac: I wonder what is the story behind the lonely clusterID ?  Topic is the central concept we use to shard the entire network isn't it ? Why do we need on the side clusterID which is just one of many possible topics ?

romanzac:
I've reverted "TestStaticShardingLimits" to its desired form and marked it as flaky. This way we can unblock this PR and start a collection of flaky tests. Flaky test could take a time to resolve, sometimes it could lead to a design change. There is still some value to keep them. What are your thoughts ? @chaitanyaprem @richard-ramos 
I suggest skip mechanism for those tests would be:
```
	if os.Getenv("RUN_FLAKY_TESTS") != "true" {

		log.Info("Skipping", zap.String("test", t.Name()),
			zap.String("reason", "RUN_FLAKY_TESTS environment variable is not set to true"))
		t.SkipNow()
	}
```

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +323,218 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+	require.NoError(t, err)
+
+	time.Sleep(100 * time.Millisecond)
+
+	// No message could be retrieved
+	tests.WaitForTimeout(t, ctx, 1*time.Second, &wg, subs1[0].Ch)
+
+}
+
+func TestStaticShardingLimits(t *testing.T) {
```
chaitanyaprem, line 424: Looks like a nice test, but maybe we may want to add more verification by sending messages in a loop randomly on a topic and waiting for them to be received to ensure all subscriptions are valid.

waku/v2/protocol/filter/filter_test.go
```diff
@@ -543,3 +543,67 @@ func (s *FilterTestSuite) BeforeTest(suiteName, testName string) {
...
+	s.ctx = ctx
+	s.ctxCancel = cancel
+
+	// Gen pubsub topic "/waku/2/rs/1024/0"
+	s.testTopic = protocol.NewStaticShardingPubsubTopic(uint16(1024), uint16(100)).String()
+
+	// Pubsub topics for neg. test cases
+	testTopics := []string{
+		"/waku/2/rs/100/1024",
+		"/waku/2/rs/1024/101",
```
chaitanyaprem, line 562: Again, as i had mentioned above...whatever we want can be achieved by using shards from same cluster. 

Let us not use shards from different cluster.

waku/v2/protocol/lightpush/waku_lightpush_test.go
```diff
@@ -332,3 +314,64 @@ func TestWakuLightPushCornerCases(t *testing.T) {
...
+
+	// Prepare pubsub topics for static sharding
+	pubSubTopic := protocol.NewStaticShardingPubsubTopic(uint16(25), uint16(0)).String()
+	testContentTopic := "/test/10/my-lp-app/proto"
+
+	node1, sub1, host1 := makeWakuRelay(t, pubSubTopic)
+	defer node1.Stop()
+	defer sub1.Unsubscribe()
+
+	node2, sub2, host2 := makeWakuRelay(t, pubSubTopic)
```
chaitanyaprem, line 330: Node2 doesn't need to be a relay node if it wants to use lightpush for publishing. 
It can just be a simple lightpush node and use node1 to publish.

waku/v2/protocol/store/waku_store_protocol_test.go
```diff
@@ -425,3 +425,86 @@ func TestWakuStoreStart(t *testing.T) {
...
+
+	host2.Peerstore().AddAddr(host1.ID(), tests.GetHostAddress(host1), peerstore.PermanentAddrTTL)
+	err = host2.Peerstore().AddProtocols(host1.ID(), StoreID_v20beta4)
+	require.NoError(t, err)
+
+	s2 := NewWakuStore(MemoryDB(t), nil, timesource.NewDefaultClock(), prometheus.DefaultRegisterer, utils.Logger())
+	s2.SetHost(host2)
+
+	// Subscribe to DefaultWakuTopic at store2 + host2
+	sub1 := relay.NewSubscription(protocol.NewContentFilter(relay.DefaultWakuTopic))
```
chaitanyaprem, line 473: Did not understand why this is required? DefaultpubsubTopic is not related to static sharding.

chaitanyaprem (changes requested):
Gave some comments.

Also CI is failing which needs to be fixed.

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +323,218 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
 	require.Len(t, result.Messages, 1)
 	require.Equal(t, msg.Timestamp, result.Messages[0].Timestamp)
 }
+
+func TestStaticShardingMultipleTopics(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	nodeNativeClusterID := uint16(20)
```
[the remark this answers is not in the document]
romanzac: I picked 2 as a minimum for the assumed use case: 
When an operator starts node, they have a control about cluster ID. They don't have any control what pubsub topic with other cluster IDs their users might subscribe to. Act of subscription is anonymous and is not (spam) protected. Provided user follows the supported pubsub topic spec.

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +323,218 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+
+	pubSubTopic2 := protocol.NewStaticShardingPubsubTopic(shardingClusterID, uint16(10))
+	pubSubTopic2Str := pubSubTopic2.String()
+	contentTopic2 := "/test/3/my-app/sharded"
+
+	require.Equal(t, nodeNativeClusterID, wakuNode1.ClusterID())
+
+	r := wakuNode1.Relay()
+
+	subs1, err := r.Subscribe(ctx, protocol.NewContentFilter(pubSubTopic1Str, contentTopic1))
```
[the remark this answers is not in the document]
romanzac: Described above. If this usage is not intended by devs, it would be better to prevent it in the code. 

Is the cluster ID an administrative concept ? More like a sticky label of a node for human operator?  ENR is hard to remember, otherwise ENR+pubsub topic does the real work with message routing. Am I getting it?

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +323,218 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+
+	_, err = r.Publish(ctx, msg2, relay.WithPubSubTopic("/waku/2/rs/100/321"))
+	require.NoError(t, err)
+
+	time.Sleep(100 * time.Millisecond)
+
+	// No message could be retrieved
+	tests.WaitForTimeout(t, ctx, 1*time.Second, &wg, subs1[0].Ch)
+
+	// Send another message to subscribed pubsub topic, but not subscribed content topic - mix it up
```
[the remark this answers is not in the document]
romanzac: Here I am checking, if filtering based on content topic works within the context of static sharding.

waku/v2/protocol/filter/filter_test.go
romanzac: Cluster IDs unified at https://github.com/waku-org/go-waku/pull/1060/commits/7404bb416fe513d2dd68387657767adaa4f5b852

waku/v2/node/wakunode2_test.go
chaitanyaprem: That is to help with discovery of peers. If we want to scale beyond 1 pubsubTopic, we can do that by having multiple pubsubTopics. But, how do we discover peers in the network then?
We use discv5 for discovery which depends on ENR. ENR has a size restriction where-in we can't include a list of pubsubTopics in it. 

Hence to make it easier to encode this information in the ENR, the pubsubTopic has been formatted into clusterID and shards. 
The idea of clusterID is kind of similar to networkID in a blockchain , where-in a node is expected to be part of 1 cluster only.
You can refer to rfc https://github.com/waku-org/specs/blob/master/standards/core/relay-sharding.md for more details. 

The limit of 1024 shards within a cluster also has similar reasons behind it.  One of them being discv5 DHT not being able to scale beyond a particular number of nodes i.e finding nodes beyond a particular size is going to take a lot of time that it becomes almost useless. IIRC that node count is 10K or something like that. 


Hope this helps.

waku/v2/node/wakunode2_test.go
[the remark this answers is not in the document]
chaitanyaprem: Ideally, we don't want users doing that. i.e start node with clusterID x and subscribe to a topic that is belonging to clusterID y.  
Wondering if we should throw an error if someone tries to subscribe to a topic from different cluster.

waku/v2/node/wakunode2_test.go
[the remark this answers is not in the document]
chaitanyaprem: Gave a brief explanation https://github.com/waku-org/go-waku/pull/1060#discussion_r1587048424

waku/v2/node/wakunode2_test.go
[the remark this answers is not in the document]
chaitanyaprem: Ok, in that case we should subscribe to a specific contentTopic  ct1 within a pubsubTopic and then send messages to ct1 and some messages to ct2. 
Add a check that msgs sent to ct1 were received by intended node and ct2 messages were not received (i.e filtered out).

waku/v2/protocol/store/waku_store_protocol_test.go
romanzac: Is there another way how can I get empty subscription ? I need it for s2.Start(ctx, sub1), but I wish s2 doesn't know anything upfront about pubsub topics of s1.

chaitanyaprem: Not sure what you mean by empty sub, but you can always ise a different pubsubTopic in same cluster that is not being used by s1.

waku/v2/node/wakunode2_test.go
romanzac: Thanks for explanation.

waku/v2/node/wakunode2_test.go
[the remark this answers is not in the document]
romanzac: I would throw an error with a cool message and reject to subscribe. Misuse, abuse or any misunderstanding like mine would create future work, which scales too well. I think it is sensible to have a lean design at this stage of Waku's  adoption.

waku/v2/protocol/store/waku_store_protocol_test.go
romanzac: New topic added at https://github.com/waku-org/go-waku/pull/1060/commits/12a8923d6270d4e185a683607bb51fe1d3beb987

waku/v2/protocol/lightpush/waku_lightpush_test.go
romanzac: Node2 is a forwarder to node1, which is only relay.  I've tried to change node2 to be only lightpush, and received confusing error. https://github.com/waku-org/go-waku/pull/1060/commits/73a90e085c8b90418abab36e2abea17abee74d58

```
--- FAIL: TestWakuLightPushWithStaticSharding (2.47s)
    waku_lightpush_test.go:368:
        	Error Trace:	/Users/roman/sources/waku-org/go-waku/waku/v2/protocol/lightpush/waku_lightpush_test.go:368
        	Error:      	Received unexpected error:
        	            	failed to negotiate protocol: protocols not supported: [/vac/waku/lightpush/2.0.0-beta1]
        	Test:       	TestWakuLightPushWithStaticSharding
```

Have I missed something ?

chaitanyaprem: Ah..ok. 
If so, then node2 has to be support relay and lightpush as service node.

romanzac: Reorganized at https://github.com/waku-org/go-waku/pull/1060/commits/093ea363a6ac347c0c424360a34075bfed62a4a7

waku/v2/node/wakunode2_test.go
romanzac: Would not be sufficient if we randomly chose one pubsub topic at each test run ?

waku/v2/node/wakunode2_test.go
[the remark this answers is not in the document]
romanzac: I have three pubsub and two content topics in the test. ct1 and ct2 are subscribed together with their respective pt1 and pt2. Cases: 1. pt1+ct1 subscribed and receiving, 2. pt3(non-subscribed)+ct2 not receiving,  and 3. pt1+ct2(non-subscribed, but known to the node) not receiving. 

I guess first and third case matches with your proposal except I have a twist, the ct2 is not completely unknown to the node ?

waku/v2/node/wakunode2_test.go
romanzac: I've unified clusterID with pubsub topic at https://github.com/waku-org/go-waku/pull/1060/commits/dd960ef93dc1c72149e3561d326c77f4721b57b4

waku/v2/node/wakunode2_test.go
chaitanyaprem: It should be, but maybe just randomly choose some shards to ensure no weird bug in the code.

waku/v2/node/wakunode2_test.go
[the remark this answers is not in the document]
chaitanyaprem: Also, you can raise an issue so that we don't forget about this.

Although not sure if we want to implement in go-waku as it is being sunset.

waku/v2/node/wakunode2_test.go
```diff
@@ -318,3 +323,217 @@ func TestDecoupledStoreFromRelay(t *testing.T) {
...
+		require.True(t, ok, "no message retrieved")
+		require.Equal(t, msg.Timestamp, env.Message().Timestamp)
+	}()
+
+	wg.Wait()
+
+	// Send another message to non-subscribed pubsub topic, but subscribed content topic
+	msg2 := tests.CreateWakuMessage(contentTopic1, utils.GetUnixEpoch(), "test message 2")
+
+	_, err = r.Publish(ctx, msg2, relay.WithPubSubTopic("/waku/2/rs/100/321"))
```
chaitanyaprem, line 402: Better to use pubsubTopic from testCluster 20 here which is not subscribed to.

waku/v2/node/wakunode2_test.go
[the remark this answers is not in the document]
chaitanyaprem: Ok, that is a weird case but fine.

Ideally a contentTopic is tied to a pubsubTopic, not sure there would be an actual usecase where the same contentTopic is used in multiple pubsubTopics

waku/v2/node/wakunode2_test.go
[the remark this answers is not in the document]
romanzac: Let's keep it in this PR only. I don't want to generate additional work, if that is what sunsetting means.

chaitanyaprem:
@romanzac CI seems to be failing, we will have to fix that. 

One more comment and once both of these are fixed, we can merge .

waku/v2/node/wakunode2_test.go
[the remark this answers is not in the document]
chaitanyaprem: My bad, no issue with this comment. 

I was talking about https://github.com/waku-org/go-waku/pull/1060#discussion_r1587172611.

romanzac:
> @romanzac CI seems to be failing, we will have to fix that.
> 
> One more comment and once both of these are fixed, we can merge .

CI if failing at Go installation step. I saw https://github.com/waku-org/go-waku/pull/1091 upgrades Go version. Could you please check for me value of ${{ needs.env.outputs.go_version }} in repo settings or give me access so I can see the value ?

chaitanyaprem:
> > @romanzac CI seems to be failing, we will have to fix that.
> > One more comment and once both of these are fixed, we can merge .
> 
> CI if failing at Go installation step. I saw #1091 upgrades Go version. Could you please check for me value of ${{ needs.env.outputs.go_version }} in repo settings or give me access so I can see the value ?

Hmm, that is not yet merged. Also, wondering why it is failing for this PR only?

Also, i don't see any such setting in repo settings->environment. Is that the place to look?

chaitanyaprem (approved):
Can be merged once CI is fixed.

chaitanyaprem:
Maybe @richard-ramos know whats happening with CI.

romanzac:
> Can be merged once CI is fixed.

I've contacted Richard few hours ago. I'll try few things before he responds.

chaitanyaprem:
Why is this branch showing soo much diff with master? 
Let us not merge when diff is like this..it means the gut tree is probably messed up. 
Ideally, we should rebase from master, pass CI and then merge it. 

Hopefully nothing got messed up.

romanzac:
> Why is this branch showing soo much diff with master? Let us not merge when diff is like this..it means the git tree is probably messed up. Ideally, we should rebase from master, pass CI and then merge it.
> 
> Hopefully nothing got messed up.

This is what I have done. Rebased from master, passed CI and then merged. No worries.

chaitanyaprem:
> > Why is this branch showing soo much diff with master? Let us not merge when diff is like this..it means the git tree is probably messed up. Ideally, we should rebase from master, pass CI and then merge it.
> > 
> > Hopefully nothing got messed up.
> 
> This is what I have done. Rebased from master, passed CI and then merged. No worries. 

Oh, interesting...in that case the diff shouldn't be this big right

romanzac:
> > > Why is this branch showing soo much diff with master? Let us not merge when diff is like this..it means the git tree is probably messed up. Ideally, we should rebase from master, pass CI and then merge it.
> > > Hopefully nothing got messed up.
> > 
> > 
> > This is what I have done. Rebased from master, passed CI and then merged. No worries.
> 
> Oh, interesting...in that case the diff shouldn't be this big right

I remember I've done merge at some point too.

PR #1063, opened 2024-03-20
feat: draft changes enabling auto sharding in all clusters

chaitanyaprem:
# Description

# Changes

- [ ] Enable autosharding for all clusters
- [ ] Don't hardcode shard count for autosharding to 8

# Tests

### DIFF
--- a/cmd/waku/node.go
+++ b/cmd/waku/node.go
@@ -455,7 +455,7 @@ func processTopics(options NodeOptions) (map[string][]string, error) {
 		if err != nil {
 			return nil, err
 		}
-		pTopic := wprotocol.GetShardFromContentTopic(contentTopic, wprotocol.GenerationZeroShardsCount)
+		pTopic := wprotocol.GetShardFromContentTopic(contentTopic, uint16(options.ClusterID), wprotocol.GenerationZeroShardsCount)
 		if _, ok := pubSubTopicMap[pTopic.String()]; !ok {
 			pubSubTopicMap[pTopic.String()] = []string{}
 		}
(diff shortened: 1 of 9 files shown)


chaitanyaprem:
@richard-ramos :Started thinking this would be straight-forward  , but as the protocol code doesn't have access to WakuNode, this is getting more complicated and would involve API changes that would affect status codebase.
So, wanted to run it by you if there is a simpler way to achieve this.

richard-ramos:
Is Status planning to use autosharding at least in short/medium term? if not, I'd rather not have to do any changes on the API if there are no plans for its usage, specially since the plan is to replace go-waku by nwaku within status-go, so it's highly likely that the effort on supporting autosharding in go-waku would be wasted.

chaitanyaprem:
> Is Status planning to use autosharding at least in short/medium term? if not, I'd rather not have to do any changes on the API if there are no plans for its usage, specially since the plan is to replace go-waku by nwaku within status-go, so it's highly likely that the effort on supporting autosharding in go-waku would be wasted.

I don't think so that is in the immediate plan. I think they would move to nwaku before moving to autosharding.
So, let us not spend time on this then. We can take it up if the need comes at a later stage.

jm-clius:
> Wanted to confirm that status is not going to use auto-sharding

This seems to be the case, yes.

fryorcraken:
Let's assume that we would use nwaku if we were going to go autosharding.

Autosharding + RLN would be useful to scale 1:! chats. Not in the roadmap for 2024 so far. But using nwaku in Status desktop is.

PR #1067, opened 2024-03-22
feat: enable CORS for REST API

chaitanyaprem:


richard-ramos (changes requested):
Instead of adding a enableCors function in all endpoints, in https://github.com/waku-org/go-waku/blob/master/cmd/waku/server/rest/waku_rest.go#L41 add the following:
```go
	mux.Use(func(h http.Handler) http.Handler {
		fn := func(w http.ResponseWriter, r *http.Request) {
			w.Header().Set("Access-Control-Allow-Origin", "*")
		}
		return http.HandlerFunc(fn)
	})
```

chaitanyaprem:
> Instead of adding a enableCors function in all endpoints, in https://github.com/waku-org/go-waku/blob/master/cmd/waku/server/rest/waku_rest.go#L41 add the following:
> 
> ```go
> 	mux.Use(func(h http.Handler) http.Handler {
> 		fn := func(w http.ResponseWriter, r *http.Request) {
> 			w.Header().Set("Access-Control-Allow-Origin", "*")
> 		}
> 		return http.HandlerFunc(fn)
> 	})
> ```

Ah, i was wondering if there was a simpler way to do this. 
Thanks! this is way simple than doing it in every place.

richard-ramos (approved):

PR #1069, opened 2024-03-22
fix: handle discv5 udp port 0

richard-ramos:
# Description
While debugging a message reliability issue in status-go I noticed that the ENRs were lacking the UDP port
This was introduced in https://github.com/status-im/status-go/pull/4610, and in this PR Il update the UDP port used in the params with whatever value is obtained from ListenUDP


chaitanyaprem (approved):
LGTM

PR #1080, opened 2024-04-06
chore(nix): refactor, fix library packages

jakubgs:
Bit of a cleanup to make it more readable and also fix building of libraries.

Moving the actual build to `default.nix` makes `flake.nix` more readable.

### DIFF
--- a/ci/Jenkinsfile.nix-flake
+++ b/ci/Jenkinsfile.nix-flake
@@ -1,4 +1,4 @@
-library 'status-jenkins-lib@v1.7.0'
+library 'status-jenkins-lib@v1.9.3'
 
 pipeline {
   agent {
@@ -27,10 +27,7 @@ pipeline {
       stages {
         stage('Build') {
           steps { script {
-            sh("""#!/usr/bin/env bash
-              ${nix._sourceProfileInline()}
-              nix build --print-out-paths .#node
-            """)
+            nix.flake('node')
           } }
         }
         stage('Check') {
@@ -45,15 +42,12 @@ pipeline {
       stages {
         stage('Build') {
           steps { script {
-            sh("""#!/usr/bin/env bash
-              ${nix._sourceProfileInline()}
-              nix build --print-out-paths .#library
-            """)
+            nix.flake('static-library')
           } }
         }
         stage('Check') {
           steps {
-            sh 'ldd ./result/bin/c'
+            sh 'readelf -h ./result/bin/libgowaku.a'
           }
         }
       }
--- a/flake.nix
+++ b/flake.nix
@@ -11,11 +11,11 @@
       ];
       forAllSystems = f: nixpkgs.lib.genAttrs supportedSystems (system: f system);
 
-      nixpkgsFor = forAllSystems (system: import nixpkgs { inherit system; });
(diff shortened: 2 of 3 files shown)


jakubgs:
For some reason when i try to build `dynamic-library` I get:
```

File: flake.nix [hunk 1]
```diff
@@ -9,42 +9,37 @@
...
+          inherit self;
+          subPkgs = ["library/c"];
+          ldflags = ["-buildmode=c-archive"];
+          output = "libgowaku.a";
+        };
+        dynamic-library = buildPackage {
+          inherit self;
+          subPkgs = ["library/c"];
+          ldflags = ["-buildmode=c-shared"];
+          output = "libgowaku.a";
```
richard-ramos, line 37: Depending on the OS, it should be `.so` / `.dll` / `.dylib`
https://github.com/waku-org/go-waku/blob/master/Makefile#L19-L30

ci/Jenkinsfile.nix-flake
```diff
@@ -47,13 +47,13 @@ pipeline {
...
               ${nix._sourceProfileInline()}
-              nix build --print-out-paths .#library
+              nix build --print-out-paths .#dynamic-library
             """)
           } }
         }
         stage('Check') {
           steps {
-            sh 'ldd ./result/bin/c'
+            sh 'ldd ./result/bin/libgowaku.a'
```
richard-ramos, line 56: See other comment.

flake.nix
jakubgs: Good point.

richard-ramos:
@jakubgs is it possible to run `go env GOGCCFLAGS` right before the build is done? i run it locally and have the following:
```
-fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build621380261=/tmp/go-build -gno-record-gcc-switches
```
I'm thinking this `-fPIC` in the GOGCCFLAGS variable might be missing in the nix build.

jakubgs:
I have tried applying `GOGCCFLAGS = "-fPIC"` in the Nix derivation, but it did not work.

I have also tried `-Wl,-Bsymbolic` but that did not help either. Maybe I'm applying them wrong. Pushed.

jakubgs:
>is it possible to run go env GOGCCFLAGS right before the build is done? 

Yes, I can add this in the build derivation:
```nix
  preBuild = ''
    go env GOGCCFLAGS
  '';
```
And I can then get:
```
 -fPIC -m64 -pthread -Wl,--no-gc-sections -fmessage-length=0 -fdebug-prefix-map=/build/go-build1875450279=/tmp/go-build -gno-record-gcc-switches
```
So it seems ot me like it is there, and yet it still fails with:
```
relocation R_X86_64_TPOFF32 against `runtime.tlsg' can not be used when making a shared object; recompile with -fPIC
```

ci/Jenkinsfile.nix-flake
jakubgs: The Jenkinsfile runs only on Linux for now so it's fine.

flake.nix
jakubgs: Fixed.

richard-ramos (approved):

PR #1085, opened 2024-04-10
fix: logout discover dead lock

qfrank:
Pls see relate [issue](https://github.com/status-im/status-go/issues/4527) include comments as reference. This PR will fix the issue.

NOTE: creating a new goroutine for every call to PushToChan might introduce overhead, especially under high load. I'm not sure if we can use sync/atomic for simple state updates, or maybe redesigning the service to minimize shared state and reduce the need for locks. But I think this PR is a good start entry to discuss 🙂

### DIFF
--- a/waku/v2/service/common_discovery_service.go
+++ b/waku/v2/service/common_discovery_service.go
@@ -49,8 +49,6 @@ func (sp *CommonDiscoveryService) GetListeningChan() <-chan PeerData {
 	return sp.channel
 }
 func (sp *CommonDiscoveryService) PushToChan(data PeerData) bool {
-	sp.RLock()
-	defer sp.RUnlock()
 	if err := sp.ErrOnNotRunning(); err != nil {
 		return false
 	}
--- a/waku/v2/service/common_service.go
+++ b/waku/v2/service/common_service.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"sync/atomic"
 )
 
 // this is common layout for all the services that require mutex protection and a guarantee that all running goroutines will be finished before stop finishes execution. This guarantee comes from waitGroup all one has to use CommonService.WaitGroup() in the goroutines that should finish by the end of stop function.
@@ -12,7 +13,7 @@ type CommonService struct {
 	cancel  context.CancelFunc
 	ctx     context.Context
 	wg      sync.WaitGroup
-	started bool
+	started atomic.Bool
 }
 
 func NewCommonService() *CommonService {
@@ -28,13 +29,13 @@ func NewCommonService() *CommonService {
 func (sp *CommonService) Start(ctx context.Context, fn func() error) error {
 	sp.Lock()
 	defer sp.Unlock()
-	if sp.started {
+	if sp.started.Load() {
 		return ErrAlreadyStarted
 	}
-	sp.started = true
+	sp.started.Store(true)
 	sp.ctx, sp.cancel = context.WithCancel(ctx)
 	if err := fn(); err != nil {
-		sp.started = false
+		sp.started.Store(false)
 		sp.cancel()
 		return err
 	}
@@ -48,18 +49,18 @@ var ErrNotStarted = errors.New("not started")
 func (sp *CommonService) Stop(fn func()) {
 	sp.Lock()
 	defer sp.Unlock()
-	if !sp.started {
+	if !sp.started.Load() {
 		return
 	}
 	sp.cancel()
 	fn()
 	sp.wg.Wait()
-	sp.started = false
+	sp.started.Store(false)
 }
 
 // This is not a mutex protected function, it is up to the caller to use it in a mutex protected context
 func (sp *CommonService) ErrOnNotRunning() error {
-	if !sp.started {
+	if !sp.started.Load() {
 		return ErrNotStarted
 	}
 	return nil


richard-ramos:
I'm confused. The code as it is right now is cancelable. Do check this small example in go playground that simulates the behavior of the current logic

```go
package main

import (
	"context"
	"fmt"
	"time"
)

type PeerData = string

func main() {

	ctx, cancel := context.WithCancel(context.Background())
	channel := make(chan PeerData)

	// This simulates a blocking read
	go func() {

		time.Sleep(10 * time.Second)
		thePeerData := <-channel
		fmt.Println("received peerData " + thePeerData)
	}()

	// This simulates logging out
	go func() {
		time.Sleep(2 * time.Second)
		cancel()
	}()

	somePeerData := PeerData("alice")

	select {
	case channel <- somePeerData:
		fmt.Println("Pushed peerdata into channel succesfully")

	case <-ctx.Done():
		fmt.Println("Cancel was executed")
	}
}
```

It should print `Cancel was executed`

cammellos:
> I'm confused. The code as it is right now is cancelable. Do check this small example in go playground that simulates the behavior of the current logic
> 
> ```go
> package main
> 
> import (
> 	"context"
> 	"fmt"
> 	"time"
> )
> 
> type PeerData = string
> 
> func main() {
> 
> 	ctx, cancel := context.WithCancel(context.Background())
> 	channel := make(chan PeerData)
> 
> 	// This simulates a blocking read
> 	go func() {
> 
> 		time.Sleep(10 * time.Second)
> 		thePeerData := <-channel
> 		fmt.Println("received peerData " + thePeerData)
> 	}()
> 
> 	// This simulates logging out
> 	go func() {
> 		time.Sleep(2 * time.Second)
> 		cancel()
> 	}()
> 
> 	somePeerData := PeerData("alice")
> 
> 	select {
> 	case channel <- somePeerData:
> 		fmt.Println("Pushed peerdata into channel succesfully")
> 
> 	case <-ctx.Done():
> 		fmt.Println("Cancel was executed")
> 	}
> }
> ```
> 
> It should print `Cancel was executed`

@qfrank @richard-ramos one thing I noticed, is that the channel is non buffered and it's used outside the struct.
I think a better way is to have the channel buffered, and have the code publishing on the channel non-blocking, otherwise the code will lock if the listener (which you don't have control over since it's outside the struct), doesn't empty the channel.
Could that be the issue with the deadlock?
i.e the listener stops listening to the channel, before context is done, but after we try to publish

cammellos:
> > I'm confused. The code as it is right now is cancelable. Do check this small example in go playground that simulates the behavior of the current logic
> > ```go
> > package main
> > 
> > import (
> > 	"context"
> > 	"fmt"
> > 	"time"
> > )
> > 
> > type PeerData = string
> > 
> > func main() {
> > 
> > 	ctx, cancel := context.WithCancel(context.Background())
> > 	channel := make(chan PeerData)
> > 
> > 	// This simulates a blocking read
> > 	go func() {
> > 
> > 		time.Sleep(10 * time.Second)
> > 		thePeerData := <-channel
> > 		fmt.Println("received peerData " + thePeerData)
> > 	}()
> > 
> > 	// This simulates logging out
> > 	go func() {
> > 		time.Sleep(2 * time.Second)
> > 		cancel()
> > 	}()
> > 
> > 	somePeerData := PeerData("alice")
> > 
> > 	select {
> > 	case channel <- somePeerData:
> > 		fmt.Println("Pushed peerdata into channel succesfully")
> > 
> > 	case <-ctx.Done():
> > 		fmt.Println("Cancel was executed")
> > 	}
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It should print `Cancel was executed`
> 
> @qfrank @richard-ramos one thing I noticed, is that the channel is non buffered and it's used outside the struct. I think a better way is to have the channel buffered, and have the code publishing on the channel non-blocking, otherwise the code will lock if the listener (which you don't have control over since it's outside the struct), doesn't empty the channel. Could that be the issue with the deadlock? i.e the listener stops listening to the channel, before context is done, but after we try to publish

ah nevermind, I noticed that's addressed in @richard-ramos , but one other thing I noticed, is that we export mutexes from `commontService`, I think if we do that, all bets are off, mutexes should only be ideally used for managing internal state, most likely the cause is due to both `commonService` and `CommonDiscoveryServices` sharing a mutex, seems like the best way is to re-think how the two coordinate and avoid sharing locks

cammellos:
> runDiscoveryV5Loop

WaitGroup is exported in other places:
```
vendor/github.com/waku-org/go-waku/waku/v2/discv5/discover.go
vendor/github.com/waku-org/go-waku/waku/v2/rendezvous/rendezvous.go
```

and it's combined with the lock as you pointed out, we should definitely try to avoid this pattern, we should keep wait group/locks/etc coordination for internal usage or top-down with a single agent, otherwise it's likely going to be a whack-a-mole

cammellos:
instinctively, I don't think that would work, if there's truly a race condition, sync/atomic will only protect that value and not the writing to channels. 
rather, I think having a lock specific to the channel is better,  since it looks like now it's used to cover multiple states, `started` and the `channel`, and that might be the issue, but i'd have to spend far more time to understand it properly

cammellos:
> why we need protect the writing to channels? could you explain a little bit? thanks in advance! because i see no reason so that was my quick fix @cammellos

Writing to a closed channel would panic, that'd be my guess

PR #1089, opened 2024-04-15
fix(peer-manager): use backoff only on connection errors

### LINKED ISSUE #1086: bug: message won't be sent over from node1 to node2 with sharded topic subscription
**Description**
During TestStaticShardingLimits execution 1024 pubsub topics were subscribed on relay at node1 and node2.  Relay instance1 was used to publish one message on randomly selected pubsub topic out of those 1024 subscribed. Subscription on relay at node2 is used to check the message was received. It looks message was not received. I have added relay.IsSubscribed() check on both relays, and this check is passing. 

Similar scenario is working at TestWakuRelayStaticSharding with the difference message is sent from relay2 to relay1.

**To Reproduce**
1) Checkout https://github.com/waku-org/go-waku/pull/1060/commits/e1210c7ce3b5d3eb3d3ae1284855d465cfe4e1df
2) Apply patch from https://github.com/waku-org/go-waku/pull/1084
3) cd go-waku/waku/v2/node
4) go test -run TestStaticShardingLimits

**Expected behavior**
Message could be received on node2 after publishing on node1.

**Log**
```
--- FAIL: TestStaticShardingLimits (33.97s)
    utils.go:402:
        	Error Trace:	/Users/roman/sources/waku-org/go-waku/tests/utils.go:402
        	            				/usr/local/go/src/runtime/asm_arm64.s:1172
        	Error:      	Message timeout
        	Test:       	TestStaticShardingLimits
FAIL
```

richard-ramos:
# Description
While attempting to debug the issue #1086 I noticed that the backoff period was being applied both to succesful connections and unsuccesful, hence why the test case described in #1086 failed to connect to the only peer defined, because initially with discv5 there's a connection to the peer, and a immediate disconnection due to peers not being subscribed to any shard, but this connection attempt (which was succesful, even if there was an immediate disconnection afterwards) ended up increasing the backoff period.

## Issue

closes #1086

### DIFF
--- a/waku/v2/peermanager/peer_connector.go
+++ b/waku/v2/peermanager/peer_connector.go
@@ -193,7 +193,6 @@ func (c *PeerConnectionStrategy) canDialPeer(pi peer.AddrInfo) bool {
 	c.mux.Lock()
 	defer c.mux.Unlock()
 	val, ok := c.cache.Get(pi.ID)
-	var cachedPeer *connCacheData
 	if ok {
 		tv := val.(*connCacheData)
 		now := time.Now()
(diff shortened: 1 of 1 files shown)


romanzac (changes requested):
Thanks for the initial attempt to fix. When I applied your changes, I am getting message delivered at about 4 out of 5 times I run the TestStaticShardingLimits. Even when I remove lines 464, 465:
```
// Wait for discovery
time.Sleep(3 * time.Second)
```	

Besides this behavior, I have much worse outcome when I subscribe to all 1024 topics at the second relay r2. Lines 502-506 at https://github.com/waku-org/go-waku/blob/b562771c6965504bf897cb30a65a9883912367f4/waku/v2/node/wakunode2_test.go Subscriptions would hang.

chaitanyaprem (approved):
Good Find

LGTM

chaitanyaprem:
Let us merge this PR as well after resolving conflicts. This would definitely benefit some use-cases in status.

romanzac (approved):

PR #1092, opened 2024-04-17
chore: rename `GOBIN` as `GOCMD`

szkl:
# Description

The `GOBIN` environment variable is a reserved name. The conditional variable assignment operator would not set the path to the `go` command if `GOBIN` is set in the parent environment.

```
	GOBIN
		The directory where 'go install' will install a command.
```

from `go help environment`.

# Changes

- Rename the environment variable.

# Tests

## How to test

- Run the `make` command with and without setting the `GOBIN` environment variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ CC_PREFIX       	:= github.com/waku-org/go-waku
 
 SHELL := bash # the shell used internally by Make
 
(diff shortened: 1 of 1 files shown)


chaitanyaprem:
Thanks for catching and fixing this.

chaitanyaprem (approved):
LGTM

PR #1096, opened 2024-05-06
chore: set go-ethereum back to v1.10.26

richard-ramos:
This is to avoid having to maintain a separate branch for go1.19. 
It also bumps some deps

Sadly enough we have to add back the `replace` for go-ethereum for it to play nicely with status-go, but since the idea is for this to be replaced by nwaku eventually, i guess it's okay. We should still think of rebasing go-ethereum to a more recent version tho.


vitvly (approved):

richard-ramos:
ci/test fails due to https://github.com/waku-org/nwaku/pull/2675

richard-ramos:
> @qfrank: Do we have a planned date for this?

No planned date for this, it's something more of a long term plan.

PR #1097, opened 2024-05-07
feat: add pubsub topic to `WakuMessageKeyValue`

richard-ramos:
# Description
Adds the pubsub topic, to match https://github.com/vacp2p/rfc/commit/5d69ab58b34d32af99e60b109ad784ad1dd9e345


waku/v2/protocol/store/pb/storev3.proto
```diff
@@ -8,6 +8,7 @@ import "waku/message/v1/message.proto";
 message WakuMessageKeyValue {
   optional bytes message_hash = 1; // Globally unique key for a Waku Message
   optional waku.message.v1.WakuMessage message = 2; // Full message content as value
+  optional string pubsub_topic = 3;
```
chaitanyaprem, line 11: Maybe we should also add validation for presence of pubsubTopic at https://github.com/waku-org/go-waku/blob/19d27befd98baf2aac58c1884795b662effe3fdb/waku/v2/protocol/store/pb/validation.go#L80-L90

chaitanyaprem (approved):
LGTM, once the comment needs to be addressed

PR #1112, opened 2024-05-25
fix: allow peer-selection without pubsubTopics

chaitanyaprem:
# Description
Noticed that peer-selection always assumes supported pubsubTopics are known for a peer. 
Added a functionality to not use pubsubTopic filter as selection if it is not passed in selectionCriteria. 
But not sure if this is a valid peer selection mechanism.

Considering peers that are discovered in status-go seemed to get added without pubsubTopic list.

Note: I am not 100% sure if this is required. Will keep this in draft until i get clarity.


chaitanyaprem:
Closing this PR as this is not a required feature.

PR #1117, opened 2024-05-31
chore: store hash query example

kaichaosun:


examples/store-hash-query/main.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	var b pb.MessageHash
+	copy(b[:], msgHash)
+
+	var opts []store.RequestOption
+	requestID := protocol.GenerateRequestID()
+	opts = append(opts, store.WithRequestID(requestID))
+	opts = append(opts, store.WithPeer(addr.ID))
+
+	result, err := wakuNode.Store().QueryByHash(ctx, []pb.MessageHash{b}, opts...)
```
richard-ramos, line 88: If we're looking for a single hash, maybe it's better to use `Store().Exists`?

examples/store-hash-query/main.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	log.Info("store.queryByHash result", zap.String("requestID", hexutil.Encode(requestID)), zap.Any("messages", result.Messages()))
+
+	if len(result.Messages()) == 0 {
+		log.Warn("store.queryByHash result empty")
+		return
+	}
+	resHash := result.Messages()[0].MessageHash
+	log.Info("store.queryByHash result hash", zap.String("hash", hexutil.Encode(resHash[:])))
```
richard-ramos, line 101: Ideally since when you use `queryByHash`, you might receive more than one page of results, for the sake of having a more complete example, it's probably a good idea to do:
```go
for !result.IsComplete() {
    for _, m := range result.Messages() {
       resHash := m.WakuMessageHash()
        log.Info("store.queryByHash result hash", zap.Stringer("hash", resHash))
    }

    err := result.Next(ctx)
    if err != nil {
	log.Warn("store.next failed", zap.Error(err))
        return
    }
}
```

Do note that i'm using `WakuMessageHash()` which automatically converts the hash into a `WakuMessageHash` that has a `String()` function

examples/store-hash-query/main.go
```diff
@@ -0,0 +1,102 @@
...
+
+	time.Sleep(3 * time.Second)
+
+	msgHash, err := hexutil.Decode("0x081bf9e247a2c892a67a1e063ec2e7b4bb367e67e53aefce43e0a6465b0d3671")
+	if err != nil {
+		log.Error("Error decoding message hash", zap.Error(err))
+		return
+	}
+	var b pb.MessageHash
+	copy(b[:], msgHash)
```
richard-ramos, line 81:
```suggestion
        b := pb.ToMessageHash(msgHash)
```

examples/store-hash-query/main.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	wakuNode.DialPeerWithInfo(ctx, *addr)
+
+	time.Sleep(3 * time.Second)
+
+	msgHash, err := hexutil.Decode("0x081bf9e247a2c892a67a1e063ec2e7b4bb367e67e53aefce43e0a6465b0d3671")
+	if err != nil {
+		log.Error("Error decoding message hash", zap.Error(err))
+		return
+	}
```
richard-ramos, line 79: Is this hash a real value? if so, it will probably dissapear in 30d. and the example will fail.
I guess this is okay because at the end of the day, it's an example, but if you want the example to always work, consider publishing a message before the query.

examples/store-hash-query/main.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		log.Error("Could not convert hex into ecdsa key", zap.Error(err))
+		return
+	}
+	hostAddr, _ := net.ResolveTCPAddr("tcp", "0.0.0.0:0")
+	wakuNode, err := node.New(
+		node.WithPrivateKey(prvKey),
+		node.WithHostAddress(hostAddr),
+		node.WithNTP(),
+		node.WithWakuStore(),
```
chaitanyaprem, line 50: Doesn't this mean localNode is also acting as storeNode ? 

As per https://github.com/waku-org/go-waku/blob/269417c5e979aeec92d323bed280d23da95379ea/waku/v2/node/wakunode2.go#L197 it seems we take this flag to enable Store in ENR bitfield which indicates local node is a store node.

examples/store-hash-query/main.go
chaitanyaprem: Or ask for hash to be passed as an argument to the binary.

examples/store-hash-query/main.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	err = wakuNode.Start(ctx)
+	if err != nil {
+		log.Error("Error starting wakunode", zap.Error(err))
+		return
+	}
+
+	time.Sleep(3 * time.Second)
+
+	addr, err := peer.AddrInfoFromString("/dns4/store-02.ac-cn-hongkong-c.shards.test.status.im/tcp/30303/p2p/16Uiu2HAm9CQhsuwPR54q27kNj9iaQVfyRzTGKrhFmr94oD8ujU6P")
```
chaitanyaprem, line 66: Maybe this could optionally be taken as an arg if someone wants to override.

PR #1118, opened 2024-05-31
chore: allow setting custom logger name and change debug level to storeV3 client

richard-ramos:
This is because while working with the storenode message counter, the store requests were always being logged due to it being INFO, and they really did not add much useful information. Also added a small change to be able to setup the base logger name.


chaitanyaprem (approved):
LGTM

PR #1125, opened 2024-06-12
feat: online checker

richard-ramos:
# Description
Allows consumers of the library to setup the node online/offline state

### DIFF
--- a/waku/v2/node/wakunode2.go
+++ b/waku/v2/node/wakunode2.go
@@ -258,7 +258,7 @@ func New(opts ...WakuNodeOption) (*WakuNode, error) {
 	//Initialize peer manager.
 	w.peermanager = peermanager.NewPeerManager(w.opts.maxPeerConnections, w.opts.peerStoreCapacity, metadata, w.log)
 
-	w.peerConnector, err = peermanager.NewPeerConnectionStrategy(w.peermanager, discoveryConnectTimeout, w.log)
+	w.peerConnector, err = peermanager.NewPeerConnectionStrategy(w.peermanager, w.opts.onlineChecker, discoveryConnectTimeout, w.log)
 	if err != nil {
 		w.log.Error("creating peer connection strategy", zap.Error(err))
 	}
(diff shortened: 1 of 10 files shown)


waku/v2/node/wakunode2.go
```diff
@@ -957,3 +957,7 @@ func GetDiscv5Option(dnsDiscoveredNodes []dnsdisc.DiscoveredNode, discv5Nodes []
 func (w *WakuNode) ClusterID() uint16 {
 	return w.opts.clusterID
 }
+
+func (w *WakuNode) PeerConnector() *peermanager.PeerConnectionStrategy {
```
chaitanyaprem, line 961: Instead of exposing this specifically, i am wondering if we should have a method that indicates node connectivity change. Then this event can further be used for various actions such as stopping and starting discv5/pxClient, pausing/unpausing peer-connector and many more. 

Right now this logic is scattered across status-go and go-waku, which doesn't look clean.

waku/v2/peermanager/peer_connector.go
```diff
@@ -236,24 +240,35 @@ func (c *PeerConnectionStrategy) dialPeers() {
...
-			}
-			addrInfo := pd.AddrInfo
-
-			if addrInfo.ID == c.host.ID() || addrInfo.ID == "" ||
-				c.host.Network().Connectedness(addrInfo.ID) == network.Connected {
+		case <-c.Context().Done():
+			return
+		default:
+			if c.isPaused() {
+				time.Sleep(100 * time.Millisecond) // Sleep for a while to avoid busy waiting
```
chaitanyaprem, line 247: I am wondering if we should take an approach of read from channel and discard in case connector is paused. 

Anyways once connectivity is resumed, peers would be discovered and connectivity to them would be triggered again from either peerManager or status-go.

chaitanyaprem (commented):
I thought we had already addressed this, but looks like it is not.

waku/v2/peermanager/peer_connector.go
richard-ramos: Sounds good. That way i can avoid having this `default:` case

waku/v2/node/wakunode2.go
richard-ramos: I'll try to extract this to something that can be used both in here and also in the filter subscriptions which i saw also use a similar idea.

richard-ramos:
I do see however that in status-go, being online in the FilterManager has a different handling initially, so I'll add a function to be able to override the onlineChecker for filter.

waku/v2/peermanager/peer_connector.go
```diff
@@ -173,6 +180,10 @@ func (c *PeerConnectionStrategy) isPaused() bool {
 	return c.paused.Load()
 }
 
+func (c *PeerConnectionStrategy) SetPause(paused bool) {
```
vpavlin, line 183: Should the name be `SetPaused` to be consistent with the name of the attribute being set?

richard-ramos: yes

chaitanyaprem (approved):
LGTM

PR #1129, opened 2024-06-17
fix: do not write tcp port 0 and remove 100 chars length limit for multiaddresses

richard-ramos:
The reason why i removed the condition of checking for the length of the multiaddress is that we can just let the error handling determine whether writing to the enr was succesful or not.

### DIFF
--- a/waku/v2/node/localnode.go
+++ b/waku/v2/node/localnode.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ethereum/go-ethereum/p2p/enode"
 	"github.com/ethereum/go-ethereum/p2p/enr"
 	"github.com/libp2p/go-libp2p/core/event"
+	"github.com/multiformats/go-multiaddr"
 	ma "github.com/multiformats/go-multiaddr"
 	"github.com/waku-org/go-waku/waku/v2/protocol"
 	wenr "github.com/waku-org/go-waku/waku/v2/protocol/enr"
@@ -20,7 +21,6 @@ func (w *WakuNode) updateLocalNode(localnode *enode.LocalNode, multiaddrs []ma.M
 	var options []wenr.ENROption
 	options = append(options, wenr.WithUDPPort(udpPort))
 	options = append(options, wenr.WithWakuBitfield(wakuFlags))
-	options = append(options, wenr.WithMultiaddress(multiaddrs...))
 
 	if advertiseAddr != nil {
 		// An advertised address disables libp2p address updates
@@ -36,30 +36,36 @@ func (w *WakuNode) updateLocalNode(localnode *enode.LocalNode, multiaddrs []ma.M
 		// Using a static ip will disable endpoint prediction.
 		options = append(options, wenr.WithIP(ipAddr))
 	} else {
-		// We received a libp2p address update, but we should still
-		// allow discv5 to update the enr record. We set the localnode
-		// keys manually. It's possible that the ENR record might get
-		// updated automatically
-		ip4 := ipAddr.IP.To4()
-		ip6 := ipAddr.IP.To16()
-		if ip4 != nil && !ip4.IsUnspecified() {
-			localnode.SetFallbackIP(ip4)
-			localnode.Set(enr.IPv4(ip4))
-			localnode.Set(enr.TCP(uint16(ipAddr.Port)))
-		} else {
-			localnode.Delete(enr.IPv4{})
-			localnode.Delete(enr.TCP(0))
-			localnode.SetFallbackIP(net.IP{127, 0, 0, 1})
-		}
+		if ipAddr.Port != 0 {
+			// We received a libp2p address update, but we should still
+			// allow discv5 to update the enr record. We set the localnode
+			// keys manually. It's possible that the ENR record might get
+			// updated automatically
+			ip4 := ipAddr.IP.To4()
+			ip6 := ipAddr.IP.To16()
+			if ip4 != nil && !ip4.IsUnspecified() {
+				localnode.SetFallbackIP(ip4)
+				localnode.Set(enr.IPv4(ip4))
+				localnode.Set(enr.TCP(uint16(ipAddr.Port)))
+			} else {
+				localnode.Delete(enr.IPv4{})
+				localnode.Delete(enr.TCP(0))
+				localnode.SetFallbackIP(net.IP{127, 0, 0, 1})
+			}
 
-		if ip4 == nil && ip6 != nil && !ip6.IsUnspecified() {
-			localnode.Set(enr.IPv6(ip6))
-			localnode.Set(enr.TCP6(ipAddr.Port))
-		} else {
-			localnode.Delete(enr.IPv6{})
-			localnode.Delete(enr.TCP6(0))
+			if ip4 == nil && ip6 != nil && !ip6.IsUnspecified() {
+				localnode.Set(enr.IPv6(ip6))
+				localnode.Set(enr.TCP6(ipAddr.Port))
+			} else {
+				localnode.Delete(enr.IPv6{})
+				localnode.Delete(enr.TCP6(0))
+			}
 		}
 	}
+
+	// Writing the IP + Port has priority over writting the multiaddress which might fail or not
+	// depending on the enr having space
+	options = append(options, wenr.WithMultiaddress(multiaddrs...))
 
 	return wenr.Update(localnode, options...)
 }
@@ -228,8 +234,28 @@ func selectCircuitRelayListenAddresses(addresses []ma.Multiaddr) ([]ma.Multiaddr
 	return result, nil
 }
 
-func (w *WakuNode) getENRAddresses(addrs []ma.Multiaddr) (extAddr *net.TCPAddr, multiaddr []ma.Multiaddr, err error) {
+func filter0Port(addresses []ma.Multiaddr) ([]ma.Multiaddr, error) {
+	var result []ma.Multiaddr
+	for _, addr := range addresses {
+		portStr, err := addr.ValueForProtocol(ma.P_TCP)
+		if err != nil && !errors.Is(err, multiaddr.ErrProtocolNotFound) {
+			return nil, err
+		}
 
+		port, err := strconv.Atoi(portStr)
+		if err != nil {
+			return nil, err
+		}
+
+		if port != 0 {
+			result = append(result, addr)
+		}
+	}
+
+	return result, nil
+}
+
+func (w *WakuNode) getENRAddresses(addrs []ma.Multiaddr) (extAddr *net.TCPAddr, multiaddr []ma.Multiaddr, err error) {
 	extAddr, err = selectMostExternalAddress(addrs)
 	if err != nil {
 		return nil, nil, err
@@ -251,6 +277,11 @@ func (w *WakuNode) getENRAddresses(addrs []ma.Multiaddr) (extAddr *net.TCPAddr, 
 		multiaddr = append(multiaddr, circuitAddrs...)
 	} else {
 		multiaddr = append(multiaddr, wssAddrs...)
+	}
+
+	multiaddr, err = filter0Port(multiaddr)
+	if err != nil {
+		return nil, nil, err
 	}
 
 	return
--- a/waku/v2/node/localnode.go
+++ b/waku/v2/node/localnode.go
@@ -67,7 +67,7 @@ func (w *WakuNode) updateLocalNode(localnode *enode.LocalNode, multiaddrs []ma.M
 	// depending on the enr having space
 	options = append(options, wenr.WithMultiaddress(multiaddrs...))
 
-	return wenr.Update(localnode, options...)
+	return wenr.Update(w.log, localnode, options...)
 }
 
 func isPrivate(addr *net.TCPAddr) bool {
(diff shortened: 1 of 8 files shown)


waku/v2/protocol/enr/localnode.go
```diff
@@ -120,9 +124,7 @@ func writeMultiaddressField(localnode *enode.LocalNode, addrAggr []multiaddr.Mul
 		fieldRaw = append(fieldRaw, maRaw...)
 	}
 
-	if len(fieldRaw) != 0 && len(fieldRaw) <= 100 { // Max length for multiaddr field before triggering the 300 bytes limit
-		localnode.Set(enr.WithEntry(MultiaddrENRField, fieldRaw))
-	}
+	localnode.Set(enr.WithEntry(MultiaddrENRField, fieldRaw))
 
 	// This is to trigger the signing record err due to exceeding 300bytes limit
 	_ = localnode.Node()
```
chaitanyaprem, line 130: Shouldn't we return the error if size goes beyond 300 bytes?

waku/v2/node/localnode.go
```diff
@@ -253,6 +279,11 @@ func (w *WakuNode) getENRAddresses(addrs []ma.Multiaddr) (extAddr *net.TCPAddr,
 		multiaddr = append(multiaddr, wssAddrs...)
```
chaitanyaprem, line 279: I am wondering if we append wss addresses, can any external node connect using it? Don't we need to have ceritficates setup in order for that to happen? Or is that only a browser restriction? I am not too clear on this

chaitanyaprem (commented):
Looks like tests and compilation is failing, that needs fixing

waku/v2/node/localnode.go
fryorcraken: if you dont have ssl setup then there is no point pushing the wss multiaddr.
This may only be useful when go-waku is used as a service node, which is out of scope.

chaitanyaprem: When you say service node, for which services? 
Because status desktop is acting as a  service node for filter, lightpush and peerexchange.

richard-ramos: I have seen this `wss` addresses being obtained automagically when using circuit relay. The node not only receives tcp addresses from the circuit relay server, but also, if that server supports wss, your node will also end up with a `wss` address. The connection will still happen because the `wss` in the multiaddress does not refer to your node, it refers to the circuit relay server.

richard-ramos:
@chaitanyaprem I fixed a race codnition in this PR that is probably affecting your other open PR. You can cherry pick that commit if needed to get past CI

chaitanyaprem:
I was also wondering why i am getting a lot of NO_RESERVATION errors when i try to connect via circuit-relay. Could it be possible that nodes announcing circuit-relay addresses via a peer are not reserving themselves with the peer? 
Was looking at libp2p circuit-relay example and the node that wants to expose a circuit-relay address is supposed to reserve itself with circuit-relay peer. https://github.com/libp2p/go-libp2p/blob/6cebdd88366696a19b61dc209768c932a8f1ec4b/examples/relay/main.go#L110-L114

`ERROR[06-19|12:38:59.878|github.com/waku-org/go-waku/waku/v2/protocol/filter/client.go:405]                          Failed to subscribe                      pubSubTopic=/waku/2/rs/16/32 contentTopics=[/waku/1/0xfce8db73/rfc26] error="failed to dial: failed to dial 16Uiu2HAm2ZckNiQUAKbHAmokqxafm8FftwXch6aNQwrMiEJX1B9A: all dials failed\n  * [/ip4/95.216.204.65/tcp/60000/p2p/16Uiu2HAmLktdCHRc63Br2YBX8LihJBiBZPkqymt5HRheeoUXGHwP/p2p-circuit] error opening relay circuit: NO_RESERVATION (204)\n  * [/ip4/88.156.142.133/tcp/61446] dial tcp4 0.0.0.0:51978->88.156.142.133:61446: i/o timeout"`


I don't see the use of reserve function in go-waku anywhere. Is my understanding correct @richard-ramos or am i missing something?

waku/v2/protocol/enr/localnode.go
chaitanyaprem: Hmm, very odd.

richard-ramos:
> @chaitanyaprem: I don't see the use of reserve function in go-waku anywhere. 

It's because the reservation process is handled by Autorelay in https://github.com/libp2p/go-libp2p/blob/6cebdd88366696a19b61dc209768c932a8f1ec4b/p2p/host/autorelay/relay_finder.go#L610 which in go-waku is setup here: https://github.com/waku-org/go-waku/blob/master/waku/v2/node/wakunode2.go#L217-L242

In the example from go-libp2p Autorelay is not used, I guess to ilustrate what's the process to follow to use circuit relay manually.

chaitanyaprem:
> > @chaitanyaprem: I don't see the use of reserve function in go-waku anywhere.
> 
> It's because the reservation process is handled by Autorelay in https://github.com/libp2p/go-libp2p/blob/6cebdd88366696a19b61dc209768c932a8f1ec4b/p2p/host/autorelay/relay_finder.go#L610 which in go-waku is setup here: https://github.com/waku-org/go-waku/blob/master/waku/v2/node/wakunode2.go#L217-L242
> 
> In the example from go-libp2p Autorelay is not used, I guess to ilustrate what's the process to follow to use circuit relay manually.

Got it, makes sense. 

But in that case i am wondering why connectivity fails with NO_RESERVATION error towards peers that advertise circuit-relay. Will require more testing to understand. Hopefully the peer address reported in ENR and circuit-relay reservation are always in sync. 
What if circuit-relay node doesn't accept reservation, will we still consider that address and add it to ENR?

richard-ramos:
> But in that case i am wondering why connectivity fails with NO_RESERVATION error towards peers that advertise circuit-relay. Will require more testing to understand. Hopefully the peer address reported in ENR and circuit-relay reservation are always in sync.

We have to test, but I'm thinking the problem is that the escenario we are seeing is the following:

1. Alice reserves with a circuit relay server
2. Alice publishes her ENR
3. Alice goes offline or loses connection to the circuit relay server
4. Circuit relay server eliminates the reserve since Alice is no longer online 
5. Bob asks for peers using discv5 / peer exchange and receives Alice's (routing tables can take some time to gc a node)
6. Bob tries to connect to Alice's circuit relay multiaddress
7. Circuit relay server receives the connection attempt and complains that there's no reserve because it no longer has a connection to Alice.

chaitanyaprem:
> > But in that case i am wondering why connectivity fails with NO_RESERVATION error towards peers that advertise circuit-relay. Will require more testing to understand. Hopefully the peer address reported in ENR and circuit-relay reservation are always in sync.
> 
> We have to test, but I'm thinking the problem is that the escenario we are seeing is the following:
> 
>     1. Alice reserves with a circuit relay server
> 
>     2. Alice publishes her ENR
> 
>     3. Alice goes offline or loses connection to the circuit relay server
> 
>     4. Circuit relay server eliminates the reserve since Alice is no longer online
> 
>     5. Bob asks for peers using discv5 / peer exchange and receives Alice's (routing tables can take some time to gc a node)
> 
>     6. Bob tries to connect to Alice's circuit relay multiaddress
> 
>     7. Circuit relay server receives the connection attempt and complains that there's no reserve because it no longer has a connection to Alice.

True that is also possible.

richard-ramos:
@chaitanyaprem I fixed the issues and got green lights in the CI.
I only need your seal of approval :)

waku/v2/protocol/enr/localnode.go
```diff
@@ -120,9 +129,7 @@ func writeMultiaddressField(localnode *enode.LocalNode, addrAggr []multiaddr.Mul
 		fieldRaw = append(fieldRaw, maRaw...)
 	}
 
-	if len(fieldRaw) != 0 && len(fieldRaw) <= 100 { // Max length for multiaddr field before triggering the 300 bytes limit
```
chaitanyaprem, line 123: I am wondering if we need to have this check or some check for ENR length crossing 300 bytes, because i ran this locally with status-desktop and noticed my ENR after address update with circuit-relay to cross 300 bytes and no error was reported. 

`enr:-OO4QG48MPpNEqllrIT-BZ_KEELMSAyigEMAc279YNCkomB0GtsiXlwv2-dKAwZUV2qdR_zsxUKb_0WX3faYqMYzIuWGAZAz0qL2gmlkgnY0gmlwhMCoAQmKbXVsdGlhZGRyc7QAMgSnYxMvBnZfpQMnACUIAh`

Screenshot 2024-06-20 at 9 44 35 AM

waku/v2/node/localnode.go
```diff
@@ -228,8 +234,28 @@ func selectCircuitRelayListenAddresses(addresses []ma.Multiaddr) ([]ma.Multiaddr
...
+		if port != 0 {
+			result = append(result, addr)
+		}
+	}
+
+	return result, nil
+}
+
+func (w *WakuNode) getENRAddresses(addrs []ma.Multiaddr) (extAddr *net.TCPAddr, multiaddr []ma.Multiaddr, err error) {
 	extAddr, err = selectMostExternalAddress(addrs)
```
chaitanyaprem, line 259: Also i was going through waku ENR spec and noticed that practically we can include upto 3 multiaddresses. Maybe we should try to include other addresses as well in that case rather than only the single most external address?

https://github.com/waku-org/specs/blob/master/standards/core/enr.md#multiaddrs-enr-key

richard-ramos: We try to include as many multiaddresses as possible: https://github.com/waku-org/go-waku/blob/795322a19665ac48ca1f5cf725f64b1fd1e67808/waku/v2/protocol/enr/localnode.go#L33-L54

The most external address is used only to populate the `tcp` and `ip` keys:
> For raw TCP & UDP connections details, [EIP-778](https://eips.ethereum.org/EIPS/eip-778) pre-defined keys SHOULD be used; The keys tcp, udp, ip (and tcp6, udp6, ip6 for IPv6) are enough to convey all necessary information;

waku/v2/protocol/enr/localnode.go
chaitanyaprem: Looks like it was copy paste issue .. this is complete ENR which seems valid.

`enr=enr:-OO4QG48MPpNEqllrIT-BZ_KEELMSAyigEMAc279YNCkomB0GtsiXlwv2-dKAwZUV2qdR_zsxUKb_0WX3faYqMYzIuWGAZAz0qL2gmlkgnY0gmlwhMCoAQmKbXVsdGlhZGRyc7QAMgSnYxMvBnZfpQMnACUIAhIhAt60bRUEoHNuLlnsM12sU2PIQwBwfLIJ8a_ZPEY2-RnkgnJzhwAQAgBAACCJc2VjcDI1NmsxoQJ5lIphhEl668B0TOuIzNRyTUJCA_s4xvhAvQVqNa7CYYN0Y3CC2zKDdWRwguqLhXdha3UyDQ`

richard-ramos: Ah interesting! how did you generate this? When the ENR is signed, if the data exceeds 300 bytes, it should have generated an error as described in https://github.com/waku-org/go-waku/pull/1134#issuecomment-2179832821
Something like:
```
panic: enode: can't sign record: record bigger than 300 bytes [recovered]
	panic: enode: can't sign record: record bigger than 300 bytes

goroutine 6 [running]:
testing.tRunner.func1.2({0x9340e0, 0xc0000b8180})
	/usr/local/go/src/testing/testing.go:1545 +0x238
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1548 +0x397
panic({0x9340e0?, 0xc0000b8180?})
	/usr/local/go/src/runtime/panic.go:914 +0x21f
github.com/ethereum/go-ethereum/p2p/enode.(*LocalNode).sign(0xc0000f2300)
	/home/richard/go/pkg/mod/github.com/status-im/go-ethereum@v1.10.25-status.4/p2p/enode/localnode.go:310 +0x51f
github.com/ethereum/go-ethereum/p2p/enode.(*LocalNode).Node(0xc0000f2300)
	/home/richard/go/pkg/mod/github.com/status-im/go-ethereum@v1.10.25-status.4/p2p/enode/localnode.go:122 +0x172
```

I think this error is complaining that the ENR is incomplete, because ENRs use RLP encoding, and the first byte iirc indicates the data length. If the number of bytes is less than the data len, I'd imagine an error like this would be thrown.

waku/v2/node/localnode.go
chaitanyaprem: Hmm, while i was testing locally..i had noticed that my LAN address was always getting added to ENR eventhough i have circuit-relay address. 
I am wondering why that could be happening when there are no other external addresses.

waku/v2/protocol/enr/localnode.go
richard-ramos: > Looks like it was copy paste issue .. this is complete ENR which seems valid.

you replied faster than my response :)

waku/v2/node/localnode.go
chaitanyaprem: Realized after our discussions that Waku multiaddrs key is not decoded by external tools. They seem to be deriving multiaddr based on other fields.

chaitanyaprem (approved):
Tested this locally and ENR generated doesn't fail to decode and doesn't include address with port 0.

Only minor thing we can probably improve is to not fill private IP address in ENR if present. It is not a major issue though as long as we include other addresses as part of Waku Multiaddrs field.

PR #1130, opened 2024-06-18
fix: filter out peers to be considered for circuit-relay 

chaitanyaprem:
While debugging conenctivity issues with peer-exchange, i had noticed that many peers have p2p circuit relay addresses formed with nodes that are not in the status fleet and probably part of TWN. As peers returned from the fleet using discv5 also include ones that are not part of cluster 16, this could be side-effect of that.
Connecting via such peers is causing failure, probably because metadata disconnects when we try to connect to them and hence circuit relay connection towards the final peer also fails. 

Hence adding a filter to the peers that are considered for circuit-relay so that peers from a different cluster are not considered as valid circuit-relay peers.

# Changes

- [ ] Circuit relay peers selection includes filtering of peers by ENR and those that have same clusterID as localnode.

### DIFF
--- a/waku/v2/node/wakunode2.go
+++ b/waku/v2/node/wakunode2.go
@@ -893,13 +893,26 @@ func (w *WakuNode) findRelayNodes(ctx context.Context) {
 		rand.Shuffle(len(peers), func(i, j int) { peers[i], peers[j] = peers[j], peers[i] })
 
 		for _, p := range peers {
+			pENR, err := w.Host().Peerstore().(wps.WakuPeerstore).ENR(p.ID)
+			if err != nil {
+				w.log.Error("could not get ENR for the peer, skipping for circuit-relay", zap.Stringer("peer", p.ID), zap.Error(err))
+				continue
+			}
+			rs, err := enr.RelayShardList(pENR.Record())
+			if err != nil {
+				w.log.Error("could not get shard info for the peer from ENR, skipping for circuit-relay", zap.Stringer("peer", p.ID), zap.Error(err))
+				continue
+			}
+			if rs.ClusterID != w.ClusterID() {
+				w.log.Debug("clusterID mismatch for the peer, skipping for circuit-relay", zap.Stringer("peer", p.ID), zap.Error(err))
+				continue
+			}
 			info := w.Host().Peerstore().PeerInfo(p.ID)
 			supportedProtocols, err := w.Host().Peerstore().SupportsProtocols(p.ID, proto.ProtoIDv2Hop)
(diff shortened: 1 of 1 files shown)


chaitanyaprem:
@waku-org/nwaku-developers : This change is being done because we don't seem to have clusterID based filter at discv5 and probably peerExchange. i.e peers returned by discv5 can belong to different cluster which could be causing nodes to consider them as valid peers and try to use them for circuit relay or connect to them.

It would help if we can also filter out the peers before they get stored in discv5 cache and pxCache with clusterID. 
cc @jm-clius @fryorcraken

fryorcraken:
> It would help if we can also filter out the peers before they get stored in discv5 cache and pxCache with clusterID.
cc @jm-clius @fryorcraken

In terms of filtering before putting in discv5 cache, the question is whether we should split discv5 p2p network per cluster id. I think it's a question we should answer unanimously and then implement across all clients. And not start to optimize some clients for it.

https://discord.com/channels/1110799176264056863/1253580627597590538/1253588456471265300

Questions:
- Is metadata protocol run when doing waku peer exchange?
- if yes, then yes, makes sense to filter enr that have same cluster id than own before putting it in px cache

richard-ramos:
> Is metadata protocol run when doing waku peer exchange?

Yes. Metadata protocol is triggered for all connections regardless of the protocol being consumed - https://github.com/waku-org/go-waku/blob/8303c592d35ce31fd3b726988070802232580e85/waku/v2/protocol/metadata/waku_metadata.go#L264-L267

fryorcraken:
> if yes, then yes, makes sense to filter enr that have same cluster id than own before putting it in px cache

shall we do that? @Ivansete-status @waku-org/go-waku-developers

Ivansete-status:
> > if yes, then yes, makes sense to filter enr that have same cluster id than own before putting it in px cache
> 
> shall we do that? (nwaku as service node and go-waku in desktop side) @Ivansete-status @waku-org/go-waku-developers

That sounds great to me! We can do that if you agree @waku-org/go-waku-developers 

As a future reference for `nwaku` team, the change is straightforward in `nwaku`. The change should go in: https://github.com/waku-org/nwaku/blob/99149ea9dc252c8d6b6d563d0804c1c474a528db/waku/waku_peer_exchange/protocol.nim#L140

richard-ramos:
It's easy to implement as well in go-waku. I'll work on a PR and tag you, @Ivansete-status!

richard-ramos:
https://github.com/waku-org/go-waku/pull/1139

Issue #1132, opened 2024-06-19
bug: use max relay peer targets in peermanager


chaitanyaprem:
**Describe the bug**
As of now gossipSubDMin is used to determine whether we have to connect immediately to discovered peers or not. 
Also, peerManager doesn't actively discover and initiate connections if we have gossipSubDMin which is 4. 

In order to form a healthy relay mesh as explained https://github.com/status-im/status-go/issues/5378 , we need to use outRelayPeerTarget in peer-connector and peerManager

chaitanyaprem:
@richard-ramos , i am working on this one as anyways there is a need to modify peerManager to work for lightNodes as per https://github.com/waku-org/go-waku/issues/1114.

PR #1135, opened 2024-06-24
feat: modify peer-manager to consider relay target peers 

chaitanyaprem:
Also start refactor to work with lightMode.
Complete refactor will happen in separate PR to avoid blocking the changes for relay.

# Description
Addresses https://github.com/waku-org/go-waku/issues/1132

### DIFF
--- a/waku/v2/peermanager/peer_discovery.go
+++ b/waku/v2/peermanager/peer_discovery.go
@@ -55,7 +55,7 @@ func (pm *PeerManager) discoverOnDemand(cluster uint16,
 
 	wakuProtoInfo, ok := pm.wakuprotoToENRFieldMap[wakuProtocol]
 	if !ok {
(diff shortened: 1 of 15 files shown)


waku/v2/peermanager/peer_discovery.go
```diff
@@ -55,7 +55,7 @@ func (pm *PeerManager) discoverOnDemand(cluster uint16,
 
 	wakuProtoInfo, ok := pm.wakuprotoToENRFieldMap[wakuProtocol]
 	if !ok {
-		pm.logger.Error("cannot do on demand discovery for non-waku protocol", zap.String("protocol", string(wakuProtocol)))
+		pm.logger.Info("cannot do on demand discovery for non-waku protocol", zap.String("protocol", string(wakuProtocol)))
```
richard-ramos, line 58:
```suggestion
		pm.logger.Warn("cannot do on demand discovery for non-waku protocol", zap.String("protocol", string(wakuProtocol)))
```

richard-ramos (approved):
LGTM

PR #1139, opened 2024-06-26
feat: filter peers stored in cache by cluster-id in peer-exchange

richard-ramos:
# Description
If a clusterID is specified, the peers stored in the cache will only store peers that match the specified cluster-id based on the shards from their ENR.

See: https://github.com/waku-org/go-waku/pull/1130 for discussion about why this change is necessary


chaitanyaprem (approved):
LGTM

PR #1141, opened 2024-06-26
fix: ignore ws from circuit relay addresses, and allow non multiaddresses in `multiaddrs` ENR key

richard-ramos:
- WS / WSS addresses are not taken into account for circuit relay
- Allows storing arbitrary data in the `multiaddrs` ENR key. If a value is not recognized, it's ignored


chaitanyaprem (approved):
LGTM

PR #1145, opened 2024-07-01
chore: use waku-org/go-libp2p-pubsub

richard-ramos:


chaitanyaprem (approved):
LGTM

nix build error needs to be fixed though before merging

PR #1150, opened 2024-07-02
feat: use mesh peers instead of all peers for determining topic health

### LINKED ISSUE #1131: feat: use gossipsub mesh size to report topic health for relay
### Problem
As of now we are using number of connected peers belonging to a pubsubTopic to report topicHealth for relay.
But a more accurate way that indicates topicHealth would be the number of fullMesh gossipsub peers available.

This would also be useful in reporting peer stats in NodeManagement tab in status desktop. 
Refer discussion here https://discord.com/channels/1110799176264056863/1252128832631345163

### Suggested solution
Use gossipsub meshpeers to identify topic health.
Should we use overall connected peer count as well somehow to indicate it? (TBD)

### Acceptance criteria
A list of tasks that need to be done for the issue to be considered resolved.

richard-ramos:
Fixes #1131
I could not figure out the logic behind how `gs.mesh` is populated in go-lib2p2-pubsub, so in the meantime I disabled the test that check for connection changes

### DIFF
--- a/examples/basic-light-client/go.mod
+++ b/examples/basic-light-client/go.mod
@@ -8,6 +8,8 @@ replace github.com/waku-org/go-waku => ../..
 
 replace github.com/ethereum/go-ethereum v1.10.26 => github.com/status-im/go-ethereum v1.10.25-status.15
 
+replace github.com/libp2p/go-libp2p-pubsub v0.11.0 => github.com/waku-org/go-libp2p-pubsub v0.0.0-20240703191659-2cbb09eac9b5
+
 require (
 	github.com/ethereum/go-ethereum v1.10.26
 	github.com/libp2p/go-libp2p v0.35.0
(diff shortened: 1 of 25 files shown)


chaitanyaprem (approved):
LGTM

chaitanyaprem:
Looks like a datarace has been detected, better to address it before merging.

richard-ramos:
Hm, it seems that the way that go-libp2p-pubsub uses the `mesh` attribute in `GossipSubRouter` is prone to data races:

- In the [`heartbeat()`](https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L1454) function which is executed by a separate goroutine `heartBeatTimer()` 
- As well as calls done by Pubsub to add/remove peers

richard-ramos:
I modified go-libp2p-pubsub to implement something similar to ListPeers. That solved the data race

PR #1153, opened 2024-07-08
feat: E2E Reliability POC v1

shash256:
# Description
POC for e2e reliability which implements the [End-to-end reliability for scalable distributed logs](https://forum.vac.dev/t/end-to-end-reliability-for-scalable-distributed-logs/293) protocol proposed. 

# Note
- Uses peers to request missing messages instead of store node
- Ephemeral messages are not considered in this implementation
- Currently no distinguishing of probabilistic nature of ACK incase of bloom filter detection.
- Does not handle scenario when the received message is older than the bloom filter window

### DIFF
--- a/examples/chat2-reliable/README.md
+++ b/examples/chat2-reliable/README.md
@@ -1,57 +1,117 @@
-# Using the `chat2-reliable` application
+# chat2-reliable: A Reliable P2P Chat Application
 
 ## Background
 
-The `chat2-reliable` application is a basic command-line chat app using the [Waku v2 suite of protocols](https://specs.vac.dev/specs/waku/v2/waku-v2). It connects to a [fleet of test nodes](fleets.status.im) to provide end-to-end p2p chat capabilities. The Waku team is currently using this application for internal testing. If you want try our protocols, or join the dogfooding fun, follow the instructions below.
+`chat2-reliable` is an enhanced version of a basic command-line chat application that uses the [Waku v2 suite of protocols](https://specs.vac.dev/specs/waku/v2/waku-v2). This version implements an end-to-end reliability protocol to ensure message delivery and causal ordering in a distributed environment.
 
-## Preparation
+## Features
+
+- P2P chat capabilities using Waku v2 protocols
+- Implementation of e2e reliability protocol
+- Support for group chats and direct communication
+- Scalable to large groups (up to 10K participants)
+- Transport-agnostic design
+
+## E2E Reliability Protocol
+
+The e2e reliability protocol in `chat2-reliable` is an implementation of the proposal at [Vac Forum](https://forum.vac.dev/t/end-to-end-reliability-for-scalable-distributed-logs/293) and includes the following key features:
+
+1. **Lamport Clocks**: Each participant maintains a Lamport clock for logical timestamping of messages.
+
+2. **Causal History**: Messages include a short causal history (preceding message IDs) to establish causal relationships.
+
+3. **Bloom Filters**: A rolling bloom filter is used to track received message IDs and detect duplicates.
+
+4. **Lazy Pull Mechanism**: Missing messages are requested from peers when causal dependencies are unmet.
+
+5. **Eager Push Mechanism**: Unacknowledged messages are resent to ensure delivery.
+
+## Usage
+
+### Building the Application
+
 ```
 make
 ```
 
-## Basic application usage
+### Starting the Application
 
-To start the `chat2-reliable` application in its most basic form, run the following from the project directory
-
+Basic usage:
 ```
 ./build/chat2-reliable
 ```
 
-You may need to set DNS server if behind a VPN,
-
+With custom DNS server:
 ```
 ./build/chat2-reliable --dns-discovery-name-server 8.8.8.8
 ```
 
-## Specifying a static peer
+### In-chat Commands
 
-In order to connect to a *specific* node as [`relay`](https://specs.vac.dev/specs/waku/v2/waku-relay) peer, define that node's `multiaddr` as a `staticnode` when starting the app:
+- `/help`: Display available commands
+- `/connect`: Interactively connect to a new peer
+- `/peers`: Display the list of connected peers
 
+Example:
 ```
-./build/chat2-reliable -staticnode=/dns4/node-01.do-ams3.waku.test.statusim.net/tcp/30303/p2p/16Uiu2HAkykgaECHswi3YKJ5dMLbq2kPVCo89fcyTd38UcQD6ej5W
+/connect /ip4/127.0.0.1/tcp/58426/p2p/16Uiu5rGt2QDLmPKas9zpsBgtr5kRzk473s9wkKSWoYwfcY4Hco33
 ```
 
-This will bypass the random peer selection process and connect to the specified node.
+## Message Format
 
-## In-chat options
-
-| Command | Effect |
-| --- | --- |
-| `/help` | displays available in-chat commands |
-| `/connect` | interactively connect to a new peer |
-| `/nick` | change nickname for current chat session |
-| `/peers` | Display the list of connected peers |
-
-## `chat2-reliable` message protobuf format
-
-Each `chat2-reliable` message is encoded as follows
+Messages in `chat2-reliable` use the following protobuf format:
 
 ```protobuf
-message Chat2Message {
-  uint64 timestamp = 1;
-  string nick = 2;
-  bytes payload = 3;
+message Message {
+  string sender_id = 1;
+  string message_id = 2;
+  int32 lamport_timestamp = 3;
+  repeated string causal_history = 4;
+  string channel_id = 5;
+  bytes bloom_filter = 6;
+  string content = 7;
 }
 ```
 
-where `timestamp` is the Unix timestamp of the message, `nick` is the relevant `chat2-reliable` user's selected nickname and `payload` is the actual chat message being sent. The `payload` is the byte array representation of a UTF8 encoded string.
+## Implementation Details
+
+1. **Lamport Clocks**: Implemented in the `Chat` struct with methods to increment, update, and retrieve the timestamp.
+
+2. **Causal History**: Stored in the `CausalHistory` field of each message, containing IDs of recent preceding messages.
+
+3. **Bloom Filters**: Implemented as a `RollingBloomFilter` to efficiently track received messages and detect duplicates.
+
+4. **Message Processing**:
+   - Incoming messages are checked against the bloom filter for duplicates.
+   - Causal dependencies are verified before processing.
+   - Messages with unmet dependencies are stored in an incoming buffer.
+
+5. **Message Recovery**:
+   - Missing messages are requested from peers.
+   - A retry mechanism with exponential backoff is implemented for failed retrievals.
+
+6. **Conflict Resolution**:
+   - Messages are ordered based on Lamport timestamps and message IDs for consistency.
+
+7. **Periodic Tasks**:
+   - Buffer sweeps to process buffered messages and resend unacknowledged ones.
+   - Sync messages to maintain consistency across peers.
+
+## Testing
+
+The implementation includes various tests to ensure the reliability features work as expected:
+
+- Lamport timestamp correctness
+- Causal ordering of messages
+- Duplicate detection using bloom filters
+- Message recovery after network partitions
+- Concurrent message sending
+- Large group scaling
+- Eager push mechanism effectiveness
+- Bloom filter window functionality
+- Conflict resolution
+- New node synchronization
+
+```
+go test -v
+```


examples/chat2-reliable/chat.go
```diff
@@ -26,9 +28,14 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-// Chat represents a subscription to a single PubSub topic. Messages
-// can be published to the topic with Chat.Publish, and received
-// messages are pushed to the Messages channel.
+const (
+	maxMessageHistory   = 100
+	bloomFilterSize     = 10000
+	bloomFilterFPRate   = 0.01
```
jm-clius, line 34: Would be interesting to understand the behaviour of the underlying library here. Presumably these parameters affect the number of hash functions applied to each added item? I also assume that false positivity rates increase with the number of items added, so not sure how the underlying library keeps the FPRate to 0.01 when we keep adding items to the filter?

My intuition was that we'll have to completely recompute the bloom filter every `t` minutes to ensure that we keep a limited number of IDs in the filter while still covering enough history to provide a useful ACK.

examples/chat2-reliable/chat.go
```diff
@@ -26,9 +28,14 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-// Chat represents a subscription to a single PubSub topic. Messages
-// can be published to the topic with Chat.Publish, and received
-// messages are pushed to the Messages channel.
+const (
+	maxMessageHistory   = 100
```
jm-clius, line 32: Perhaps we also want to make the number of IDs kept in the causal history configurable?

examples/chat2-reliable/chat.go
```diff
@@ -137,20 +151,65 @@ func (c *Chat) receiveMessages() {
...
+		// Add to incoming buffer
+		c.incomingBuffer = append(c.incomingBuffer, msg)
+	}
+}
+
+func (c *Chat) checkCausalDependencies(msg *pb.Message) bool {
+	for _, depID := range msg.CausalHistory {
+		if !c.bloomFilter.Test([]byte(depID)) {
+			return false
+		}
```
jm-clius, line 197: Causal dependencies should preferably be checked against the actual local history (i.e. `messageHistory`), not the bloom filter over the history. For one, the bloom filter should only be over _recent_ history to keep the number of included items low (thus reducing fp rate) whereas causal history should cover "everything". Causal history should also be a deterministic mechanism, while bloom filter is probabilistic.

examples/chat2-reliable/chat.go
```diff
@@ -137,20 +151,65 @@ func (c *Chat) receiveMessages() {
...
+	}
+	return true
+}
+
+func (c *Chat) updateBloomFilter(messageID string) {
+	c.bloomFilter.Add([]byte(messageID))
```
jm-clius, line 203: Not sure how the underlying library works here, but my intuition is that we'll need to recreate the bloom filter periodically and not just add items forever (which would increase FP rate).

examples/chat2-reliable/chat.go
```diff
@@ -137,20 +151,65 @@ func (c *Chat) receiveMessages() {
...
+
+func (c *Chat) processReceivedMessage(msg *pb.Message) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	// Update Lamport timestamp
+	if msg.LamportTimestamp > c.lamportTimestamp {
+		c.lamportTimestamp = msg.LamportTimestamp
+	}
+	c.lamportTimestamp++
```
jm-clius, line 175: I need to think about this. Not sure it's necessary/desirable to increase the `lamportTimestamp` after every received message (certainly necessary before sending a new message). However, I can't currently think of a good reason _not_ to do this. 🤔

examples/chat2-reliable/chat.go
```diff
@@ -137,20 +151,65 @@ func (c *Chat) receiveMessages() {
...
+
```
jm-clius, line 176: I think we can review the ACK status of messages in the `outgoingBuffer` here based on the received `bloomFilter` _and_ the received `causalHistory` (if our outgoing message is in the received causal history it can deterministically be marked as "sent" and probabilistically so if included in the `bloomFilter`).

jm-clius (commented):
Great job! Good to see this taking shape. I've added some comments (which in fairness you may have planned to address in any case as this is still a `Draft`). Feel free to request full review when ready to merge. :)

examples/chat2-reliable/chat.go
```diff
@@ -137,20 +153,123 @@ func (c *Chat) receiveMessages() {
...
+		c.updateBloomFilter(msg.MessageId)
+
+		// Add to message history
+		c.addToMessageHistory(msg)
+
+		// Update received bloom filter for the sender
+		if msg.BloomFilter != nil {
+			receivedFilter := &bloom.BloomFilter{}
+			err := receivedFilter.UnmarshalBinary(msg.BloomFilter)
+			if err == nil {
```
vpavlin, line 194: Would we not want to log or handle the error? Also in golang you usually check for `err != nil` to handle the error, so this might be better to do that, log error and drop the assignement into `else`

examples/chat2-reliable/chat.go
```diff
@@ -259,28 +378,46 @@ func (c *Chat) parseInput() {
 }
 
 func (c *Chat) SendMessage(line string) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	c.lamportTimestamp++
```
vpavlin, line 384: Seems like this is the second occurence of increment of the timestamp - that seems like one of them being redundant

vpavlin:
To avoid having to setup RLN, we need to use a node supporting `DefaultPubsubTopic`. I am running a node, so you can use this command to test it

```
./build/chat2-reliable -l DEBUG  --dns-discovery=false --discv5-discovery=true  --discv5-enr-auto-update=true  --staticnode=/ip4/62.3.50.142/tcp/60000/p2p/16Uiu2HAmAEQc3SuHKYHjRfaFPy3evmskRN4yafhTn93PFdpPSqLQ --storenode=/ip4/62.3.50.142/tcp/60000/p2p/16Uiu2HAmAEQc3SuHKYHjRfaFPy3evmskRN4yafhTn93PFdpPSqLQ
```

examples/chat2-reliable/peer_retrieval.go
```diff
@@ -0,0 +1,191 @@
...
+	wpb "github.com/waku-org/go-waku/waku/v2/protocol/pb"
+	"github.com/waku-org/go-waku/waku/v2/protocol/relay"
+	"github.com/waku-org/go-waku/waku/v2/protocol/store"
+	"google.golang.org/protobuf/proto"
+)
+
+// below functions are specifically for peer retrieval of missing msgs instead of store
+func (c *Chat) doRequestMissingMessageFromPeers(messageID string) error {
+	peers := c.node.Host().Network().Peers()
+	for _, peerID := range peers {
```
vpavlin, line 24: Any particula reason why you request the message from **every** known peer?

examples/chat2-reliable/reliability.go
```diff
@@ -0,0 +1,458 @@
...
+		// Add to incoming buffer
+		c.addToIncomingBuffer(msg)
+	}
+}
+
+func (c *Chat) processBufferedMessages() {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	processed := make(map[string]bool)
```
vpavlin, line 175: How is this supposed to be used? I don't see any assignment into the map

examples/chat2-reliable/reliability.go
```diff
@@ -0,0 +1,458 @@
...
+
+	for _, msg := range c.incomingBuffer {
+		if processed[msg.MessageId] {
+			continue
+		}
+
+		missingDeps := c.checkCausalDependencies(msg)
+		if len(missingDeps) == 0 {
+			// Release the lock while processing the message
+			c.mutex.Unlock()
```
vpavlin, line 186: Hmm, I feel like this might mess up with the for loop in case the `incommingBuffer` gets changed while unlocked. I guess we need separate locs for these crutial attributes

examples/chat2-reliable/peer_retrieval.go
jm-clius: If I understand the logic correctly it will stop requesting once the first peer responds. Should be a reasonable approach.

examples/chat2-reliable/reliability.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+}
+
+func (c *Chat) processReceivedMessage(msg *pb.Message) {
+	// Check if the message is already in the bloom filter
+	if c.bloomFilter.Test(msg.MessageId) {
+		// Review ACK status of messages in the unacknowledged outgoing buffer
+		c.reviewAckStatus(msg)
+		return
+	}
```
jm-clius, line 141: It feels like I might be missing something obvious, but why only review ACK status of outgoing messages if the received ID is already in the bloom filter? Shouldn't it be the opposite, i.e. for every newly received message we want to check if it happens to ACK some of our outgoing messages?

examples/chat2-reliable/reliability.go
```diff
@@ -0,0 +1,458 @@
...
+
+	c.ui.ErrorMessage(fmt.Errorf("failed to retrieve missing message after %d attempts: %s", maxRetries, messageID))
+}
+
+func (c *Chat) checkCausalDependencies(msg *pb.Message) []string {
+	var missingDeps []string
+	for _, depID := range msg.CausalHistory {
+		if !c.bloomFilter.Test(depID) {
+			missingDeps = append(missingDeps, depID)
+		}
```
jm-clius, line 257: Probably OK for the POC, but why not check missingDeps against the actual `messageHistory` to avoid false positives? The bloom filter is also likely to cover a shorter span of history than the `messageHistory`, which places a new constraint on this mechanism - in fact, messages with dependencies older than the current bloom filter scope will never have their dependencies met.

examples/chat2-reliable/reliability.go
```diff
@@ -0,0 +1,458 @@
...
+
+	c.messageHistory = append(c.messageHistory, msg)
+	c.messageHistory = c.resolveConflicts(c.messageHistory)
+
+	if len(c.messageHistory) > maxMessageHistory {
+		c.messageHistory = c.messageHistory[len(c.messageHistory)-maxMessageHistory:]
+	}
+}
+
+func (c *Chat) resolveConflicts(messages []*pb.Message) []*pb.Message {
```
jm-clius, line 274: Unsure why we have to reprocess the entire `messageHistory`? Why not simply change the `append` function to consider only the incoming new message, add it at its lamport timestamp and only resolve any conflict if there is a matching timestamp entry already? There may be some go-language idiosyncrasies at play here that I'm missing, though. :)

examples/chat2-reliable/chat.go
```diff
@@ -259,28 +275,48 @@ func (c *Chat) parseInput() {
 }
 
 func (c *Chat) SendMessage(line string) {
-	tCtx, cancel := context.WithTimeout(c.ctx, 3*time.Second)
-	defer func() {
-		cancel()
-	}()
+	c.incLamportTimestamp()
```
jm-clius, line 278: Shouldn't the part of the send logic that updates reliability constructs be in the reliability.go layer rather than in the app?

examples/chat2-reliable/reliability.go
```diff
@@ -0,0 +1,458 @@
...
+		// Review ACK status of messages in the unacknowledged outgoing buffer
+		c.reviewAckStatus(msg)
+		return
+	}
+
+	// Update Lamport timestamp
+	c.updateLamportTimestamp(msg.LamportTimestamp)
+
+	// Update bloom filter
+	c.bloomFilter.Add(msg.MessageId)
```
jm-clius, line 147: Not important for a POC, but we should eventually consider what should happen here if the received message is older than the bloom filter window. This would also affect checking causal dependencies: what if a message in the incoming buffer has a dependency on a historical message that we do indeed eventually receive, but is too old to be included in our bloom filter? We want to make sure that any processed, received message do indeed complete pending causal dependency chains where they form a link.

jm-clius (approved):
Protocol logic LGTM. I've made a couple of comments, most of which are not critical for the POC to be launched and this PR to be merged. Great job! :)

examples/chat2-reliable/peer_retrieval.go
```diff
@@ -0,0 +1,191 @@
...
+		msg, err := c.requestMessageFromPeer(peerID, messageID)
+		if err == nil && msg != nil {
+			c.processReceivedMessage(msg)
+			return nil
+		}
+	}
+	return fmt.Errorf("no peers could provide the missing message")
```
richard-ramos, line 31:
```suggestion
	return errors.New("no peers could provide the missing message")
```

examples/chat2-reliable/peer_retrieval.go
richard-ramos, line 24: Should this be done concurrently? or is sequential ok?

examples/chat2-reliable/peer_retrieval.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func (c *Chat) requestMessageFromPeer(peerID peer.ID, messageID string) (*pb.Message, error) {
+	ctx, cancel := context.WithTimeout(c.ctx, 30*time.Second)
+	defer cancel()
+
+	stream, err := c.node.Host().NewStream(ctx, peerID, protocol.ID("/chat2-reliable/message-request/1.0.0"))
```
richard-ramos, line 38: Extract `protocol.ID("/chat2-reliable/message-request/1.0.0"` to a constant and use it in line 143 too

examples/chat2-reliable/peer_retrieval.go
```diff
@@ -0,0 +1,191 @@
...
+		return err
+	}
+
+	// Add a delay before closing the stream
+	time.Sleep(1 * time.Second)
+
+	err = stream.Close()
+	if err != nil {
+		return err
+	}
```
richard-ramos, line 84: Stream is also being closed in `requestMessageFromPeer`

examples/chat2-reliable/peer_retrieval.go
```diff
@@ -0,0 +1,191 @@
...
+
+	_, err = stream.Write(data)
+	if err != nil {
+		return err
+	}
+
+	// Add a delay before closing the stream
+	time.Sleep(1 * time.Second)
```
richard-ramos, line 79: Is this delay needed?

examples/chat2-reliable/peer_retrieval.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+// Helper functions for protobuf message reading/writing
+func writeProtobufMessage(stream network.Stream, msg proto.Message) error {
+	data, err := proto.Marshal(msg)
+	if err != nil {
+		return err
+	}
+
+	_, err = stream.Write(data)
```
richard-ramos, line 73: Why use stream.Write instead of:  ?
```go
writer := pbio.NewDelimitedWriter(stream)
err = writer.WriteMsg(msg)
if err != nil {
	return err
}
```

examples/chat2-reliable/peer_retrieval.go
```diff
@@ -0,0 +1,191 @@
...
+			}
+			return err
+		}
+		data = append(data, buffer[:n]...)
+	}
+
+	err = proto.Unmarshal(data, msg)
+	if err != nil {
+		return err
+	}
```
richard-ramos, line 110: Why use `stream.Read` instead of: ?
```go
reader := pbio.NewDelimitedReader(stream, math.MaxInt32)
err = reader.ReadMsg(msg)
if err != nil {
    return err
}
```

examples/chat2-reliable/peer_retrieval.go
```diff
@@ -0,0 +1,191 @@
...
+	request := &pb.MessageRequest{}
+	err := readProtobufMessage(stream, request)
+	if err != nil {
+		c.ui.ErrorMessage(fmt.Errorf("failed to read message request: %w", err))
+		return
+	}
+
+	c.mutex.Lock()
+	var foundMessage *pb.Message
+	for _, msg := range c.messageHistory {
```
richard-ramos, line 126: I wonder if it would be a good idea to modify `messageHistory` to use an ordered map. Might be an overkill tho

examples/chat2-reliable/peer_retrieval.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	x := store.MessageHashCriteria{
+		MessageHashes: []wpb.MessageHash{wpb.ToMessageHash(hash)},
+	}
+
+	peers, err := c.node.PeerManager().SelectPeers(peermanager.PeerSelectionCriteria{
+		SelectionType: peermanager.Automatic,
+		Proto:         store.StoreQueryID_v300,
+		PubsubTopics:  []string{relay.DefaultWakuTopic},
```
richard-ramos, line 162: Consider instead of using the default waku topic, to use instead some shard. Even though go-waku supports named sharding for now, named sharding is deprecated.

examples/chat2-reliable/reliability.go
```diff
@@ -0,0 +1,379 @@
...
+	}
+}
+
+func (c *Chat) resendMessage(msg *pb.Message, resendAttempts int) {
+	go func() {
+		delay := resendBaseDelay * time.Duration(1<<uint(resendAttempts))
+		if delay > maxResendDelay {
+			delay = maxResendDelay
+		}
+		time.Sleep(delay)
```
richard-ramos, line 298: Consider making this sleep 'cancellable':
```suggestion
select {
  case <-c.ctx.Done():
      return
  case <-time.After(delay):
     // do nothing
  }
```

richard-ramos (approved):

PR #1157, opened 2024-07-11
chore(wakunode2): add ability to specify PX options in wakunode2

vpavlin:
# Description
The PR https://github.com/waku-org/go-waku/pull/1043/files adds support for Rate Limiting in Peer Exchange.  This PR extends it by allowing configuration of the rate limiter in wakunode2.

When adding a simple test, I also noticed the PX Requests gets stuck waiting on response if it is rate limited - not sure if that is on purpose, but in case it is not, added a `stream.Reset()` to make it not block the requesting peer.

# Changes

- [x] Add `peerExchangeOptions` in `WakuNodeParameters`
- [x] Reset the stream when the request is rate limited

# Tests

Spin up 2 nodes, enable PX, make more requests than rate limit allows, observe error


waku/v2/protocol/peer_exchange/protocol.go
```diff
@@ -100,6 +100,9 @@ func (wakuPX *WakuPeerExchange) onRequest() func(network.Stream) {
 			wakuPX.metrics.RecordError(rateLimitFailure)
 			wakuPX.log.Error("exceeds the rate limit")
 			// TODO: peer exchange protocol should contain an err field
+			if err := stream.Reset(); err != nil {
+				wakuPX.log.Error("resetting connection", zap.Error(err))
```
chaitanyaprem, line 104: Let's not log an error here, because this is not an error condition on peerExchange service node side, maybe we can log debug.

chaitanyaprem (approved):
LGTM, one minor comment.

waku/v2/protocol/peer_exchange/protocol.go
vpavlin: Hmm this is an error coming out of `stream.Reset()` - what is the difference between this and all the other handling of `stream.Reset()` errors later? (I don't mind changing, just trying to understand the reasoning)

richard-ramos: Looking at other instances of `stream.Reset()` in the server side, we're using `Error` there.

richard-ramos (approved):

waku/v2/protocol/peer_exchange/protocol.go
chaitanyaprem: my bad, this is fine. 
i wanted to indicate if we should log an error in previous line above `101` when rate limit is exceeded. Because a node may get flooded with error logs if someone is trying to constantly use it despite rate limit error. 
Minor comment, can be addressed later as well.

PR #1158, opened 2024-07-11
feat: support for lightpush to use more than 1 peer

chaitanyaprem:
# Description
Add support to specify number of peers to published for lightpush.

Note that this is a very basic support and may require handling of complex scenarios such as at the SDK layer which can be taken up as separate PR if required
- what if 2 peers are specified and publish to 1 fails? we need to retry and ensure we don't select same peer that publish was success to
- what if lightpush publish fails to all peers, should we retry internally in an API layer in go-waku 2-3 times?

# Changes

- [x] new lightpush option to specify number of peers to publish to
- [x] updated test for lightpush to use 2 peers


waku/v2/protocol/lightpush/waku_lightpush_option.go
```diff
@@ -58,7 +66,7 @@ func WithPeer(p peer.ID) RequestOption {
 func WithPeerAddr(pAddr multiaddr.Multiaddr) RequestOption {
 	return func(params *lightPushRequestParameters) error {
 		params.peerAddr = pAddr
```
richard-ramos, line 68: Should the same behavior of appending to a list of addresses be added here?

waku/v2/protocol/lightpush/waku_lightpush.go
```diff
@@ -327,26 +328,42 @@ func (wakuLP *WakuLightPush) Publish(ctx context.Context, message *wpb.WakuMessa
...
-	if err != nil {
-		logger.Error("could not publish message", zap.Error(err))
-		return wpb.MessageHash{}, err
+	for _, response := range responses {
+		if response.GetIsSuccess() {
+			successCount++
+		} else {
+			if response.GetInfo() != "" {
+				errMsg += *response.Info
+			}
```
richard-ramos, line 357: Do you think it would make sense to indicate the peerId that generated the error?

richard-ramos (approved):
Very cool!

waku/v2/protocol/lightpush/waku_lightpush_option.go
chaitanyaprem: That is not required as it gets handled here already 

https://github.com/waku-org/go-waku/blob/23b23b02b42420013b5c7356e75b194c486cc216/waku/v2/protocol/lightpush/waku_lightpush.go#L279-L286

fryorcraken:
Is this used in status-go?

PR #1160, opened 2024-07-13
feat: lightclient err handling

### LINKED ISSUE #1124: feat: unsubscribe in case invalid push requests are received from server
### Problem
There are few scenarios noticed during dogfooding that Filter-Push is received for subscriptions LightNode no longer has. 

One scenario where this could happen is 
1. Filter Subscribe for contentTopics C1, C2 and C3 to node1. 
2. Internet is disconnected and reconnected again. 
3. Now C1 and C2 are subscribed to node2 whereas C3 is still subscribed to node1
4. Now node1 thinks we are subscribed to C1 and c2 still as we keep sending Filter-Ping.

### Suggested solution
When we receive an invalid push request, send a unsubscribe request to that peer for that specific contentTopic. 

### Alternatives considered
When we regain connectivity, send unsubscribeAll to all peers with which we had filter subscriptions before connectivityLoss. 

### Acceptance criteria
Have a test to validate this scenario.

chaitanyaprem:
# Description
some error handling for lightClients

# Changes

- [x] fixes #1124 
- [x] handle context cancellation in filter push request notification
- [x] generate different requestID for each lightpush peer

# Tests

### DIFF
--- a/waku/v2/protocol/lightpush/waku_lightpush.go
+++ b/waku/v2/protocol/lightpush/waku_lightpush.go
@@ -189,13 +189,6 @@ func (wakuLP *WakuLightPush) reply(stream network.Stream, responsePushRPC *pb.Pu
 
 // request sends a message via lightPush protocol to either a specified peer or peer that is selected.
 func (wakuLP *WakuLightPush) request(ctx context.Context, req *pb.PushRequest, params *lightPushRequestParameters, peer peer.ID) (*pb.PushResponse, error) {
-	if params == nil {
-		return nil, errors.New("lightpush params are mandatory")
-	}
-
-	if len(params.requestID) == 0 {
-		return nil, ErrInvalidID
-	}
 
 	logger := wakuLP.log.With(logging.HostID("peer", peer))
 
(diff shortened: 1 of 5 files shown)


waku/v2/protocol/lightpush/waku_lightpush.go
```diff
@@ -336,6 +329,7 @@ func (wakuLP *WakuLightPush) Publish(ctx context.Context, message *wpb.WakuMessa
 	for _, peerID := range params.selectedPeers {
 		wg.Add(1)
 		go func(id peer.ID) {
+			params.requestID = protocol.GenerateRequestID()
```
richard-ramos, line 332: I wonder if it wouldnt be best to send the same requestID to all selected peers.
Also, it's not  big deal, but since `params` is a pointer, due to this being executed in a go routine, it's possible that we'll end up with a race condition here (not super problematic tbh since it's only the requestID being modified, and maybe can be solved with
```
paramsValue := *params 
paramsValue.requestID = protocol.GenerateRequestID()
defer wg.Done()
response, err := wakuLP.request(ctx, req, &paramsValue, id)
```
)

richard-ramos (approved):

waku/v2/protocol/lightpush/waku_lightpush.go
chaitanyaprem: good catch....will fix it. 
wrt requestID, it felt that it makes it easier for tracing...hence wanted to have separate one for each peer..

PR #1166, opened 2024-07-23
fix: filter ping timeout and retry in case of failure

chaitanyaprem:
# Description
While going through filter ping code, had noticed that 
- ping timeout was set to PingInterval which is 1 minute and is too large to identify failures in case peer is down or network disconnected etc. So, reducing it to 5 seconds to catch errors faster
-  aligning with rfc https://github.com/waku-org/specs/blob/master/standards/application/req-res-reliability.md#regular-pings 
> -  added a second ping in case of failure before we give-up and move subscriptions to another peer. 
> This is as per  , and this has been implemented in js-waku as well.
> - Perform filter pings only when we are online

### DIFF
--- a/waku/v2/protocol/filter/filter_health_check.go
+++ b/waku/v2/protocol/filter/filter_health_check.go
@@ -8,6 +8,8 @@ import (
 	"go.uber.org/zap"
 )
 
+const PingTimeout = 5 * time.Second
+
 func (wf *WakuFilterLightNode) PingPeers() {
 	//Send a ping to all the peers and report their status to corresponding subscriptions
 	// Alive or not or set state of subcription??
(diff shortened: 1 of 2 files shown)


waku/v2/protocol/filter/filter_health_check.go
```diff
@@ -17,17 +19,23 @@ func (wf *WakuFilterLightNode) PingPeers() {
 }
 
 func (wf *WakuFilterLightNode) PingPeer(peer peer.ID) {
-	ctxWithTimeout, cancel := context.WithTimeout(wf.CommonService.Context(), wf.peerPingInterval)
+	ctxWithTimeout, cancel := context.WithTimeout(wf.CommonService.Context(), PingTimeout)
```
richard-ramos, line 22: Should we get rid of `wf.peerPingInterval` ?

richard-ramos (approved):

waku/v2/protocol/filter/filter_health_check.go
chaitanyaprem: that is still required as it is used in https://github.com/waku-org/go-waku/blob/8cfe01d7d4776c41053aa2c8ca435b58469730e0/waku/v2/protocol/filter/filter_health_check.go#L45

PR #1167, opened 2024-07-23
feat: ping lightpush peers

richard-ramos:
Ping the filter nodes the node is subscribed to

### DIFF
--- a/waku/v2/node/keepalive.go
+++ b/waku/v2/node/keepalive.go
@@ -23,7 +23,7 @@ const maxAllowedPingFailures = 2
 // the peers if they don't reply back
 const sleepDetectionIntervalFactor = 3
 
-const maxPeersToPing = 10
+const maxPeersToPingPerProtocol = 10
 
 const maxAllowedSubsequentPingFailures = 2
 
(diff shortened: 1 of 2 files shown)


waku/v2/node/keepalive.go
```diff
@@ -59,13 +59,22 @@ func (w *WakuNode) startKeepAlive(ctx context.Context, randomPeersPingDuration t
...
+					for _, p := range w.Relay().PubSub().ListPeers(t) {
+						relayPeersSet[p] = struct{}{}
+					}
 				}
+				peersToPing = append(peersToPing, maps.Keys(relayPeersSet)...)
 			}
-			peersToPing = maps.Keys(relayPeersSet)
+
+			// We ping all filter nodes
+			filterPeersSet := make(map[peer.ID]struct{})
```
chaitanyaprem, line 73: Since we are pinging filter peers in shorter interval these are not required. 
Also, shouldn't this be in else condition of `w.opts.enableRelay` as relay and filter will never be enabled together?

waku/v2/node/keepalive.go
```diff
@@ -81,36 +90,45 @@ func (w *WakuNode) startKeepAlive(ctx context.Context, randomPeersPingDuration t
...
+					if peerLen > len(relayPeers) {
+						peerLen = len(relayPeers)
+					}
+					peersToPing = append(peersToPing, relayPeers[0:peerLen]...)
 				}
-				peersToPing = append(peersToPing, relayPeers[0:peerLen]...)
 			}
 
+			// We also ping all filter nodes
+			filterPeersSet := make(map[peer.ID]struct{})
```
chaitanyaprem, line 126: shouldn't this be in else condition of `w.opts.enableRelay`?

chaitanyaprem:
@richard-ramos can we address comments and merge this?

waku/v2/node/keepalive.go
richard-ramos: go-waku does allow a node to be a filter client and also use relay at the same time (even if it does not make sense :thinking: )

chaitanyaprem: Ah right, i remember at some point discussing about this as a potential possibility.

chaitanyaprem (approved):
LGTM

PR #1171, opened 2024-07-29
refactor: move rate limiter and priority queue from status-go to api package

richard-ramos:


waku/v2/api/publish/message_queue.go
```diff
@@ -0,0 +1,158 @@
...
+	item.index = n
+	*pq = append(*pq, item)
+}
+
+func (pq *envelopePriorityQueue) Pop() any {
+	old := *pq
+	n := len(old)
+	item := old[n-1]
+	old[n-1] = nil  // avoid memory leak
+	item.index = -1 // for safety
```
kaichaosun, line 59: I don't see the usage of index, should it be removed?

waku/v2/api/publish/message_queue.go
```diff
@@ -0,0 +1,158 @@
...
+
+// MessagePriority determines the ordering for the message priority queue
+type MessagePriority = int
+
+const (
+	LowPriority    MessagePriority = 1
+	MediumPriority MessagePriority = 2
+	HighPriority   MessagePriority = 3
+
+	DefaultPriority MessagePriority = MediumPriority
```
kaichaosun, line 18: Make LowPriority as the default may give more flexibility, as most of messages is default and low (maybe renamed it to NormalPriority and remove the const DefaultPriority). 
A few will be MediumPriority, and only occasionally there is HighPriority messages.

waku/v2/api/publish/message_queue_test.go
```diff
@@ -0,0 +1,91 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+
+	queue := NewMessageQueue(10, true)
+	go queue.Start(ctx)
+
+	queue.Push(protocol.NewEnvelope(&pb.WakuMessage{Timestamp: proto.Int64(0)}, 0, "A"), LowPriority)
+	queue.Push(protocol.NewEnvelope(&pb.WakuMessage{Timestamp: proto.Int64(1)}, 0, "B"), LowPriority)
+	queue.Push(protocol.NewEnvelope(&pb.WakuMessage{Timestamp: proto.Int64(2)}, 0, "C"), HighPriority)
+	queue.Push(protocol.NewEnvelope(&pb.WakuMessage{Timestamp: proto.Int64(3)}, 0, "D"), MediumPriority)
+	queue.Push(protocol.NewEnvelope(&pb.WakuMessage{Timestamp: proto.Int64(4)}, 0, "E"), HighPriority)
```
kaichaosun, line 52: This test can not different if it's because of fifo or timestamp order, right?

waku/v2/api/publish/message_queue.go
richard-ramos: It's used by `func (pq envelopePriorityQueue) Swap(i, j int)`

waku/v2/api/publish/message_queue.go
richard-ramos: I don't think having all the messages `LowPriority` by default is a good idea. Having the messages being by default a medium priority will give you more flexibility as some messages you probably don't care much for them to be sent immediately, while others you want them to be delivered ASAP. Having all the messages by default being Low would mean that you cannot deprioritize the messages any further.

However, renaming `MediumPriority` to `NormalPriority` does in fact sound like a good idea!

waku/v2/api/publish/message_queue_test.go
richard-ramos: Yah, it's deterministic

waku/v2/api/publish/message_queue.go
kaichaosun: ah, missed that one!

kaichaosun (approved):
👍

PR #1173, opened 2024-07-30
fix: replace references to old statusim.net domain

jakubgs:
Use of `statusim.net` domain been deprecated since March: https://github.com/status-im/infra-shards/commit/7df38c14


richard-ramos (approved):

chaitanyaprem (approved):
Thanks! 
LGTM

PR #1180, opened 2024-08-05
chore: move outgoing message check from status-go to go-waku

kaichaosun:
# Description

Move outgoing message check logic from status-go to go-waku

### DIFF
--- a/waku/v2/api/publish/message_check.go
+++ b/waku/v2/api/publish/message_check.go
@@ -25,8 +25,8 @@ type MessageSentCheck struct {
 	messageIDs         map[string]map[common.Hash]uint32
 	messageIDsMu       sync.RWMutex
 	storePeerID        peer.ID
-	messageStoredChan  chan common.Hash
-	messageExpiredChan chan common.Hash
+	MessageStoredChan  chan common.Hash
+	MessageExpiredChan chan common.Hash
 	ctx                context.Context
 	store              *store.WakuStore
 	timesource         timesource.Timesource
@@ -37,8 +37,8 @@ func NewMessageSentCheck(ctx context.Context, store *store.WakuStore, timesource
 	return &MessageSentCheck{
 		messageIDs:         make(map[string]map[common.Hash]uint32),
 		messageIDsMu:       sync.RWMutex{},
-		messageStoredChan:  make(chan common.Hash, 1000),
-		messageExpiredChan: make(chan common.Hash, 1000),
+		MessageStoredChan:  make(chan common.Hash, 1000),
+		MessageExpiredChan: make(chan common.Hash, 1000),
 		ctx:                ctx,
 		store:              store,
 		timesource:         timesource,
@@ -182,12 +182,12 @@ func (m *MessageSentCheck) messageHashBasedQuery(ctx context.Context, hashes []c
 
 		if found {
 			ackHashes = append(ackHashes, hash)
-			m.messageStoredChan <- hash
+			m.MessageStoredChan <- hash
 		}
 
 		if !found && uint32(m.timesource.Now().Unix()) > relayTime[i]+messageExpiredPerid {
 			missedHashes = append(missedHashes, hash)
-			m.messageExpiredChan <- hash
+			m.MessageExpiredChan <- hash
 		}
 	}
 
--- a/waku/v2/api/publish/message_check.go
+++ b/waku/v2/api/publish/message_check.go
@@ -1,0 +1,198 @@
+package publish
+
+import (
+	"bytes"
+	"context"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/waku-org/go-waku/waku/v2/protocol"
+	"github.com/waku-org/go-waku/waku/v2/protocol/pb"
+	"github.com/waku-org/go-waku/waku/v2/protocol/store"
+	"github.com/waku-org/go-waku/waku/v2/timesource"
+	"go.uber.org/zap"
+)
+
+const maxHashQueryLength = 100
+const hashQueryInterval = 3 * time.Second
+const messageSentPeriod = 3    // in seconds
+const messageExpiredPerid = 10 // in seconds
+
+type MessageSentCheck struct {
+	messageIDs         map[string]map[common.Hash]uint32
+	messageIDsMu       sync.RWMutex
+	storePeerID        peer.ID
+	messageStoredChan  chan common.Hash
+	messageExpiredChan chan common.Hash
+	ctx                context.Context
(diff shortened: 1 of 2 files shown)


waku/v2/api/publish/message_check.go
```diff
@@ -0,0 +1,198 @@
...
+		MessageStoredChan:  make(chan common.Hash, 1000),
+		MessageExpiredChan: make(chan common.Hash, 1000),
+		ctx:                ctx,
+		store:              store,
+		timesource:         timesource,
+		logger:             logger,
+	}
+}
+
+func (m *MessageSentCheck) Add(topic string, messageID common.Hash, sentTime uint32) {
```
chaitanyaprem, line 49: Does this have to be invoked by user? 
Maybe we should invoke this for relay and lightpush if this feature is enabled? 
Wondering if this can be done at api layer for relay and lightpush i.e have a send wrapper which would internally decide which protocol to use and at the same time add to this queue if this feature is enabled.

WDYT @richard-ramos

waku/v2/api/publish/message_check.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/waku-org/go-waku/waku/v2/timesource"
+	"go.uber.org/zap"
+)
+
+const maxHashQueryLength = 100
+const hashQueryInterval = 3 * time.Second
+const messageSentPeriod = 3    // in seconds
+const messageExpiredPerid = 10 // in seconds
+
+type MessageSentCheck struct {
```
chaitanyaprem, line 24: can we rename to something like SentMsgConfirmer or something in that line?

waku/v2/api/publish/message_check.go
```diff
@@ -0,0 +1,198 @@
...
+	m.messageIDsMu.Lock()
+	defer m.messageIDsMu.Unlock()
+
+	if _, ok := m.messageIDs[topic]; !ok {
+		m.messageIDs[topic] = make(map[common.Hash]uint32)
+	}
+	m.messageIDs[topic][messageID] = sentTime
+}
+
+func (m *MessageSentCheck) DeleteByMessageIDs(messageIDs []common.Hash) {
```
chaitanyaprem, line 59: When will this be used?

waku/v2/api/publish/message_check.go
```diff
@@ -0,0 +1,198 @@
...
+			m.logger.Debug("messages for next store hash query", zap.Any("messageIds", m.messageIDs))
+			m.messageIDsMu.Unlock()
+
+		}
+	}
+}
+
+func (m *MessageSentCheck) messageHashBasedQuery(ctx context.Context, hashes []common.Hash, relayTime []uint32, pubsubTopic string) []common.Hash {
+	selectedPeer := m.storePeerID
+	if selectedPeer == "" {
```
chaitanyaprem, line 145: Why not rely on peer selection in this case?  Worst case if there are no peers, failure will come from peer selection itself. 
This way static sharding becomes easier.

waku/v2/api/publish/message_check.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/waku-org/go-waku/waku/v2/protocol/pb"
+	"github.com/waku-org/go-waku/waku/v2/protocol/store"
+	"github.com/waku-org/go-waku/waku/v2/timesource"
+	"go.uber.org/zap"
+)
+
+const maxHashQueryLength = 100
+const hashQueryInterval = 3 * time.Second
+const messageSentPeriod = 3    // in seconds
+const messageExpiredPerid = 10 // in seconds
```
chaitanyaprem, line 22: Can we put all these into a config object the user can pass? 
We can leave the defaults to these values though in case nothing is passed while creating this object.

waku/v2/api/publish/message_check.go
richard-ramos: It does not sound like a bad idea at all, specially since I think the parameters for this `Add` function could be obtained from the envelope

waku/v2/api/publish/message_check.go
richard-ramos: This sounds like a candidate for an Options pattern

richard-ramos (approved):
Overall the PR LGTM, but since this is an API, it probably makes sense to add comments in the public functions about their purpose/usage

waku/v2/api/publish/message_check.go
kaichaosun: Peer selection is not used because selected store not working properly in the past, the situation may have changed. 
We can make such changes later.

waku/v2/api/publish/message_check.go
kaichaosun: When MVDS confirmation is received, it will call this method to removing the messageID.

waku/v2/api/publish/message_check.go
kaichaosun: Is it a convention to use something like `confirmer` or `checker`?
I find the name `MessageSentCheck` not worse than `SentMsgConfirmer` to express its logic actually.

waku/v2/api/publish/message_check.go
kaichaosun: Yeah, sounds great!
We can add a sent method to replace lightpush.publish and relay.publish with the added message sent check.
I will try such refactor in new PRs.

waku/v2/api/publish/message_check.go
chaitanyaprem: > This sounds like a candidate for an Options pattern

Yea, options pattern seems like a better choice.

waku/v2/api/publish/message_check.go
chaitanyaprem: Can we have an issue tracking so that we don't miss this?

waku/v2/api/publish/message_check.go
chaitanyaprem: Nothing like that, just that it is a routine performing periodic task of checking/confirming whether messages are sent. In general i have seen such terms being used when routines are involved. 
No hard opinion, just to make it more reader friendly.

chaitanyaprem (approved):
LGTM once comments are addressed.

waku/v2/api/publish/message_check.go
kaichaosun: via https://github.com/waku-org/go-waku/issues/1181

kaichaosun:
It seems some un-related test failed, wondering how did you manage to merge the PRs? @chaitanyaprem @richard-ramos

chaitanyaprem:
> It seems some un-related test failed, wondering how did you manage to merge the PRs? @chaitanyaprem @richard-ramos

I used my admin role to override branch protection 😉
But we ought to fix the root-cause which i don't know.

chaitanyaprem:
@kaichaosun , issue has been fixed in master thanks to @richard-ramos 

If you rebase your branch, CI should pass.

PR #1183, opened 2024-08-06
fix: keep channels open

richard-ramos:
@igor-sirotin reported the following error:

panic: send on closed channel
```
goroutine 228896 [running]:
github.com/waku-org/go-waku/waku/v2/api/publish.(*MessageQueue).Push(...)
        /Users/igorsirotin/Repositories/Status/status-go/vendor/github.com/waku-org/go-waku/waku/v2/api/publish/message_queue.go:131
github.com/status-im/status-go/wakuv2.(*Waku).Send(0x14002e9cf00, {0x14000bb8400?, 0xa?}, 0x1400347a000?, 0x107a22d50)
        /Users/igorsirotin/Repositories/Status/status-go/wakuv2/message_publishing.go:56 +0x140
github.com/status-im/status-go/wakuv2.(*PublicWakuAPI).Post(0x14006e4a3f0, {0x14001258f00?, 0x84?}, {{0x14001259100, 0x40}, {0x0, 0x0, 0x0}, {0x14001258f00, 0x40}, ...})
        /Users/igorsirotin/Repositories/Status/status-go/wakuv2/api.go:259 +0x568
```

Since this is happening because the channel is closed when the context is done, and the publish function is still being executed, attempting to add a message to the queue. To avoid this error, i just remove the need to close this channel, since it will be garbage collected eventually, since it will determine the channel to be unreachable will have its resources reclaimed whether or not it is closed, and also made Push and Pop functions cancelable by passing a context

Required to fix https://github.com/status-im/status-go/issues/5661

### DIFF
--- a/waku/v2/api/publish/message_queue.go
+++ b/waku/v2/api/publish/message_queue.go
@@ -110,7 +110,7 @@ func (m *MessageQueue) Start(ctx context.Context) {
 
 // Push an envelope into the message queue. The priority is optional, and will be ignored
 // if the message queue does not use a priority queue
-func (m *MessageQueue) Push(ctx context.Context, envelope *protocol.Envelope, priority ...MessagePriority) {
+func (m *MessageQueue) Push(ctx context.Context, envelope *protocol.Envelope, priority ...MessagePriority) error {
 	if m.usePriorityQueue {
 		msgPriority := NormalPriority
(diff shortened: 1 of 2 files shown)


igor-sirotin (approved):
Thank you!

Issue #1181, opened 2024-08-06
feat: send API for abstracting lightpush and relay publish with sent status check logic


kaichaosun:
Reference https://github.com/waku-org/go-waku/pull/1180#discussion_r1703906433

PR #1187, opened 2024-08-08
chore: refactor sender api

kaichaosun:
# Description

Fix https://github.com/waku-org/go-waku/issues/1181

API usage in status-go: https://github.com/status-im/status-go/pull/5676

### DIFF
--- a/waku/v2/api/publish/message_sender.go
+++ b/waku/v2/api/publish/message_sender.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/libp2p/go-libp2p/core/peer"
 	"github.com/waku-org/go-waku/waku/v2/protocol"
 	"github.com/waku-org/go-waku/waku/v2/protocol/lightpush"
 	"github.com/waku-org/go-waku/waku/v2/protocol/relay"
@@ -102,4 +103,16 @@ func (ms *MessageSender) Send(env *protocol.Envelope) error {
 
 func (ms *MessageSender) PublishMethod() PublishMethod {
 	return ms.publishMethod
+}
+
+func (ms *MessageSender) MessagesDelivered(messageIDs []common.Hash) {
+	if ms.messageSentCheck != nil {
+		ms.messageSentCheck.DeleteByMessageIDs(messageIDs)
+	}
+}
+
+func (ms *MessageSender) SetStorePeerID(peerID peer.ID) {
+	if ms.messageSentCheck != nil {
+		ms.messageSentCheck.SetStorePeerID(peerID)
+	}
 }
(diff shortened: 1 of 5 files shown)


waku/v2/api/publish/rate_limiting.go
```diff
@@ -35,3 +35,13 @@ func (p *PublishRateLimiter) ThrottlePublishFn(ctx context.Context, publishFn Pu
 		return publishFn(envelope, logger)
 	}
 }
+
+func (p *PublishRateLimiter) Check(ctx context.Context, logger *zap.Logger) error {
```
richard-ramos, line 39: We can probably reuse this function too in line 29 to avoid code duplication!

waku/v2/api/publish/message_sender.go
```diff
@@ -0,0 +1,124 @@
...
+			return errors.New("relay is not available")
+		}
+		peerCnt := len(ms.relay.PubSub().ListPeers(env.PubsubTopic()))
+		logger.Info("publishing message via relay", zap.Int("peerCnt", peerCnt))
+		_, err := ms.relay.Publish(ms.ctx, env.Message(), relay.WithPubSubTopic(env.PubsubTopic()))
+		return err
+	}
+
+	ephemeral := env.Message().Ephemeral
+	if ms.messageSentCheck != nil && (ephemeral == nil || !*ephemeral) {
```
richard-ramos, line 97: I think you can use this to avoid that nil pointer check
```suggestion
	if ms.messageSentCheck != nil && !env.Message().GetEphemeral() {
```

waku/v2/api/publish/message_sender.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	return nil
+}
+
+func (ms *MessageSender) Start() {
+	if ms.messageSentCheck != nil {
+		go ms.messageSentCheck.Start()
+	}
+}
```
richard-ramos, line 108: I'm thinking the context should be received here instead on the constructor.

waku/v2/api/publish/message_sender.go
```diff
@@ -0,0 +1,124 @@
...
+	ms.messageSentCheck = messageSentCheck
+	return ms
+}
+
+func (ms *MessageSender) WithRateLimiting(rateLimiter *PublishRateLimiter) *MessageSender {
+	ms.rateLimiter = rateLimiter
+	return ms
+}
+
+func (ms *MessageSender) Send(env *protocol.Envelope) error {
```
richard-ramos, line 70: What do you think of instead using the context from the struct, to use instead a context passed as a parameter in this function? that way you can make the Send function cancellable.

richard-ramos (approved):
LGTM

chaitanyaprem (approved):
LGTM

waku/v2/api/publish/message_sender.go
kaichaosun: Thanks for the tips!

kaichaosun:
Fixed the comments, also add an interface ISentCheck for MessageSentCheck, we can probably add interfaces to relay and lightpush for easily mock in tests in future. How do you think about it? @richard-ramos @chaitanyaprem

PR #1190, opened 2024-08-09
chore: limit the maximum number of message hashes to request per query

richard-ramos:
Otherwise it can request too many records and add too much load to the storenodes

### DIFF
--- a/waku/v2/api/missing/missing_messages.go
+++ b/waku/v2/api/missing/missing_messages.go
@@ -20,6 +20,7 @@ import (
 )
 
 const maxContentTopicsPerRequest = 10
+const maxMsgHashesPerRequest = 50
 
 // MessageTracker should keep track of messages it has seen before and
 // provide a way to determine whether a message exists or not. This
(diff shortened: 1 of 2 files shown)


waku/v2/api/missing/missing_messages.go
```diff
@@ -247,38 +248,55 @@ func (m *MissingMessageVerifier) fetchMessagesBatch(c chan<- *protocol.Envelope,
...
-		if err != nil {
-			if !errors.Is(err, context.Canceled) {
-				logger.Error("storenode not available", zap.Error(err))
+
+			for !result.IsComplete() {
+				for _, mkv := range result.Messages() {
+					select {
+					case c <- protocol.NewEnvelope(mkv.Message, mkv.Message.GetTimestamp(), mkv.GetPubsubTopic()):
+					default:
+						m.logger.Warn("subscriber is too slow!")
```
chaitanyaprem, line 278: does this mean these messages will not be fetched again?
wondering if that is ok.

waku/v2/api/missing/missing_messages.go
```diff
@@ -247,38 +248,55 @@ func (m *MissingMessageVerifier) fetchMessagesBatch(c chan<- *protocol.Envelope,
...
+				}
+				return
 			}
-			return result, nil
-		}, logger.With(zap.String("cursor", hex.EncodeToString(result.Cursor()))), "retrieving next page")
-		if err != nil {
-			if !errors.Is(err, context.Canceled) {
-				logger.Error("storenode not available", zap.Error(err))
+
+			for !result.IsComplete() {
```
chaitanyaprem, line 273: wondering if this is correct, what would happen to last result returned with let's say a few messages? 
Will that mark as result as complete as well or we will still do one more query to store node which will then return no more data but mark result as completed?

chaitanyaprem (approved):
LGTM, few comments.

waku/v2/api/missing/missing_messages.go
richard-ramos: It determines if it's complete based on whether the store query returned a cursor or not. If it didn't then the result is considered the last page

waku/v2/api/missing/missing_messages.go
richard-ramos: Not sure either. 
Do you think i should change this to this version instead? I originally tried to copy what pubsub does
```
select {
case c <- protocol.NewEnvelope(.................
case <- ctx.Done():
    return
}
```

kaichaosun (approved):

waku/v2/api/missing/missing_messages.go
chaitanyaprem: Nah, let's leave it as is...anyways if this condition is hit..it means app is not able to process normal messages as well. But not sure if we should stop doing queries if this condition is hit too often in a short time. 
Or maybe let's leave it and not complicate it altogether.

PR #1192, opened 2024-08-13
chore: use utc time in logs to avoid user location getting disclosed

### LINKED ISSUE #1191: bug: go-waku logs zone-sensitive time
# Problem 

Status user reported there're some logs that use local timezone, which exposes user's location. 
I fixed a few in status-go, but there're some in `go-waku` as well:

https://github.com/waku-org/go-waku/blob/19a47a1ac1f58de30cc3d987850f39633fc51711/waku/v2/peermanager/peer_connector.go#L210 https://github.com/waku-org/go-waku/blob/19a47a1ac1f58de30cc3d987850f39633fc51711/waku/v2/peermanager/peer_connector.go#L214 https://github.com/waku-org/go-waku/blob/19a47a1ac1f58de30cc3d987850f39633fc51711/waku/v2/peermanager/peer_connector.go#L231

# Solution

When logging, we should always convert time to UTC.

chaitanyaprem:
# Description
fixes #1191 

# Changes

- [x] changed default source to return time in UTC . @richard-ramos i have noticed NTP is not being used in status, but if NTP is used would the timesource provide time in UTC or local tz?
- [x] changed other places where time is logged to use UTC

### DIFF
--- a/waku/v2/peermanager/peer_connector.go
+++ b/waku/v2/peermanager/peer_connector.go
@@ -207,11 +207,11 @@ func (c *PeerConnectionStrategy) canDialPeer(pi peer.AddrInfo) bool {
 		now := time.Now()
 		if now.Before(tv.nextTry) {
 			c.logger.Debug("Skipping connecting to peer due to backoff strategy",
-				zap.Time("currentTime", now), zap.Time("until", tv.nextTry))
+				logging.UTCTime("currentTime", now), logging.UTCTime("until", tv.nextTry))
 			return false
 		}
 		c.logger.Debug("Proceeding with connecting to peer",
-			zap.Time("currentTime", now), zap.Time("nextTry", tv.nextTry))
+			logging.UTCTime("currentTime", now), logging.UTCTime("nextTry", tv.nextTry))
 	}
 	return true
 }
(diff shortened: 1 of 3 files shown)


waku/v2/peermanager/peer_connector.go
```diff
@@ -204,7 +204,7 @@ func (c *PeerConnectionStrategy) canDialPeer(pi peer.AddrInfo) bool {
 	val, ok := c.cache.Get(pi.ID)
 	if ok {
 		tv := val.(*connCacheData)
-		now := time.Now()
+		now := time.Now().UTC()
```
kaichaosun, line 207: It would be handy if it's wrapped in a utility function.

kaichaosun (approved):

waku/v2/peermanager/peer_connector.go
chaitanyaprem: yup, thought of the same. 
let me see which approach is the best.

richard-ramos:
> if NTP is used would the timesource provide time in UTC or local tz?
You still get the local time

richard-ramos (approved):

PR #1193, opened 2024-08-14
feat: shard aware pruning of peer store

### LINKED ISSUE #925: feat: shard aware peer store pruning
### Problem
With sharded peer management pruning of peers should consider shards the node is supporting and number of peers connected per shard. 

Refer to https://github.com/waku-org/nwaku/pull/2167 and once it is validated in nwaku with simulations, implement the same in go-waku.

### Acceptance criteria

- [ ] verify sharded peer store pruning by running same sim as nwaku

chaitanyaprem:
# Description
Peer store is not being pruned in go-waku rather new peer addition was being blocked once capacity is hit. This would cause issues once community sharding would be implemented. 

Replicating changes made in nwaku https://github.com/waku-org/nwaku/pull/2167 to address this. Since algorithm is tested in nwaku, not adding any specific tests.

Please do thorough review 🙏

# Changes

- [x] Periodic loop to prune peer store
- [x] pubsubTopic aware peer store pruning
- [x] run peerstore pruning for lightclients

closes #925


waku/v2/peermanager/peer_manager.go
```diff
@@ -241,6 +243,98 @@ func (pm *PeerManager) Start(ctx context.Context) {
...
+
+	notConnectedPeers := pm.getPeersBasedOnconnectionStatus("", network.NotConnected)
+	peersByTopic := make(map[string]peer.IDSlice)
+
+	if pm.RelayEnabled {
+		//prune not connected peers without shard
+		for _, peerID := range notConnectedPeers {
+			topics, err := pm.host.Peerstore().(wps.WakuPeerstore).PubSubTopics(peerID)
+			//Prune peers without pubsubtopics.
+			if err != nil || len(topics) == 0 {
```
richard-ramos, line 296: Should we log the error? Looking at the implementation of `PubsubTopics` in go-waku and `Get` in go-libp2p, seems to me that getting an error should be rare, so perhaps we might want to log it?

waku/v2/peermanager/peer_manager.go
```diff
@@ -241,6 +243,98 @@ func (pm *PeerManager) Start(ctx context.Context) {
...
+		if numPeers < pm.maxPeers {
+			pm.logger.Debug("finished pruning peer store", zap.Int("capacity", pm.maxPeers), zap.Int("beforeNumPeers", peerCntBeforePruning), zap.Int("afterNumPeers", numPeers))
+			return
+		}
+	}
+
+	notConnectedPeers := pm.getPeersBasedOnconnectionStatus("", network.NotConnected)
+	peersByTopic := make(map[string]peer.IDSlice)
+
+	if pm.RelayEnabled {
```
richard-ramos, line 291: Peers that are discovered via dnsdisc and peerExchange also populate the peerstore. Should we prune those as well? I just wonder if a light client wouldnt end up with an infinitely growing peerstore

waku/v2/peermanager/peer_manager.go
chaitanyaprem: yea, error should only happen if fetch failed which is very rare.
nonetheless better to log it since it is anyways rare.

waku/v2/peermanager/peer_manager.go
chaitanyaprem: good point, i added this check in the end. 
because i don't think we include pubsubTopics while adding peers in lightclient. 
will confirm this once and if we are indeed specifying pubsubTopics, then this check can be removed.

discovered peers will get pruned if they are not connected to.
let me review this from lightclient perspective once again...will try to run this using status-desktop and see what is the behaviour

waku/v2/peermanager/peer_manager.go
```diff
@@ -241,6 +243,98 @@ func (pm *PeerManager) Start(ctx context.Context) {
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+			pm.prunePeerStore()
+		}
+	}
+}
+
+func (pm *PeerManager) prunePeerStore() {
```
kaichaosun, line 262: I'm wondering if it's possible to do this,
- addPeer, check if meets maxPeers, if yes, return, otherwise add the peer.
- regular check and clear disconnected peers and meets the max retry.
- we may also need the max shards supported, each shard may have different maxPeers. 

This seems a bit easier to maintain.

chaitanyaprem: > addPeer, check if meets maxPeers, if yes, return, otherwise add the peer.

this was there before, but this is not proper as we don't know on what basis we are blocking new peer addition.

chaitanyaprem: > we may also need the max shards supported, each shard may have different maxPeers.

did not get this...why do we need max shards supported and maxPeers per shard? maxPeers indicates capacity in peerStore which should be for whole node. can you elaborate on any specific benefits you see with this approach.

kaichaosun: I see there is average calculation in your code, and this maxPeerPerShard may improve it and not bring too much burden to the whole node, also benefits scenarios like different shard needs different max peers for connection (this may not be true or needed).

chaitanyaprem: Hmm, maybe we can consider min peer per shard(but then again this may be hard in scenarios where a new shard is just coming up and only community owner's node is present)? what number should we consider for maxPeerPerShard ?

`scenarios like different shard needs different max peers for connection` 

why would this be a config? maybe not getting what you mean.

kaichaosun: For relay, the minPeerPerShard should be 4 to maintain network healthy, but lower can be accepted if really no available peers or poor connection situation. The maxPeerPerShard is more like a bandwidth consideration, could be around 6 by default, isn't this already set by libp2p pubsub protocol?

For other protocols, we may have different min and max peers for connection.

I may misunderstand the point of this PR, shouldn't peer store only clear dead connection?

maxShards and maxPeerPerShard (the sum of maxPeerPerProtocol in one shard) may provide an alternative way to maxPeers related logic. 

With more thinking, we don't need to configure different min/max peers for different shard until there is a strong requirement for this.

The current implementation LGTM, providing above feedback for discussion.

chaitanyaprem: `For relay, the minPeerPerShard should be 4 to maintain network healthy, but lower can be accepted if really no available peers or poor connection situation. The maxPeerPerShard is more like a bandwidth consideration, could be around 6 by default, isn't this already set by libp2p pubsub protocol`

i understand wrt minPeersPerShard, but maxPeerPerShard is where i assumed wrong as well. This is actual full-mesh peers and connected peers may not necessarily mean full-mesh peers. We need more connected peers in general so that gossip happens wrt messages in the network. otherwise it leads to unhealthy mesh. ref https://github.com/waku-org/go-waku/issues/1132


`For other protocols, we may have different min and max peers for connection.`

true, for light protocols this has to be determined but as of now i am thinking something like min of 2 per shard and max of 5-6. 

`maxShards and maxPeerPerShard (the sum of maxPeerPerProtocol in one shard) may provide an alternative way to maxPeers related logic.`

while implementing changes for lightClient i did realize that somewhere we need to link number of shards with maxPeerCount otherwise it may become tough. So, thanks for bringing this up. will look into it.

e.g: consider status communties where 2 common shards are present and a user is part of 20 communities which are on their own shards. in this case maxPeers should be `22*<per shard value>` .  Per shard value would be based on protocols mounted.

waku/v2/peermanager/peer_manager.go
chaitanyaprem: updated in new commit https://github.com/waku-org/go-waku/pull/1193/commits/d044e56111a2795ad217bddf7f497481466b436c

kaichaosun (approved):

waku/v2/peermanager/peer_manager.go
```diff
@@ -234,13 +237,112 @@ func (pm *PeerManager) SetPeerConnector(pc *PeerConnectionStrategy) {
...
+	avgPerTopic := min(1, total/maxPeerCnt)
+	// prune peers from shard with higher than avg count
+
+	for topic, peers := range peersByTopic {
+		count := max(len(peers)-avgPerTopic, 0)
+		for i, pID := range peers {
+			if i > count {
+				break
+			}
+			pm.logger.Debug("pruning peer as higher than average", zap.Stringer("peer", pID), zap.String("topic", topic))
```
richard-ramos, line 335: Could this log be too noisy? Wondering if it would be better to instead accumulate the peers that are going to be pruned per topic and do a single log line, i.e:
```
pm.logger.Debug("prining peers as higher than average", zap.Stringers("peers", peers), zap.String("topic", topic))
```
Same concern applies to line 302 and 306

richard-ramos (approved):
Good work!

Issue #1189, opened 2024-08-09
bug: Store request should timeout in a few seconds


kaichaosun:
**Describe the bug**
The store request with hash now takes a few minutes, waku api should return an error if reached timeout, the timeout should be a few seconds, how about 10s?

cc @richard-ramos @chaitanyaprem @plopezlpz

chaitanyaprem:
> **Describe the bug** The store request with hash now takes a few minutes, waku api should return an error if reached timeout, the timeout should be a few seconds, how about 10s?
> 
> cc @richard-ramos @chaitanyaprem @plopezlpz

sounds like a good idea, but i don't think this should be addressed in go-waku. The caller (status-go) can just create a context with timeout and pass it to store request which would then get handled automatically.

The context is already there so that caller will have a choice to decide timeout in such scenarios.

PR #1204, opened 2024-08-22
fix: set default store hash query timeout to 30s

kaichaosun:
# Description

Hash query may take too long, set the timeout to 30s to raise alert for any longer queries.

closes https://github.com/waku-org/go-waku/issues/1189


chaitanyaprem (approved):
LGTM

richard-ramos (approved):

PR #1206, opened 2024-08-23
fix: prevent panics in peermanager and WakuRelay

igor-sirotin:
# Description

Fixes https://github.com/status-im/status-go/issues/5759
More details in the issue comments.

It's possible that those functions are called with such parameters that causes panics.
While we should probably find the root cause and fix it, I'd like to prevent panics in the first place.


richard-ramos (approved):
Approving in the meantime but will open an issue so we can investigate better what's going on

Issue #1207, opened 2024-08-23
bug:


richard-ramos:
As described in https://github.com/status-im/status-go/issues/5759#issuecomment-2305221720 by @igor-sirotin 
The code as it is right now would allow publishing a message while not being subscribed to a pubsub topic.

I suggest either checking if a subscription to the pubsub topic before publishing and return an error or maybe attempt to subscribe with this code. I kinda prefer to return an error instead since then the user of the library can choose to subscribe or not depending on the error (and also because publishing without being subscribed does look like an incorrect usage of gowaku)

```
//Check if gossipsub subscription already exists for pubSubTopic
		if !w.IsSubscribed(pubSubTopic) {
			_, err := w.subscribeToPubsubTopic(cFilter.PubsubTopic)
			if err != nil {
				//TODO: Handle partial errors.
				w.log.Error("failed to subscribe to pubsubTopic", zap.Error(err), zap.String("pubsubTopic", cFilter.PubsubTopic))
				return nil, err
			}
		}
```

kaichaosun:
I found this piece of code which check similar logic, wondering why is it not working in this case?
https://github.com/waku-org/go-waku/blob/master/waku/v2/protocol/relay/waku_relay.go#L285

igor-sirotin:
> wondering why is it not working in this case?

I think it works 🙂 Check the logs I attached in https://github.com/status-im/status-go/issues/5759#issuecomment-2305818102 - `could not send message: not enough peers to publish`.

The question is, should we maybe auto-subscribe to topic in such case.

kaichaosun:
> should we maybe auto-subscribe to topic in such case.

It seems we need a wrapper API for auto subscribe. what do you think? @richard-ramos

chaitanyaprem:
> > should we maybe auto-subscribe to topic in such case.
> 
> It seems we need a wrapper API for auto subscribe. what do you think? @richard-ramos

it may not be so straight-forward, because when you subscribe to a pubsubtopic doesn't mean you have peers connected wrt topic. So, doing it as part of publish may not solve the issue rather mislead further. 

either we subscribe and wait to have atleast 1 peer for topic(which itself is not gauranteed msg will be reliably propagated).
or return an error so that caller can subscribe to topic and wait for it to be healthy before publishing.
wondering if we ca use topic-health to decide whether to publish or not rather than min-peers. wdyt @richard-ramos

richard-ramos:
[...]

PR #1212, opened 2024-08-28
fix: check subscription when relay publish message

kaichaosun:
# Description

Relates to: https://github.com/waku-org/go-waku/issues/1207

# Changes

- [ ] ...
- [ ] ...

# Tests

### DIFF
--- a/waku/v2/protocol/lightpush/waku_lightpush_test.go
+++ b/waku/v2/protocol/lightpush/waku_lightpush_test.go
@@ -336,7 +336,7 @@ func TestWakuLightPushCornerCases(t *testing.T) {
 
 	// Test corner case with default pubSub topic
 	_, err = client.Publish(ctx, msg2, WithDefaultPubsubTopic(), WithPeer(host2.ID()))
-	require.NoError(t, err)
+	require.Error(t, err)
 
 	// Test situation when cancel func is nil
 	lightPushNode2.cancel = nil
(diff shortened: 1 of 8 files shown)


waku/v2/protocol/lightpush/waku_lightpush_test.go
```diff
@@ -336,7 +336,7 @@ func TestWakuLightPushCornerCases(t *testing.T) {
 
 	// Test corner case with default pubSub topic
 	_, err = client.Publish(ctx, msg2, WithDefaultPubsubTopic(), WithPeer(host2.ID()))
-	require.NoError(t, err)
+	require.Error(t, err)
```
kaichaosun, line 339: I'm wondering if this is the right fix for this test?

richard-ramos: I guess a way to make sure the test is correct could be to check if the error emitted is really a `errors.New("cannot publish to unsubscribed topic")`?

richard-ramos (approved):

waku/v2/protocol/relay/waku_relay.go
```diff
@@ -280,12 +280,20 @@ func (w *WakuRelay) Publish(ctx context.Context, message *pb.WakuMessage, opts .
 		if err != nil {
 			return pb.MessageHash{}, err
 		}
+		_, err = w.subscribeToPubsubTopic(params.pubsubTopic)
```
chaitanyaprem, line 283: why subscribe only in case pubsubtopic is empty?
shouldn't we return errro in this case as well simliar to line 294

kaichaosun: For relay REST API, it supports publish a message without pubsubTopic, this is a quick fix for it.
Not sure if there's a better way for this scenario.

PR #1215, opened 2024-09-04
fix: make the envelope priority queue safe for concurrent access

richard-ramos:
Required to fix: https://github.com/status-im/status-mobile/issues/21174


chaitanyaprem (approved):
LGTM

PR #1214, opened 2024-09-04
fix: use correct ticker for all peers ping

chaitanyaprem:
# Description
while going through code, noticed that randomPeersTicker was being over-ridden by allpeersTicker.

PR #1216, opened 2024-09-04
fix: use correct ticker for all peers ping

iurimatias:
https://github.com/waku-org/go-waku/pull/1214 for 2.30.x branch


chaitanyaprem (approved):
Good to merge

PR #1218, opened 2024-09-11
ci: use GIT_REF for building docker image when set

siddarthkay:
# Description
This PR updates the build stage which builds the Docker Image to use `GIT_REF` when its set.
This ensures that when we perform a parameterised build from Jenkins we actually get a docker image from the branch or commit we set in `GIT_REF`

### DIFF
--- a/ci/Jenkinsfile.docker
+++ b/ci/Jenkinsfile.docker
@@ -43,9 +43,10 @@ pipeline {
   stages {
     stage('Build') {
       steps { script {
+        def commitOrRef = params.GIT_REF ? params.GIT_REF.take(8) : GIT_COMMIT.take(8)
         image = docker.build(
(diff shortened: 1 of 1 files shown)


chaitanyaprem (approved):
LGTM, does this mean specific commit build was never working?

jakubgs (commented):
This is wrong. `GIT_REF` is used to specify what version should be built, `GIT_COMMIT` is the variable set by Jenkins based on what has been actually checked out:

>GIT_COMMIT - The commit hash being checked out.
https://ci.status.im/env-vars.html/

You should not use `GIT_REF` for at least two reasons:

- It can be anything: branch, tag, or a commit.
- It might not be set at all if build source version is hardcoded.

You will need to undo or revert this commit.

PR #1221, opened 2024-09-17
feat: ratelimit store queries and add options to Next

richard-ramos:
Part of the refactoring for the storenode cycle. This will rate limit the store queries to a max of 8 queries per second per peerID. (This is the value used in staging). It's possible to skip this rate limit via options. 

In addition to that while debugging an issue with the storenode message counter, i found that not being able to pass options to `Next` to be somewhat limiting, specially if you want to change the requestID to some deterministic value, hence i modified it to pass options.

### DIFF
--- a/waku/v2/protocol/store/client.go
+++ b/waku/v2/protocol/store/client.go
@@ -19,6 +19,7 @@ import (
 	"github.com/waku-org/go-waku/waku/v2/protocol/store/pb"
 	"github.com/waku-org/go-waku/waku/v2/timesource"
 	"go.uber.org/zap"
+	"golang.org/x/time/rate"
 	"google.golang.org/protobuf/proto"
 )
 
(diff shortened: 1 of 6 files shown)


waku/v2/protocol/store/options.go
```diff
@@ -115,6 +116,13 @@ func IncludeData(v bool) RequestOption {
 	}
 }
 
+func SkipRateLimit() RequestOption {
```
chaitanyaprem, line 119: might be a good idea to have an option to set rate-limit as well?
not sure how useful it will be, but i am wondering if in the fleet rate-limit is increased does it makes sense to have this as an option as well.

waku/v2/protocol/store/client.go
```diff
@@ -32,6 +33,8 @@ const MaxPageSize = 100
 // DefaultPageSize is the default number of waku messages per page
 const DefaultPageSize = 20
 
+const maxQueriesPerSecond = 8
```
chaitanyaprem, line 36: maybe add a comment indicating why default is set to 8

chaitanyaprem (approved):
LGTM, apart from few minor comments.

PR #1222, opened 2024-09-18
feat: emit an event in EventBus upon dial error

adklempner:
# Description

Upon error dialing in `PeerConnectionStrategy.dialPeer` or `WakuNode.connect`, emit a `DialError` event. This allows consumers of `go-waku` to subscribe to a channel which outputs the dial errors as they occur.

# Changes

- Create `DialError` interface in utils
- Add an event emitter for `DialError` events to `PeerConnectionStrategy` (created in `Start()`, closed in `Stop()`)
- Add a function `HandleDialError` to `PeerConnectionStrategy` that:
  - runs existing logic on dial failure (add backoff, add to connection failures in waku peer store)
  - emit a `DialError` event using the ID of the peer and the returned error message from `c.host.Connect(ctx, pi)`
- use above function in `PeerConnectionStrategy.dialPeer` and `WakuNode.connect` when an error is returned by `host.Connect`

# Tests

### DIFF
--- a/waku/v2/node/wakunode2.go
+++ b/waku/v2/node/wakunode2.go
@@ -751,10 +751,7 @@ func (w *WakuNode) DialPeerWithInfo(ctx context.Context, peerInfo peer.AddrInfo)
 
 func (w *WakuNode) connect(ctx context.Context, info peer.AddrInfo) error {
 	err := w.host.Connect(ctx, info)
-	if err != nil {
-		w.host.Peerstore().(wps.WakuPeerstore).AddConnFailure(info.ID)
-		return err
-	}
+	w.peerConnector.HandleDialError(err, info.ID)
 
 	for _, addr := range info.Addrs {
 		// TODO: this is a temporary fix
--- a/waku/v2/node/wakunode2.go
+++ b/waku/v2/node/wakunode2.go
@@ -751,7 +751,12 @@ func (w *WakuNode) DialPeerWithInfo(ctx context.Context, peerInfo peer.AddrInfo)
 
 func (w *WakuNode) connect(ctx context.Context, info peer.AddrInfo) error {
 	err := w.host.Connect(ctx, info)
-	w.peerConnector.HandleDialError(err, info.ID)
+	if err != nil {
+		if w.peermanager != nil {
+			w.peermanager.HandleDialError(err, info.ID)
+		}
+		return err
+	}
 
 	for _, addr := range info.Addrs {
 		// TODO: this is a temporary fix
--- a/waku/v2/peermanager/peer_connector.go
+++ b/waku/v2/peermanager/peer_connector.go
@@ -19,7 +19,9 @@ import (
 	"github.com/waku-org/go-waku/waku/v2/onlinechecker"
 	wps "github.com/waku-org/go-waku/waku/v2/peerstore"
 	"github.com/waku-org/go-waku/waku/v2/service"
+	"github.com/waku-org/go-waku/waku/v2/utils"
 
+	"github.com/libp2p/go-libp2p/core/event"
 	"go.uber.org/zap"
 
 	lru "github.com/hashicorp/golang-lru"
@@ -39,8 +41,9 @@ type PeerConnectionStrategy struct {
 	*service.CommonDiscoveryService
 	subscriptions []subscription
 
-	backoff backoff.BackoffFactory
-	logger  *zap.Logger
+	backoff      backoff.BackoffFactory
+	logger       *zap.Logger
+	evtDialError event.Emitter
 }
 
 type subscription struct {
@@ -156,6 +159,11 @@ func (c *PeerConnectionStrategy) SetHost(h host.Host) {
 // Start attempts to connect to the peers passed in by peerCh.
 // Will not connect to peers if they are within the backoff period.
 func (c *PeerConnectionStrategy) Start(ctx context.Context) error {
+	var err error
+	c.evtDialError, err = c.host.EventBus().Emitter(new(utils.DialError))
+	if err != nil {
+		return err
+	}
 	return c.CommonDiscoveryService.Start(ctx, c.start)
 
 }
@@ -171,7 +179,9 @@ func (c *PeerConnectionStrategy) start() error {
 
 // Stop terminates the peer-connector
 func (c *PeerConnectionStrategy) Stop() {
-	c.CommonDiscoveryService.Stop(func() {})
+	c.CommonDiscoveryService.Stop(func() {
+		c.evtDialError.Close()
+	})
 }
 
 func (c *PeerConnectionStrategy) isPaused() bool {
@@ -277,11 +287,19 @@ func (c *PeerConnectionStrategy) dialPeer(pi peer.AddrInfo, sem chan struct{}) {
 	ctx, cancel := context.WithTimeout(c.Context(), c.dialTimeout)
 	defer cancel()
 	err := c.host.Connect(ctx, pi)
-	if err != nil && !errors.Is(err, context.Canceled) {
-		c.addConnectionBackoff(pi.ID)
-		c.host.Peerstore().(wps.WakuPeerstore).AddConnFailure(pi.ID)
-		c.logger.Warn("connecting to peer", logging.HostID("peerID", pi.ID), zap.Error(err))
-	}
+	c.HandleDialError(err, pi.ID)
 	c.host.Peerstore().(wps.WakuPeerstore).ResetConnFailures(pi.ID)
 	<-sem
+}
+
+func (c *PeerConnectionStrategy) HandleDialError(err error, peerID peer.ID) {
+	if err != nil && !errors.Is(err, context.Canceled) {
+		c.addConnectionBackoff(peerID)
+		c.host.Peerstore().(wps.WakuPeerstore).AddConnFailure(peerID)
+		c.logger.Warn("connecting to peer", logging.HostID("peerID", peerID), zap.Error(err))
+		emitterErr := c.evtDialError.Emit(utils.DialError{Err: err, PeerID: peerID})
+		if emitterErr != nil {
+			c.logger.Error("failed to emit DialError", zap.Error(emitterErr))
+		}
+	}
 }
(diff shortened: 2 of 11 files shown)


waku/v2/peermanager/peer_connector.go
```diff
@@ -272,6 +272,11 @@ func (c *PeerConnectionStrategy) dialPeers() {
 	}
 }
 
+type DialError struct {
```
chaitanyaprem, line 275: Better to place this type in utils package so its accessible from other places.

This file mainly dials/connects to peers for relay and when the peers are added to peerstore. 
we need to capture on-demand dialing scenarios especially for req/resp protocols such as  lightpush/filter/store e.g https://github.com/waku-org/go-waku/blob/280039120431f7085a9bd5beb6d9804889cf3d3f/waku/v2/protocol/lightpush/waku_lightpush.go#L196

So, my recommendation would be to search the code for NewStream or Connect methods and look for dial errors and report them in the same way.

i am not too sure, but we may have to check in status-go code as well because some peers are dialed directly from there (which i think are mostly fleet peers). @richard-ramos can confirm.

waku/v2/peermanager/peer_connector.go
```diff
@@ -281,6 +286,15 @@ func (c *PeerConnectionStrategy) dialPeer(pi peer.AddrInfo, sem chan struct{}) {
 		c.addConnectionBackoff(pi.ID)
 		c.host.Peerstore().(wps.WakuPeerstore).AddConnFailure(pi.ID)
 		c.logger.Warn("connecting to peer", logging.HostID("peerID", pi.ID), zap.Error(err))
+		em, emitterErr := c.host.EventBus().Emitter(new(DialError))
```
richard-ramos, line 289: The emitter could be part of the `PeerConnectionStrategy` struct to reuse it. That way it would not be necessary to create a new emitter every time you have a dial error. (See example in  `CreateEventEmitters()` on `topic_events.go`)

adklempner: Refactored in latest commit. The emitter is now part of `PeerConnectionStrategy` struct, created in `Start()` and closed in `Stop()`

waku/v2/peermanager/peer_connector.go
adklempner: Refactored `DialError` into utils and added a public function to `PeerConnector` to emit the error. Called both in `PeerConnectionStrategy.dialPeer` and `WakuNode.connect`

waku/v2/peermanager/peer_connector.go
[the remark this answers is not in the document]
adklempner: refactored ALL the dial error handling logic into `HandleDialError`

waku/v2/peermanager/peer_connector.go
chaitanyaprem: > Connections are now done via `AddDiscoveredPeer`

that is when a peer is getting added. Now, consider the peer was discovered and added in lightclient but not used so the connection may get dropped due to being idle. And when the client wants to send a msg via lightpush a new peer or the same peer may get selected and new connection will be established via NewStream call. isn't it? unless i am missing something.

adklempner: I took a look at go-libp2p and it does seem to call `host.connect` in `NewStream`, at least here: https://github.com/libp2p/go-libp2p/blob/9038a72c4863778e6a26814aab201cee6d93d640/p2p/host/basic/basic_host.go#L689

Does it make sense to use this same logic for handling dial errors in every call to `NewStream` in go-waku? https://github.com/waku-org/go-waku/blob/4bba45617a48d3e3653a0fcce2641a65d2d5b79f/waku/v2/peermanager/peer_connector.go#L295

chaitanyaprem: i think so, because stream creation errors would mostly be due to underlying connectivity issues only unless there is a bug in stream management which i think is rare.

adklempner: I moved the function to `PeerManager` and added a call to it in every place where `NewStream` is used

waku/v2/node/wakunode2.go
```diff
@@ -751,10 +751,7 @@ func (w *WakuNode) DialPeerWithInfo(ctx context.Context, peerInfo peer.AddrInfo)
 
 func (w *WakuNode) connect(ctx context.Context, info peer.AddrInfo) error {
 	err := w.host.Connect(ctx, info)
-	if err != nil {
-		w.host.Peerstore().(wps.WakuPeerstore).AddConnFailure(info.ID)
-		return err
-	}
+	w.peermanager.HandleDialError(err, info.ID)
```
chaitanyaprem, line 754: maybe add a check for peermanager nil, afair PeerManager is optional to start WakuNode and can lead to panic.

waku/v2/peermanager/peer_connector.go
```diff
@@ -277,10 +276,9 @@ func (c *PeerConnectionStrategy) dialPeer(pi peer.AddrInfo, sem chan struct{}) {
 	ctx, cancel := context.WithTimeout(c.Context(), c.dialTimeout)
 	defer cancel()
 	err := c.host.Connect(ctx, pi)
-	if err != nil && !errors.Is(err, context.Canceled) {
-		c.addConnectionBackoff(pi.ID)
-		c.host.Peerstore().(wps.WakuPeerstore).AddConnFailure(pi.ID)
-		c.logger.Warn("connecting to peer", logging.HostID("peerID", pi.ID), zap.Error(err))
+	if err != nil {
+		c.pm.HandleDialError(err, pi.ID)
+		return
```
chaitanyaprem, line 281: returning from here means we should release semaphore which is done in line 284.

waku/v2/protocol/lightpush/waku_lightpush_test.go
```diff
@@ -264,6 +265,9 @@ func TestWakuLightPushCornerCases(t *testing.T) {
 
 	// Prepare peer manager instance to include in test
 	pm := peermanager.NewPeerManager(10, 10, nil, nil, true, utils.Logger())
+	pc, err := peermanager.NewPeerConnectionStrategy(pm, onlinechecker.NewDefaultOnlineChecker(true), 10*time.Second, utils.Logger())
```
chaitanyaprem, line 268: why is this required for a lightnode? 

peerManager alone should be sufficient right.

waku/v2/protocol/peer_exchange/client.go
```diff
@@ -76,9 +76,7 @@ func (wakuPX *WakuPeerExchange) Request(ctx context.Context, numPeers int, opts
 
 	stream, err := wakuPX.h.NewStream(ctx, params.selectedPeer, PeerExchangeID_v20alpha1)
 	if err != nil {
-		if ps, ok := wakuPX.h.Peerstore().(peerstore.WakuPeerstore); ok {
-			ps.AddConnFailure(params.selectedPeer)
-		}
+		wakuPX.pm.HandleDialError(err, params.selectedPeer)
```
chaitanyaprem, line 79: check for `pm != nil`

waku/v2/protocol/store/client.go
```diff
@@ -281,10 +281,7 @@ func (s *WakuStore) queryFrom(ctx context.Context, storeRequest *pb.StoreQueryRe
 
 	stream, err := s.h.NewStream(ctx, params.selectedPeer, StoreQueryID_v300)
 	if err != nil {
-		logger.Error("creating stream to peer", zap.Error(err))
-		if ps, ok := s.h.Peerstore().(peerstore.WakuPeerstore); ok {
-			ps.AddConnFailure(params.selectedPeer)
-		}
+		s.pm.HandleDialError(err, params.selectedPeer)
```
chaitanyaprem, line 284: check for `pm != nil`

waku/v2/node/wakunode2.go
adklempner: Fixed in last commit

waku/v2/peermanager/peer_connector.go
adklempner: I see, it seemed strange to record a dial error and then reset connection failures right after. I removed the return and use an if/else statement to only reset connection failures if there is no error

waku/v2/protocol/peer_exchange/client.go
adklempner: Fixed in last commit

waku/v2/protocol/store/client.go
adklempner: Fixed in last commit

waku/v2/peermanager/peer_manager.go
```diff
@@ -719,3 +729,24 @@ func (pm *PeerManager) addPeerToServiceSlot(proto protocol.ID, peerID peer.ID) {
...
+			pm.host.Peerstore().(wps.WakuPeerstore).AddConnFailure(peerID)
+		}
+		pm.logger.Warn("connecting to peer", logging.HostID("peerID", peerID), zap.Error(err))
+		if pm.evtDialError != nil {
+			emitterErr := pm.evtDialError.Emit(utils.DialError{Err: err, PeerID: peerID})
+			if emitterErr != nil {
+				pm.logger.Error("failed to emit DialError", zap.Error(emitterErr))
+			}
+		}
+	}
```
richard-ramos, line 751:
```suggestion
	if err == nil || errors.Is(err, context.Canceled {
		return
	}

	if pm.peerConnector != nil {
		pm.peerConnector.addConnectionBackoff(peerID)
	}
	if pm.host != nil {
		pm.host.Peerstore().(wps.WakuPeerstore).AddConnFailure(peerID)
	}
	pm.logger.Warn("connecting to peer", logging.HostID("peerID", peerID), zap.Error(err))
	if pm.evtDialError != nil {
		emitterErr := pm.evtDialError.Emit(utils.DialError{Err: err, PeerID: peerID})
		if emitterErr != nil {
			pm.logger.Error("failed to emit DialError", zap.Error(emitterErr))
		}
	}

```

richard-ramos (approved):
LGTM!

waku/v2/peermanager/peer_manager.go
adklempner: Fixed

PR #1238, opened 2024-10-03
fix: simple backoff strategy after 3 subscribe failures

chaitanyaprem:
# Description
Fixes https://github.com/status-im/status-go/issues/5679 where due to DNS failure,  filter subscribe loop is going into infinite loop leading to log flooding. 

- [x] Added a errorCount which is incremented in case of error that requires a backoff such as "no suitable peers found" . As of now the backoff is fixed and not exponential.

### DIFF
--- a/waku/v2/api/filter/filter.go
+++ b/waku/v2/api/filter/filter.go
@@ -3,10 +3,12 @@ package filter
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"time"
 
 	"github.com/google/uuid"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/libp2p/go-libp2p/p2p/net/swarm"
 	"github.com/waku-org/go-waku/waku/v2/onlinechecker"
 	"github.com/waku-org/go-waku/waku/v2/protocol"
 	"github.com/waku-org/go-waku/waku/v2/protocol/filter"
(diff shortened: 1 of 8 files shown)


waku/v2/api/filter/filter.go
```diff
@@ -181,6 +187,10 @@ func (apiSub *Sub) resubscribe(failedPeer peer.ID) {
 	apiSub.multiplex(subs)
 }
 
+func possibleRecursiveError(err error) bool {
+	return strings.Contains(err.Error(), utils.ErrNoPeersAvailable.Error()) || strings.Contains(err.Error(), "dial backoff")
```
richard-ramos, line 191: I wonder if this would work so it's not necessary to check for the error string:
```
return errors.Is(err, utils.ErrNoPeersAvailable) || errors.Is(err, swarm.ErrDialBackoff)
```

waku/v2/utils/peer.go
```diff
@@ -10,6 +12,10 @@ type DialError struct {
 	PeerID peer.ID
 }
 
+// ErrNoPeersAvailable is emitted when no suitable peers are found for
+// some protocol
+var ErrNoPeersAvailable = errors.New("no suitable peers found")
+
```
richard-ramos, line 18: I'm confused on why `ErrNoPeersAvailable` was moved inside `utils`. Isn't this error emitted/used only by the peer manager package?

richard-ramos (approved):

waku/v2/utils/peer.go
chaitanyaprem: i had to use this in `api/filter.go` line `191` , hence had to move it out.

waku/v2/api/filter/filter.go
chaitanyaprem: ah, thanks..that would definitely be better...will make the change.

PR #1244, opened 2024-10-18
fix:  use bad peer removal logic only for lightpush and filter and option to restart missing message verifier

chaitanyaprem:
# Description
Bunch of fixes and changes that help with offline state.

# Changes

- [x] https://github.com/waku-org/go-waku/pull/1241 had some side-effects where-in store nodes were being removed from peerstore which we do not want because in status store nodes are fixed (i.e fleet nodes).
- [x] Apart from that i had noticed that when node is offline, connections are being triggered which result in failures and many good peers such as fleet nodes are getting removed from peer-store.
- [x] Added a more strict check and now migrated the bad peer removal logic to only be done from lightpush and filter and that too based on specific errors to avoid such side-effects.
- [x] Found an issue with offline to online state change handling in filter-manager where filter ping to existing subs will never happen.
- [x] Added a check to filter ping error handling in case node is already offline so that sub are not cleared
- [x] ~Added functionality to specify preferred-peers for filter. One of the peers used for filter subscription out of n will be randomly chosen from list of preferred peers.~
- [x] Added a fix for missing messages to stop it when node goes offline in order to reduce backlog it accumulates and never being able to catch-up.

# Tests

Being dogfooded https://github.com/status-im/status-mobile/issues/21172 for various network disconnection/switch scenarios.

### DIFF
--- a/waku/v2/api/missing/missing_messages.go
+++ b/waku/v2/api/missing/missing_messages.go
@@ -35,6 +35,7 @@ type MessageTracker interface {
 // MissingMessageVerifier is used to periodically retrieve missing messages from store nodes that have some specific criteria
 type MissingMessageVerifier struct {
 	ctx    context.Context
+	cancel context.CancelFunc
 	params missingMessageVerifierParams
 
 	storenodeRequestor common.StorenodeRequestor
@@ -43,10 +44,12 @@ type MissingMessageVerifier struct {
 	criteriaInterest   map[string]criteriaInterest // Track message verification requests and when was the last time a pubsub topic was verified for missing messages
 	criteriaInterestMu sync.RWMutex
 
-	C <-chan *protocol.Envelope
+	C chan *protocol.Envelope
 
-	timesource timesource.Timesource
-	logger     *zap.Logger
+	timesource   timesource.Timesource
+	logger       *zap.Logger
+	isRunning    bool
+	runningMutex sync.RWMutex
 }
 
 // NewMissingMessageVerifier creates an instance of a MissingMessageVerifier
@@ -63,6 +66,8 @@ func NewMissingMessageVerifier(storenodeRequester common.StorenodeRequestor, mes
 		messageTracker:     messageTracker,
 		logger:             logger.Named("missing-msg-verifier"),
 		params:             params,
+		criteriaInterest:   make(map[string]criteriaInterest),
+		C:                  make(chan *protocol.Envelope, 1000),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


waku/v2/protocol/filter/client.go
```diff
@@ -351,11 +356,20 @@ func (wf *WakuFilterLightNode) handleFilterSubscribeOptions(ctx context.Context,
 		params.selectedPeers = append(params.selectedPeers, pData.AddrInfo.ID)
 	}
 	reqPeerCount := params.maxPeers - len(params.selectedPeers)
+	for _, p := range params.selectedPeers {
+		if params.peersToExclude == nil {
+			params.peersToExclude = make(peermanager.PeerSet)
+		}
+		//exclude peers that are preferredpeers so that they don't get selected again.
+		if _, ok := params.peersToExclude[p]; !ok {
```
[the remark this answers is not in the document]
chaitanyaprem: Ah, i see this is confusing. will change it to reduce confusion.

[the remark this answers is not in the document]
chaitanyaprem: Anyways we are not using the logic of preferredPeers in status-go, maybe i will remove this code itself...it is making it too complex.

waku/v2/protocol/filter/filter_health_check.go
```diff
@@ -24,7 +24,7 @@ func (wf *WakuFilterLightNode) PingPeer(peer peer.ID) {
 	ctxWithTimeout, cancel := context.WithTimeout(wf.CommonService.Context(), PingTimeout)
 	defer cancel()
 	err := wf.Ping(ctxWithTimeout, peer)
-	if err != nil {
+	if err != nil && wf.onlineChecker.IsOnline() {
```
richard-ramos, line 27: I wonder if we should instead do
```
if wf.onlineChecker.IsOnline() {
   err := wf.Ping(ctxWithTimeout, peer)
   if err != nil {
              ...............
   }
}
```
to not even attempt to ping if the node is offline

waku/v2/api/missing/missing_messages.go
```diff
@@ -123,24 +140,33 @@ func (m *MissingMessageVerifier) Start(ctx context.Context) {
...
 			}
 		}
 	}()
 }
 
+func (m *MissingMessageVerifier) Stop() {
+	m.cancel()
+	m.runningMutex.Lock()
+	defer m.runningMutex.Unlock()
+	m.isRunning = false
```
richard-ramos, line 167: Do you think we should recreate the channel `C` on `Stop()` to reset it?

richard-ramos (approved):

fryorcraken:
You crossed out

>  Added functionality to specify preferred-peers for filter. One of the peers used for filter subscription out of n will be randomly chosen from list of preferred peers.

But it is not reflected in PR title. Is that in or out? (hopefully out)

chaitanyaprem:
> But it is not reflected in PR title. Is that in or out? (hopefully out)

I did update the PR title to not include it :) Weird if you are still seeing it in the title .

Current title is `fix: use bad peer removal logic only for lightpush and filter and provide option to restart missing message verifier`

Could be some state update issue in github?

waku/v2/protocol/filter/filter_health_check.go
chaitanyaprem: that is already handled in https://github.com/waku-org/go-waku/blob/e0a9ae0447cecfe3e301f33a1c6b8509a2bb59b0/waku/v2/protocol/filter/filter_health_check.go#L52-L55.

While going through the logs i had noticed that many times ping goes through and then node goes offline and failures cause issue, hence i added this additional check while checking for ping result. It is not completely fool-proof but it does take care of additional scenarios.

I did not want to complicate the code further by tying up ping context to online status, hence left it at this.

waku/v2/api/missing/missing_messages.go
chaitanyaprem: Valid point, but the caller also needs to keep recreating the routine to listen to channel every time we recreate which seemed like it would complicate things. hence i left it as is. 

Any specific issue you see if we don't recreate the channel?

richard-ramos: No specific issue, just curious.
I just find that losing the channel after a stop seemed logical as it kinda notifies all subscribers to it that they're no longer going to receive messages, but it's totally okay to leave the channel alive.

PR #1245, opened 2024-10-20
chore: use decred secp256k1 directly

BabyHalimao:
Use `github.com/decred/dcrd/dcrec/secp256k1/v4` directly rather than `github.com/btcsuite/btcd/btcec/v2` which is just a wrapper around the underlying decred library. Inspired by ethereum/go-ethereum#30595

`github.com/btcsuite/btcd/btcec/v2` has a very annoying breaking change when upgrading from v2.3.3 to v2.3.4. The easiest way to work around this is to remove the wrapper.

### DIFF
--- a/waku/v2/utils/crypto.go
+++ b/waku/v2/utils/crypto.go
@@ -3,21 +3,21 @@ package utils
 import (
 	"crypto/ecdsa"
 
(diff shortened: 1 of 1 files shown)


richard-ramos (approved):
Thank you!!!

PR #1248, opened 2024-10-22
refactor: use protobuffer for API storenode queries

richard-ramos:
To allow more flexibility at the moment of doing queries, instead of having custom functions that receive N different combination of parameters to do a store query, we use instead a single function which receives instead a StoreQueryRequest protobuffer

The json representation of the StoreQueryRequest matches what is used in libwaku, and since protobuffers in nwaku and go-waku are the same (and are supposed to be backward compatible), we can safely use them. It has the side effect of making the API simpler too

cc: @gabrielmer, @Ivansete-status

### DIFF
--- a/waku/v2/protocol/store/client.go
+++ b/waku/v2/protocol/store/client.go
@@ -194,6 +194,35 @@ func (s *WakuStore) Request(ctx context.Context, criteria Criteria, opts ...Requ
 	return result, nil
 }
 
+func (s *WakuStore) RequestRaw(ctx context.Context, peerID peer.ID, storeRequest *pb.StoreQueryRequest) (Result, error) {
+	err := storeRequest.Validate()
+	if err != nil {
+		return nil, err
+	}
+
+	var params Parameters
+	params.selectedPeer = peerID
+	if params.selectedPeer == "" {
+		return nil, ErrMustSelectPeer
+	}
+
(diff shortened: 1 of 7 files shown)


waku/v2/protocol/store/client.go
```diff
@@ -194,6 +194,35 @@ func (s *WakuStore) Request(ctx context.Context, criteria Criteria, opts ...Requ
...
 
+func (s *WakuStore) RequestRaw(ctx context.Context, peerID peer.ID, storeRequest *pb.StoreQueryRequest) (Result, error) {
+	err := storeRequest.Validate()
+	if err != nil {
+		return nil, err
+	}
+
+	var params Parameters
+	params.selectedPeer = peerID
+	if params.selectedPeer == "" {
```
chaitanyaprem, line 205: I am guessing peer-selection would be added later.

chaitanyaprem (approved):
LGTM

waku/v2/protocol/store/client.go
richard-ramos: Yes. I want to get the store queries to work first as they do now, with nwaku

waku/v2/api/common/storenodeRequestor.go
```diff
@@ -0,0 +1,12 @@
+package common
```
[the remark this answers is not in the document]
richard-ramos: yup

PR #1249, opened 2024-10-22
refactor: extract ping interface

richard-ramos:
This would allow replacing the Ping function by a nim-libp2p implementation in status-go + nwaku.
A default implementation is also provided to be used in status-go + gowaku

cc: @gabrielmer  @Ivansete-status

### DIFF
--- a/waku/v2/api/common/pinger.go
+++ b/waku/v2/api/common/pinger.go
@@ -1,0 +1,37 @@
+package common
+
+import (
+	"context"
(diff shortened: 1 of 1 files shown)


chaitanyaprem (approved):
LGTM

wondering what is the need of ping in status-go if storenodeCycle is being moved to go-waku? ideally we should move any such logic into the API as we have been doing for a lot of waku related code.

PR #1252, opened 2024-10-29
fix: use bad peer removal logic only for lightpush and filter

chaitanyaprem:
# Description
https://github.com/waku-org/go-waku/pull/1241 had some side-effects where-in store nodes were being removed from peerstore which we do not want because in status store nodes are fixed (i.e fleet nodes).

- Apart from that i had noticed that when node is offline, connections are being triggered which result in failures and many good peers such as fleet nodes are getting removed from peer-store. 
- Added a more strict check and now migrated the bad peer removal logic to only be done from lightpush and filter and that too based on specific errors to avoid such side-effects. 
- Found an issue with offline to online state change handling in filter-manager where filter ping to existing subs will never happen. 
- Added a check to filter ping error handling in case node is already offline so that sub are not cleared

This still needs to be tested, will update the PR as i get more info from the testing.


chaitanyaprem:
Closing this as changes are being considered in https://github.com/waku-org/go-waku/pull/1244

PR #1256, opened 2024-11-20
fix: setup shards

richard-ramos:
# Description
go-waku `waku` command did not setup the list of shards on the ENR when relay was disabled. This ended up causing the node to not be able to connect due to the ENR not containing any shard. This change sets the list of shards regardless of relay being enabled or not

Fixes: https://github.com/waku-org/go-waku/issues/1255


chaitanyaprem (approved):
LGTM

PR #1258, opened 2024-11-25
feat: filter rate limit

richard-ramos:
Also includes refactor to have a rate limiter per peer with TTL
cc: @fryorcraken

### DIFF
--- a/waku/v2/protocol/filter/client.go
+++ b/waku/v2/protocol/filter/client.go
@@ -167,6 +167,9 @@ func (wf *WakuFilterLightNode) onRequest(ctx context.Context) func(network.Strea
 
 		if !wf.limiter.Allow(peerID) {
 			wf.metrics.RecordError(rateLimitFailure)
+			if err := stream.Reset(); err != nil {
+				wf.log.Error("resetting connection", zap.Error(err))
+			}
 			return
 		}
 
@@ -318,8 +321,8 @@ func (wf *WakuFilterLightNode) request(ctx context.Context, requestID []byte,
 	return nil
 }
 
-func (wf *WakuFilterLightNode) handleFilterSubscribeOptions(ctx context.Context, contentFilter protocol.ContentFilter, opts []SubscribeOption) (*SubscribeParameters, map[string][]string, error) {
-	params := new(SubscribeParameters)
+func (wf *WakuFilterLightNode) handleFilterSubscribeOptions(ctx context.Context, contentFilter protocol.ContentFilter, opts []FilterSubscribeOption) (*FilterSubscribeParameters, map[string][]string, error) {
+	params := new(FilterSubscribeParameters)
 	params.log = wf.log
 	params.host = wf.h
 	params.pm = wf.pm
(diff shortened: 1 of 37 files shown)


waku/v2/protocol/filter/options.go
```diff
@@ -53,19 +54,41 @@ type (
...
+	FilterFullNodeParameters struct {
 		Timeout        time.Duration
 		MaxSubscribers int
 		pm             *peermanager.PeerManager
+		limitR         rate.Limit
+		limitB         int
 	}
 
-	Option func(*FilterParameters)
+	FullNodeOption func(*FilterFullNodeParameters)
```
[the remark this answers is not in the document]
richard-ramos: I used FullNodeOption because both lightnode and fullnode have their own options, also, using `Filter` stutters when you use the package name due to repeating the word "filter", i.e. "filter.FilterFullNodeOption` vs `filter.FullNodeOption`.

Hm. I should change the parameters struct name too.

waku/v2/protocol/filter/client.go
```diff
@@ -287,7 +302,7 @@ func (wf *WakuFilterLightNode) request(ctx context.Context, requestID []byte,
 
 	}
 
-	if filterSubscribeResponse.RequestId != request.RequestId {
+	if filterSubscribeResponse.RequestId != "N/A" && filterSubscribeResponse.RequestId != request.RequestId {
```
chaitanyaprem, line 305: what is "N/A"? don't remember seeing it before.

also is this implemented in nwaku as well? otherwise this condition will fail if filter server is nwaku node.

waku/v2/protocol/filter/options.go
```diff
@@ -11,30 +11,31 @@ import (
...
 	}
 }
 
 type (
-	FilterPingParameters struct {
+	PingParameters struct {
 		requestID []byte
 	}
-	FilterPingOption func(*FilterPingParameters)
+	PingOption func(*PingParameters)
```
chaitanyaprem, line 28: same comment as before, if it is not required better not to change now.

waku/v2/protocol/filter/options.go
```diff
@@ -53,27 +54,49 @@ type (
...
+func WithLightNodeRateLimiter(r rate.Limit, b int) LightNodeOption {
+	return func(params *LightNodeParameters) {
+		params.limitR = r
+		params.limitB = b
+	}
+}
+
+func DefaultLightNodeOptions() []LightNodeOption {
+	return []LightNodeOption{
+		WithLightNodeRateLimiter(rate.Inf, 0),
```
chaitanyaprem, line 86: what would be these rate limits for status? can we set those as defaults

waku/v2/utils/limiter.go
```diff
@@ -0,0 +1,72 @@
...
+	r        rate.Limit
+	b        int
+}
+
+func NewRateLimiter(r rate.Limit, b int) *RateLimiter {
+	return &RateLimiter{
+		r: r,
+		b: b,
+		limiters: ttlcache.New[peer.ID, *rate.Limiter](
+			ttlcache.WithTTL[peer.ID, *rate.Limiter](30 * time.Minute),
```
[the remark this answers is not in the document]
richard-ramos: It shouldnt be a problem I think? because if for some reason you are connected to that peer, exceed the 30m and do a new request, we'd create a rate limiter from scratch, altho I imagine that it would be rare for a filter peer to remain connected for so long without doing any request? At least ping requests should happen, and ttlcache does extend the expiration date when you call a `Get`.

waku/v2/protocol/filter/options.go
richard-ramos: In status-go it's set as 1 per second.

chaitanyaprem: ok, i was asking more from the point of the new per peer based rate limiter. Is that also set to 1 per sec?

waku/v2/utils/limiter.go
```diff
@@ -0,0 +1,72 @@
...
+		r.limiters.Set(peerID, limiter, ttlcache.DefaultTTL)
+	} else {
+		v := r.limiters.Get(peerID)
+		limiter = v.Value()
+	}
+	return limiter
+}
+
+func (r *RateLimiter) Allow(peerID peer.ID) bool {
+	return r.getOrCreate(peerID).Allow()
```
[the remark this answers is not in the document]
chaitanyaprem: ah, ok....as long as it is reference based it is fine.
Thanks for clarifying.

chaitanyaprem (approved):
LGTM

waku/v2/protocol/filter/options.go
chaitanyaprem: nvm, just noticed that you had already changed defaults.

PR #1261, opened 2024-11-29
feat: option to emit event to bus upon message sent

adklempner:
# Description

Adds an option when creating message sender to emit an event each time a message is sent.

# Changes

- [ ] new option `WithMessageSender` which instructs `MessageSender` to push an event to the `EventBus` each time a message is sent
- [ ] new event `MessageSent` containing timestamp and payload size

# Tests

Added a new unit test and tested locally in status-go via this PR https://github.com/status-im/status-go/pull/6157

### DIFF
--- a/waku/v2/api/publish/message_sender.go
+++ b/waku/v2/api/publish/message_sender.go
@@ -6,6 +6,8 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/libp2p/go-libp2p/core/event"
+	"github.com/libp2p/go-libp2p/core/host"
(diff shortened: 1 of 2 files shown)


richard-ramos (approved):

chaitanyaprem (approved):
change LGTM, but test needs to be fixed that is failing. 

probably due to invoking wg.wait in a go routine.

adklempner:
> change LGTM, but test needs to be fixed that is failing.
> 
> probably due to invoking wg.wait in a go routine.

fixed, the problem was not closing the subscription (it got triggered after the test ended making it fail)

PR #1265, opened 2024-12-10
fix: fatal error: concurrent map writes

plopezlpz:
# Description

Getting an error in status-go:
```
fatal error: concurrent map writes
```


chaitanyaprem (approved):
Good find
LGTM

richard-ramos:
Can we get this in 2.32?

richard-ramos (approved):

PR #1266, opened 2024-12-12
fix: concurrent peerMap iteration and write

qfrank:
IMG_0467


waku/v2/peermanager/peer_selection.go
```diff
@@ -130,6 +130,8 @@ func (pm *PeerManager) selectServicePeer(criteria PeerSelectionCriteria) (PeerSe
 			if len(criteria.PubsubTopics) == 0 || (len(criteria.PubsubTopics) == 1 && criteria.PubsubTopics[0] == "") {
 				return slot.getRandom(criteria.MaxPeers, criteria.ExcludePeers)
 			} else { //PubsubTopic based selection
+				slot.mu.RLock()
+				defer slot.mu.RUnlock()
```
plopezlpz, line 134: instead of using defer, can you move this to line 141 just after the for loop, it is because defer will run only at the end of the function call, we don't need to lock this for that long

plopezlpz (approved):

chaitanyaprem (approved):
LGTM

Interesting that this is not detected by go race detector.

PR #1269, opened 2024-12-19
refactor: use peerInfo and multiaddresses instead of peerID

richard-ramos:
Required by https://github.com/status-im/status-go/pull/5946

The interaction with libwaku is done thru multiaddresses instead of peerID, because the peerstore is not exposed like we do in go-waku. So, to make it possible to use both go-waku and nwaku in status-go (depending on a build flag), changes are done both in go-waku and the `api` package to use multiaddresses and/or peer.AddrInfo where needed.

It's worth mentioning that the cases in which a `peer.AddrInfo` are those in which a node can have more than one multiaddress (like the storenodes), and in these cases the `peer.AddrInfo` will be translated into a comma separated string containing all the multiaddresses like it is done here: https://github.com/status-im/status-go/blob/1348510acfdfd93dbe5fc5b0a23e9caf532272a3/wakuv2/nwaku.go#L2904-L2910

This PR should not be merged until  https://github.com/status-im/status-go/pull/5946 is ready to be merged

cc: @Ivansete-status @gabrielmer

### DIFF
--- a/library/node.go
+++ b/library/node.go
@@ -352,7 +352,7 @@ func AddPeer(instance *WakuInstance, address string, protocolID string) (string,
 		return "", err
 	}
 
-	peerID, err := instance.node.AddPeer(ma, peerstore.Static, instance.relayTopics, libp2pProtocol.ID(protocolID))
+	peerID, err := instance.node.AddPeer([]multiaddr.Multiaddr{ma}, peerstore.Static, instance.relayTopics, libp2pProtocol.ID(protocolID))
 	if err != nil {
 		return "", err
 	}
--- a/cmd/waku/node.go
+++ b/cmd/waku/node.go
@@ -386,7 +386,7 @@ func Execute(options NodeOptions) error {
 	if options.PeerExchange.Enable && options.PeerExchange.Node != nil {
 		logger.Info("retrieving peer info via peer exchange protocol")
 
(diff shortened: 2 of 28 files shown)


library/node.go
```diff
@@ -352,7 +352,7 @@ func AddPeer(instance *WakuInstance, address string, protocolID string) (string,
 		return "", err
 	}
 
-	peerID, err := instance.node.AddPeer(ma, peerstore.Static, instance.relayTopics, libp2pProtocol.ID(protocolID))
+	peerID, err := instance.node.AddPeer([]multiaddr.Multiaddr{ma}, peerstore.Static, instance.relayTopics, libp2pProtocol.ID(protocolID))
```
Ivansete-status, line 355: Would it be possible to avoid that "cast" , `[]multiaddr.Multiaddr{...}`? Maybe we can make it once within the `AddPeer` func.

Ivansete-status (approved):
LGTM! Thanks for it! :100: 
Just one nitpick comment

chaitanyaprem (approved):
Changes LGTM

library/node.go
richard-ramos: Probably a late reply, but this is not doing a cast, but defining a slice of multiaddr of a single element!

PR #1278, opened 2025-03-20
fix: setting peerId for store query

gabrielmer:
# Description
Improving a small change introduced in https://github.com/waku-org/go-waku/pull/1269 which breaks the store queries in `status-go` 

# Changes

- [x] setting the peerId before performing a store query and adding the peer to the peer store

## Issue
 https://github.com/status-im/status-go/issues/6439 and https://github.com/status-im/status-mobile/issues/22343


chaitanyaprem (approved):
LGTM

richard-ramos (approved):

PR #1284, opened 2025-05-24
fix: remove automatic relay unsubscribe in a goroutine

igor-sirotin:
# Description

1. I've removed this:
    https://github.com/waku-org/go-waku/blob/8b0e03113db6db3dcd2fbef9a97380f2b8631098/waku/v2/protocol/relay/waku_relay.go#L441-L445
    
    Because the client will call `Unsubscribe` in the end anyway. 
    At the same time for current implementation we need to provide some `ctx`, which is complicating `status-go` code.
    
    I've kept the `ctx` argument for `Subscribe` and `Unsubscribe` not to break compatibility.

2. Also added a getter for `Subscription.contentFilter` to simplify status-go code

### DIFF
--- a/waku/v2/protocol/relay/subscription.go
+++ b/waku/v2/protocol/relay/subscription.go
@@ -3,8 +3,9 @@ package relay
 import (
 	"context"
 
-	"github.com/waku-org/go-waku/waku/v2/protocol"
 	"golang.org/x/exp/slices"
+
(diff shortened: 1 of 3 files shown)


richard-ramos (approved):
Should be okay to remove the the `ctx` parameter.

jazzz (approved):
👍 - Much cleaner

PR #1301, opened 2026-04-21
fix: SetClosing no longer holds SubscriptionDetails lock across a potentially blocking channel send

alexjba:
# Description

A deadlock has been identified in Status app when stress testing the background/foreground app cycles. https://github.com/status-im/status-app/issues/20460

The deadlock is triggered by status-go through `ConnectionChanged`.

- SetClosing(): send moved out of the s.Lock() critical section, performed via select/default to tolerate stalled receivers and concurrent CloseC that may have closed the channel.
- Closing channel: changed from unbuffered (make(chan bool)) to buffered size 1 (make(chan bool, 1)) in NewSubscription. Pairs with the s.Closed guard (single send per subscription), so the sender never parks.

# Changes

  waku/v2/protocol/subscription/subscription_details.go

  - SetClosing(): send moved out of the s.Lock() critical section, performed via select/default to tolerate stalled receivers and concurrent CloseC that may have closed the channel.

  waku/v2/protocol/subscription/subscriptions_map.go

  - Closing channel: changed from unbuffered (make(chan bool)) to buffered size 1 (make(chan bool, 1)) in NewSubscription. Pairs with the s.Closed guard (single send per subscription), so the sender never parks.

  waku/v2/protocol/subscription/subscriptions_map_test.go

  - Added TestSetClosingDoesNotHoldInnerLock: reproduces the deadlock on the unpatched code (reader calling GetSubscriptionsForPeer → isPartOf hangs while a SetClosing with no receiver holds the inner mutex) and passes after the patch.
 
# Tests

TestSetClosingDoesNotHoldInnerLock

### DIFF
--- a/waku/v2/protocol/subscription/subscription_details.go
+++ b/waku/v2/protocol/subscription/subscription_details.go
@@ -111,11 +111,18 @@ func (s *SubscriptionDetails) Close() error {
 }
 
 func (s *SubscriptionDetails) SetClosing() {
+	// Flip the Closed flag under s.Lock(), but do NOT hold s.mu across the
+	// channel send
 	s.Lock()
-	defer s.Unlock()
-	if !s.Closed {
-		s.Closed = true
-		s.Closing <- true
+	shouldSignal := !s.Closed
+	s.Closed = true
+	s.Unlock()
+
+	if shouldSignal {
+		select {
+		case s.Closing <- true:
+		default:
+		}
 	}
 }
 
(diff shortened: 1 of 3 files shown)


chaitanyaprem (approved):
LGTM
Thanks!

alexjba:
@chaitanyaprem Thanks for reviewing it! I'll probably need some help to merge it as well. I don't have enough mana in this repo to do any action.

PR #1303, opened 2026-06-02
fix(filter): suppress subscription renewal in background mode

xAlisher:
## Problem

On mobile (Android/iOS), Waku filter subscriptions have a relay-side TTL of ~13.5 minutes (observed). When a subscription expires, the `closing` channel fires in `subscriptionLoop`, `checkAndResubscribe` runs, and a new `wf.Subscribe()` RPC is issued. This wakes the LTE modem from RRC Idle state.

With the LTE modem's RRC tail timer (5–15s per carrier), each renewal produces a radio active window. On a loaded account (50 contacts, 3 communities) this generates **~144 mAh/hr** of battery drain while the screen is locked — measured via `dumpsys batterystats` (status-im/status-app#21045).

The renewal runs at the same rate whether the user is actively using the app or the phone has been locked for hours.

## Fix

Adds a `backgroundMode atomic.Bool` to `Sub` and a `SetBackgroundMode(bool)` method to both `Sub` and `FilterManager`.

**When `background=true`** (screen locked):
- The 5s health-check ticker is skipped
- Expired subscription IDs from the `closing` channel are dropped without resubscription
- No network I/O occurs

**When `background=false`** (app returns to foreground):
- `SetBackgroundMode(false)` immediately sends to `closing` to trigger a resubscription
- All expired filters reconnect in one burst on foreground open

## Impact

| Scenario | Before | After |
|---|---|---|
| Overnight cellular (loaded account) | ~144 mAh/hr | ~5–10 mAh/hr (modelled) |
| Radio wakeups/hr from filter renewal | ~4.4 | ~0 in background |
| Foreground latency | unchanged | +1 reconnect burst on open |

Messages sent while backgrounded are still delivered via the mailserver on foreground — not lost.

## Usage (status-go)

```go
// On background:
filterManager.SetBackgroundMode(true)

// On foreground:
filterManager.SetBackgroundMode(false)  // triggers immediate resubscription
```

Companion status-go PR: https://github.com/status-im/status-go/pull/7508

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/waku/v2/api/filter/filter.go
+++ b/waku/v2/api/filter/filter.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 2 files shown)


xAlisher:
Wrong repo — resubmitting to correct target.

PR #1308, opened 2026-06-08
fix: reset peer backoff on recovery so reconnect is not blocked

### LINKED ISSUE #1307: bug: stale peer backoff prevents reconnection after recovery (sleep/wake)
### Summary
After a recovery event — device suspend/resume, or keep-alive detecting the node is
disconnected — go-waku kills all connections to force a fresh start. But the per-peer
connection backoff cache in `PeerConnectionStrategy` still holds the accumulated
backoffs (which grow up to ~1 hour). The peer connector then refuses to redial those
peers until the backoff expires, so the node stays disconnected far longer than needed.

### Steps to reproduce
1. Run a node connected to several peers and let connection backoffs accumulate.
2. Trigger a recovery path: suspend/resume the host, or let keep-alive pings fail
   `maxAllowedSubsequentPingFailures` times.
3. keep-alive logs "Killing all connections" and disconnects all peers.
4. **Observed:** the node does not reconnect promptly — peers are skipped due to their
   existing backoff until it expires.

**Expected Behavior**
After a full-disconnect recovery event, all known peers should be eligible for immediate redial.

### Environment
go-waku v0.10.x; affects mobile/desktop after sleep/wake or network loss.

### go-waku version/commit hash
v0.9.0, v0.10.1

kblinichkin:
# Description
When keep-alive detects the node is disconnected (no recent keep-alive, or N subsequent
ping failures) it disconnects all peers to force a fresh start. But `PeerConnectionStrategy`
still holds each peer's accumulated backoff, so the connector skips those peers until the
backoff — which can grow to ~1h — expires, leaving the node disconnected. This adds
`ResetBackoff()` and calls it on the recovery paths so every known peer can be redialed
immediately once the network is back.

# Changes
- [x] Add `PeerConnectionStrategy.ResetBackoff()` to purge the per-peer backoff cache
- [x] Call it in `startKeepAlive` after `disconnectAllPeers` on both recovery branches

# Tests
- [x] `go build ./...` and `gofmt` clean
- [ ] Manually verified peers are redialed immediately after a forced full disconnect
- [x] Manually test the client app reconnects to peers after a recovery event

## Issue

closes #1307

### DIFF
--- a/waku/v2/node/keepalive.go
+++ b/waku/v2/node/keepalive.go
@@ -90,11 +90,13 @@ func (w *WakuNode) startKeepAlive(ctx context.Context, randomPeersPingDuration t
 				lastTimeExecuted = w.timesource.Now()
 				w.log.Warn("keep alive hasnt been executed recently. Killing all connections")
 				disconnectAllPeers(w.host, w.log)
+				w.peerConnector.ResetBackoff()
 				continue
 			} else if iterationFailure >= maxAllowedSubsequentPingFailures {
 				iterationFailure = 0
 				w.log.Warn("Pinging random peers failed, node is likely disconnected. Killing all connections")
 				disconnectAllPeers(w.host, w.log)
+				w.peerConnector.ResetBackoff()
 				continue
 			}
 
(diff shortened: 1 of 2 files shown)


igor-sirotin (approved):
Looks good, thank you!

PR #1316, opened 2026-08-02
fix(build): fix Android library builds on Go 1.23+ and non-Linux hosts

kblinichkin:
# Description
Building the go-waku C libraries for Android currently fails in two ways, and
release artifacts are larger than they need to be:

1. **Go 1.23+ rejects the build.** The indirect dependency `github.com/wlynxg/anet`
   (pulled in via go-libp2p) uses `//go:linkname` in `interface_android.go`, which
   Go 1.23 disallows by default. That file is Android-only, so CI on Linux/macOS
   never compiles it — but every Android build of the C libraries hits it.

2. **Cross-compiling from a non-Linux host produces the wrong artifact.** The
   shared-library extension and soname are selected from `detected_OS` (the *build
   host*), not the target. Building for Android on macOS/Windows therefore yields
   `libgowaku.dylib` / `libgowaku.dll` with no soname, instead of an ELF
   `libgowaku.so`.

3. **Release builds keep debug data they don't need.** The library targets pass no
   linker flags, so the symbol table and DWARF info are retained even in release
   artifacts — roughly 10 MB of dead weight in a shipped application. `-s -w` is
   applied only for `BUILD_TYPE=release`, so releases shrink while other build
   types keep their symbols. The default is unchanged.

Android also cannot load a versioned soname — an APK only ships `lib*.so`, symlinks
do not survive APK packaging, and bionic has no soname versioning — so the Android
`.so` stays unversioned and skips the `.so.0` rename/symlink that desktop Linux uses.

# Changes
- [x] Force `GOBIN_SHARED_LIB_EXT := so` and an unversioned soname when `GOOS=android`, independent of the build host
- [x] Add `GO_LDFLAGS_EXTRA` with `-checklinkname=0` for `GOOS=android`
- [x] Skip the `.so.0` rename/symlink for Android
- [x] Strip the symbol table and DWARF (`-s -w`) for `BUILD_TYPE=release` to reduce library size

# Tests
- [x] `make -n` verified for `build`, `dynamic-library`
- [x] Android (arm64, API 21, NDK 26.1) library built with:
      ```
      GOOS=android GOARCH=arm64 CGO_ENABLED=1 \
        CC=$ANDROID_NDK/toolchains/llvm/prebuilt/<host>/bin/aarch64-linux-android21-clang \
        CXX=$ANDROID_NDK/toolchains/llvm/prebuilt/<host>/bin/aarch64-linux-android21-clang++ \
        BUILD_TYPE=release make dynamic-library
      ```
      and verified loading on-device
- [x] A versioned soname was tested and confirmed **not** loadable on Android, hence the unversioned soname
- [x] No behaviour change for desktop Linux (soname `libgowaku.so.0` + symlink), macOS (`libgowaku.dylib`) or Windows (`libgowaku.dll`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,15 @@ else
  GOBIN_SHARED_LIB_CGO_LDFLAGS := CGO_LDFLAGS="-Wl,-soname,libgowaku.so.0"
 endif
 
+# Cross-compiling for Android always produces an ELF .so, whatever the build
+# host is, so the host detection above would pick the wrong extension on macOS
+# and Windows. Android also cannot load a versioned soname: the APK only ships
+# lib*.so and its linker has no version support, so use a plain libgowaku.so.
+ifeq ($(GOOS),android)
+GOBIN_SHARED_LIB_EXT := so
+GOBIN_SHARED_LIB_CGO_LDFLAGS := CGO_LDFLAGS="-Wl,-soname,libgowaku.so"
+endif
+
 GIT_COMMIT = $(shell git rev-parse --short HEAD)
 VERSION = $(shell cat ./VERSION)
 UID := $(shell id -u)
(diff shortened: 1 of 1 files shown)


igor-sirotin (approved):
I see all same flags are used in `status-go` 👍

PR #1318, opened 2026-08-04
feat(library): emit peer connection status events in the C API

kblinichkin:
# Description
go-waku can already notify about peer connect/disconnect events via
`node.WithConnectionNotification(ch)`, but the C API does not expose it. C/C++
consumers therefore have no way to observe connectivity changes and must poll
`waku_peer_cnt` to infer them.

This forwards those notifications to the existing signal callback as a
`conStatus` event, so consumers receive them the same way they already receive
other asynchronous events.

# Changes
- [x] Add `connCh chan node.PeerConnection` to `WakuInstance` and register it via `node.WithConnectionNotification`
- [x] Emit a `conStatus` signal carrying `{"peerId": "...", "connected": true|false}`
- [x] Forward notifications from a goroutine started in `Start()`, exiting on `ctx.Done()` or channel close
- [x] Close the channel in `Stop()`, after `stop()` has unregistered the connection notifier

# Tests
- [x] `gofmt`, `go vet ./library/...`, `go build ./library/...` clean
- [x] Verified from a C++ consumer: signals arrive as
      `{"type":"conStatus","event":{"peerId":"16Uiu2HAm...","connected":true}}`

# Notes
- Purely additive (+38/−0): no existing symbol or behaviour changes; consumers that
  ignore `conStatus` are unaffected.
- The channel is closed only after `stop()` runs, because `node.Stop()` unregisters
  the ConnectionNotifier (via `StopNotify`) before `host.Close()` fires disconnect
  events. Closing earlier risks a send-on-closed-channel panic.

### DIFF
--- a/library/node.go
+++ b/library/node.go
@@ -39,6 +39,7 @@ type WakuInstance struct {
 
 	node                *node.WakuNode
 	cb                  unsafe.Pointer
+	connCh              chan node.PeerConnection
 	mobileSignalHandler MobileSignalHandler
 
 	relayTopics []string
@@ -127,6 +128,18 @@ func validateInstance(instance *WakuInstance, validationType ValidationType) err
 	return nil
 }
 
+type connectionStatMsg struct {
+	PeerID    string `json:"peerId"`
+	Connected bool   `json:"connected"`
+}
+
+func toConnectionStatus(msg node.PeerConnection) *connectionStatMsg {
+	return &connectionStatMsg{
+		PeerID:    msg.PeerID.String(),
+		Connected: msg.Connected,
+	}
+}
+
 // NewNode initializes a waku node. Receives a JSON string containing the configuration, and use default values for those config items not specified
 func NewNode(instance *WakuInstance, configJSON string) error {
 	if err := validateInstance(instance, NotConfigured); err != nil {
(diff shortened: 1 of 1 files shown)


library/node.go
```diff
@@ -272,6 +288,21 @@ func Start(instance *WakuInstance) error {
...
 
+	go func() {
+		defer utils.LogOnPanic()
+		for {
+			select {
+			case <-instance.ctx.Done():
+				return
+			case item, ok := <-instance.connCh:
+				if !ok {
+					return
```
Ivansete-status, line 300: Perhaps we can have better error debug log?

Ivansete-status (approved):

library/node.go
```diff
@@ -272,6 +288,21 @@ func Start(instance *WakuInstance) error {
...
+				}
+				send(instance, "conStatus", toConnectionStatus(item))
```
igor-sirotin:
```suggestion
				send(instance, "connectionStatus", toConnectionStatus(item))
```

igor-sirotin (commented):
Just one comment, please resolve it's good to go

library/node.go
kblinichkin: Renamed to connectionStatus as suggested.

library/node.go
kblinichkin: Added a debug log before the listener exits. Used Debug since a closed channel is the normal shutdown path.

igor-sirotin (approved):

PR #1319, opened 2026-08-08
feat(library): allow a user_data pointer for the event callback

kblinichkin:
# Description
`WakuCallBack` is declared as `void (*)(int ret_code, const char* msg, void* user_data)`,
and 16 of the exported C functions already accept a `userData` pointer that is handed
back to the callback. `waku_set_event_callback` is the exception: it provides no way to
set one, so the signal callback always receives `user_data == NULL`.

That makes it impossible to associate a signal with a specific context without global
state — a problem for any consumer running more than one waku instance in a process,
for example a C++ wrapper that needs to route signals to the owning object.

This adds `waku_set_event_callback_with_user_data`, leaving the existing function and
the callback ABI untouched.

# Changes
- [x] Add `waku_set_event_callback_with_user_data(ctx, cb, userData)` C export
- [x] Add `SetEventCallbackWithUserData`; `SetEventCallback` now delegates to it with `nil`
- [x] Store the pointer on `WakuInstance` and pass it through `ServiceSignalEvent`
- [x] Correct the `waku_set_event_callback` doc comment, which described a callback
      signature (`void myCallback(char* signalJSON)`) that no longer matches `WakuCallBack`

# Tests
- [x] `gofmt`, `go vet ./library/...`, `go build ./library/...` clean
- [x] `make static-library` and `make dynamic-library` build; the generated `libgowaku.h`
      shows `waku_set_event_callback` unchanged with the new symbol added alongside:
      ```
      extern void waku_set_event_callback(void* ctx, WakuCallBack cb);
      extern void waku_set_event_callback_with_user_data(void* ctx, WakuCallBack cb, void* userData);
      ```
- [x] Verified from a C++ consumer: signals now arrive with the supplied `user_data`

### DIFF
--- a/library/c/api.go
+++ b/library/c/api.go
@@ -297,7 +297,7 @@ func waku_set_event_callback(ctx unsafe.Pointer, cb C.WakuCallBack) {
 		panic(err.Error()) // TODO: refactor to return an error instead of panic
 	}
 
-	library.SetEventCallback(instance, unsafe.Pointer(cb))
+	library.SetEventCallback(instance, unsafe.Pointer(cb), nil)
 }
 
 // Same as waku_set_event_callback, but the callback receives the given userData
@@ -307,10 +307,10 @@ func waku_set_event_callback(ctx unsafe.Pointer, cb C.WakuCallBack) {
 func waku_set_event_callback_with_user_data(ctx unsafe.Pointer, cb C.WakuCallBack, userData unsafe.Pointer) {
 	instance, err := getInstance(ctx)
 	if err != nil {
-		panic(err.Error()) // TODO: refactor to return an error instead of panic
+		panic(err.Error())
 	}
 
-	library.SetEventCallbackWithUserData(instance, unsafe.Pointer(cb), userData)
+	library.SetEventCallback(instance, unsafe.Pointer(cb), userData)
 }
 
 // Retrieve the list of peers known by the waku node
--- a/library/c/api.go
+++ b/library/c/api.go
@@ -287,8 +287,8 @@ func waku_default_pubsub_topic(cb C.WakuCallBack, userData unsafe.Pointer) C.int
 }
 
 // Register callback to act as signal handler and receive application signals
-// (in JSON) which are used to react to asynchronous events in waku. The function
-// signature for the callback should be `void myCallback(char* signalJSON)`
+// (in JSON) which are used to react to asynchronous events in waku. The callback
+// is invoked as `myCallback(0, signalJSON, NULL)`.
 //
 //export waku_set_event_callback
 func waku_set_event_callback(ctx unsafe.Pointer, cb C.WakuCallBack) {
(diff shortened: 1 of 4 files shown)


igor-sirotin (commented):
I don't quite like the idea of passing arbitrary data in event. 
But I agree that identifying the node instance is impossible at the moment.

@kblinichkin would something like this work for you instead?

- https://github.com/logos-messaging/logos-delivery-go/pull/1324

kblinichkin:
> I don't quite like the idea of passing arbitrary data in event. But I agree that identifying the node instance is impossible at the moment.
> 
> @kblinichkin would something like this work for you instead?
> 
> * #1324

Yes, with #1324 it is possible to implement multiple instance management. And as far as I see it fixes one more real issue...

However, on the C++ client side it costs a static id->object registry, a mutex, deregistration in the destructor, and a lookup per signal, versus a single pointer cast. 
Moreover, user_data stays dead weight: it is the canonical way a C++ wrapper binds a static trampoline back to its object, and waku_set_event_callback would remain the one export that cannot use it.

library/signals.go
```diff
@@ -50,19 +50,26 @@ func send(instance *WakuInstance, signalType string, event interface{}) {
...
+// SetEventCallbackWithUserData is like SetEventCallback, but passes cbUserData
+// back to the callback as its user_data argument.
+func SetEventCallbackWithUserData(instance *WakuInstance, cb unsafe.Pointer, cbUserData unsafe.Pointer) {
 	if err := validateInstance(instance, None); err != nil {
 		panic(err.Error())
 	}
 
 	instance.cb = cb
+	instance.cbUserData = cbUserData
 }
```
igor-sirotin, line 68: In C it's indeed nice to have 2 separate functions (to avoid breaking changes).
But here we're okay to just add the arg directly.

```suggestion
// SetEventCallback is to set a callback in order to receive application
// signals which are used to react to asynchronous events in waku.
func SetEventCallback(instance *WakuInstance, cb unsafe.Pointer, cbUserData unsafe.Pointer) {
    if err := validateInstance(instance, None); err != nil {
        panic(err.Error())
    }
    
    instance.cb = cb
  	instance.cbUserData = cbUserData
}
```

library/c/api.go
```diff
@@ -300,6 +300,19 @@ func waku_set_event_callback(ctx unsafe.Pointer, cb C.WakuCallBack) {
...
 }
 
+// Same as waku_set_event_callback, but the callback receives the given userData
+// as its user_data argument instead of NULL.
+//
+//export waku_set_event_callback_with_user_data
+func waku_set_event_callback_with_user_data(ctx unsafe.Pointer, cb C.WakuCallBack, userData unsafe.Pointer) {
+	instance, err := getInstance(ctx)
+	if err != nil {
+		panic(err.Error()) // TODO: refactor to return an error instead of panic
```
igor-sirotin: I see that `panic` is mostly used in this project in such cases, so I'd keep it the same here.
Please don't leave a TODO here, either resolve or leave as is

igor-sirotin: (I see it's not your comment though)

igor-sirotin (commented):
> However, on the C++ client side it costs a static id->object registry, a mutex, deregistration in the destructor, and a lookup per signal, versus a single pointer cast. 
Moreover, user_data stays dead weight: it is the canonical way a C++ wrapper binds a static trampoline back to its object, and waku_set_event_callback would remain the one export that cannot use it.

@kblinichkin I agree. Let's keep your solution.
Though you'd probably need a mutex anyway, even with `user_data` in the callback. 

I'm happy with it, left some minor comments. Please resolve and it's good to go.

igor-sirotin (approved):

PR #1324, opened 2026-08-21
feat(library): identify the emitting instance in every signal

igor-sirotin:
# Description

`waku_set_event_callback(ctx, cb)` already registers the callback per instance, but the signal itself carries no identity. A consumer running more than one waku instance in a process registers the same function pointer on each of them — a C++ wrapper has one static trampoline, it cannot synthesise a distinct function pointer per object — and then has no way to tell which instance a signal came from.

This tags every signal with the id of the instance that emitted it, and adds `waku_instance_id` so a C consumer can map its opaque `ctx` to that id and build a routing table.

Alternative to #1319, which solves the same problem by threading a `user_data` pointer
through to the callback. Trade-offs in the section below; happy to take either, or both.

### DIFF
--- a/examples/c-bindings/main.c
+++ b/examples/c-bindings/main.c
@@ -56,7 +56,7 @@ void callBack(int ret, const char *signal, void *user_data)
 
   // Example signal:
   /*{
-      "nodeId":1,
+      "instanceId":0,
       "type":"message",
       "event":{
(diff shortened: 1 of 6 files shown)


igor-sirotin:
Closing in favour of https://github.com/logos-messaging/logos-delivery-go/pull/1319

